PR #11, opened 2021-10-17
Add a couple exports to ConnectionProvider

Ericnr:
I'm trying to build on top of the lib and this little change will allow me to not need to maintain a fork. Been looking for a lib like this for a while, I'll be making the repo public soon!


Ericnr:
That could work, but it'd be a little more work for me. I need access to the `DefaultConnectionProvider` and `SingleConnectionProvider` classes, so with the interface I could implement them myself, but since they're already implemented in the lib I could just use them too

koskimas:
Ok, I can export these. It's just that the more I export, the more it limits what I can change in the library without breaking the API. The less I can change, the slower the development of the library will be.

PR #49, opened 2022-01-07
Fix issue with esbuild class transform

samsouder:
If you set `this.#props` and call `super()` in both `if` blocks, when using esbuild it will transform them and put the `this` call first causing runtime failure.

### DIFF
--- a/src/kysely.ts
+++ b/src/kysely.ts
@@ -94,16 +94,16 @@ export class Kysely<DB> extends QueryCreator<DB> {
       )
 
       superProps = { executor, parseContext }
-      props = freeze({
+      props = {
         config: args,
         executor,
         dialect,
         driver: runtimeDriver,
         parseContext,
-      })
+      }
     }
     super(superProps);
-    this.#props = props;
+    this.#props = freeze(props);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


samsouder:
esbuild haven't covered this scenario as they feel it is very uncommon and hard to implement on their side: https://github.com/evanw/esbuild/issues/1918

samsouder:
For a bit more background -- this tweak will allow folks using the built-in Lambda creation with AWS CDK to use Kysely out of the box without custom JS builders.

koskimas:
I'm ok with this fix, but the implementation could be better. `_super` and `_props` have the type `any`. Also variables that start with an underscore are used nowhere in kysely's codebase, so maybe just give more descriptive names for them:

```ts
let superProps: QueryProps | undefined
let props: KyselyProps | undefined
```

koskimas:
You could also move the `freeze` call to the last line.

samsouder:
All fixed up, let me know if that works for you or not.

koskimas:
Perfect!

koskimas:
I'll release a new version soon.

koskimas:
0.16.0 is out with this fix

PR #50, opened 2022-01-08
Make esm output deno compatible

barthuijgen:
Reference changes for this comment: https://github.com/koskimas/kysely/issues/33#issuecomment-1007995224

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "test:esmimports": "node scripts/check-esm-imports.js",
     "prettier": "prettier --write src/**/*.ts test/**/*.ts",
     "build": "npm run clean && npm run build:esm && npm run build:cjs && npm run script:module-fixup && npm run script:copy-interface-doc",
-    "build:esm": "tsc -p tsconfig.json",
+    "build:esm": "tsc -p tsconfig.json && node scripts/add-types-directive.js",
     "build:cjs": "tsc -p tsconfig-cjs.json",
     "docs:build": "typedoc --excludePrivate --excludeInternal --customCss ./docs-theme.css src/index.ts && npm run docs:postprocess",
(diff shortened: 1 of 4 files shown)


koskimas:
The missing `.js` endings have now been fixed in 0.16.2.

We can't remove `Buffer`. It's important for node users (most users). Instead, in 0.16.2 `Buffer` is optional and should work in deno as well.

The only thing missing now are those `/// <reference types="./${baseName}" />` headers, meaning of which I don't understand.

barthuijgen:
That's good news, i'll try out 0.16.2 soon!

The triple slash derectives are interpeted by Deno, they would not be used by any other runtime as far as I know. see: https://deno.land/manual@v1.9.2/typescript/types#using-the-triple-slash-reference-directive

The comment I made in the "Node-less version" thread should explain in more detail. https://github.com/koskimas/kysely/issues/33#issuecomment-1007995224

The `/// <reference types="./${baseName}" />` was just the snippet of the node script I used, it would need to be a link to the files .d.ts version so in `index.js` you would add `/// <reference types="./index.d.ts" />`

This is something that would not be needed when using Skypack (assuming it builds correctly), since it uses the `X-TypeScript-Types` header as replacement for the triple slash directive.

barthuijgen:
Found a different CDN that actually runs kysely in Deno, but again has the problem types are missing.

```
import { Kysely } from "https://cdn.jsdelivr.net/npm/kysely/dist/esm/index-nodeless.js/+esm";

const db = new Kysely({});
```

koskimas:
Adding the `/// <reference types="./${baseName}" />` lines seems to break things for node 🤦

barthuijgen:
I'll close this since it served it's purpose. Any further discussion can be done in https://github.com/koskimas/kysely/issues/33

PR #71, opened 2022-03-08
docs: fix typo concate -> concat

KristjanTammekivi:
I didn't change it in docs/modules.html, I assume it gets generated automatically...somewhere. When I ran docs:build around 200 files changed so that didn't see mright either


koskimas:
Thanks! docs:build updates links to source code in every file. That's why you get a bunch of changes.

PR #78, opened 2022-04-16
add binded sql to kysely instance

ejose19:
Basically a convenience property to not have to import `sql` separately nor specifying any argument on `execute` method.

Please let me know if there's a way to get rid of the cast in

```ts
// src/raw-builder/raw-builder.ts [line 34]
this.#executor = executor as (B extends true ? QueryExecutor : undefined)
```

and the redundant type guard on 

```ts
// src/raw-builder/raw-builder.ts [line 133]
if (executor === undefined) {
  throw new Error('Unreachable code')
}
```


koskimas:
I don't think this is a good idea. Having two different `sql` functions will cause confusion. Kysely used to have a bound method like this but I specifically removed it for simplicity.

PR #106, opened 2022-06-28
Stream queries

elderapo:
- [x] Dialect implementations:
  - [x] `mysql`
  - [x] `sqlite` - since `better-sqlite3` [doesn't/doesn't plan to support streaming](https://github.com/WiseLibs/better-sqlite3/issues/241) error is thrown on use.
  - [x] `postgres`
- [x] More tests.
- [x] Write documentation?

closes https://github.com/koskimas/kysely/issues/105


koskimas:
Great work so far!

> It requires [pg-cursor](https://github.com/brianc/node-postgres/tree/master/packages/pg-cursor)/[pg-query-stream](https://github.com/brianc/node-postgres/tree/master/packages/pg-query-stream) external dependencies. WHAT DO???!!

I think the ~only option is to pass the `Cursor` constructor as a field to the `PostgresDialect`:

```ts
import Cursor from 'pg-cursor'

new PostgresDialect({
  pool: pool,
  cursor: Cursor
}
```

If the cursor constructor is not provided, an error should be thrown if the `stream` method is used.

About the MySQL code you now have: You never release the connection. What happens if the iteration is ended before the stream ends? You don't seem to handle that case either.

src/query-executor/query-executor-base.ts
```diff
@@ -68,6 +68,19 @@ export abstract class QueryExecutorBase implements QueryExecutor {
     })
   }
 
+  async *stream<R>(
+    compiledQuery: CompiledQuery,
+    queryId: QueryId
+  ): AsyncIterableIterator<QueryResult<R>> {
+    const connection = await this.provideConnection(
+      async (connection) => connection
```
koskimas, line 76: This code releases the connection to the pool before returning the connection. So you are actually using a released connection here.

You need to do the work inside the `provideConnection`'s callback argument. Only once the stream is done, should you release the connection back to the pool.

elderapo: I can't really do the awaiting inside of `provideConnection` callback in this case because of `yield` (can't return async generator as `T` from `provideConnection` because connection has to be active during lifetime of async generator). I have 2 ideas though:
1. I can prevent `provideConnection` from returning until the connection is not needed anymore (iterator is done).
2. Add "unmanaged" `getConnection` to `ConnectionProvider` interface:
```ts
export interface ConnectionProvider {
  provideConnection<T>(
    consumer: (connection: DatabaseConnection) => Promise<T>
  ): Promise<T>

  getConnection(): Promise<{ connection: DatabaseConnection, release: () => void }>
}
```

koskimas: Whichever is fine. I can refactor this if I find a better solution. (I probably won't,☺️)

src/dialect/postgres/postgres-driver.ts
```diff
@@ -112,8 +122,38 @@ class PostgresConnection implements DatabaseConnection {
...
 
-  async *executeQueryStream<R>(): AsyncIterableIterator<QueryResult<R>> {
-    throw new Error('@TODO: postgres::executeQueryStream')
+  async *executeQueryStream<O>(
+    compiledQuery: CompiledQuery,
+    chunkSize: number
+  ): AsyncIterableIterator<QueryResult<O>> {
+    if (!this.#options.cursorIpl) {
+      throw new Error(
+        "'cursorIpl' is not present is postgres dialect config! It's required for streaming for work in postgresql!"
```
boehs, line 131:
```suggestion
        "'cursorImpl' is not present in your postgres dialect config! It's required for streaming for work in postgresql!"
```

I'll see myself out :P

No but seriously, nice work. I'm also wondering about `Ipl` vs `Impl`. Are the various uses of `Ipl` throughout this change a typo?

Personally, I would prefer `cursor` instead of `cursorImpl` from a configuration standpoint for consistancy with `pool`, but no big deal either way

elderapo: > No but seriously, nice work. I'm also wondering about Ipl vs Impl. Are the various uses of Ipl throughout this change a typo?

Actually, I intended to use `Impl`. Good catch!

> Personally, I would prefer cursor instead of cursorImpl from a configuration standpoint for consistancy with pool, but no big deal either way

@koskimas what's your opinion on this one?

koskimas: I prefer just `cursor`. Also I think it should take the constructor itself and not a function. That returns an instance. Simply `cursor: Cursor`

boehs (approved):
Untested but looks good from a code quality standpoint

src/util/deferred.ts
```diff
@@ -0,0 +1,29 @@
+export class Deferred<T> {
+  private readonly _promise: Promise<T>
```
koskimas, line 2: You can (and should) use real private properties. The minimum supported node version already supports private properties and they are used extensively in Kysely.

src/query-executor/query-executor-base.ts
```diff
@@ -68,6 +69,35 @@ export abstract class QueryExecutorBase implements QueryExecutor {
...
+    }).catch((ex) => connectionDefer.reject(ex))
+
+    const connection = await connectionDefer.promise
+
+    try {
+      for await (const result of connection.executeQueryStream(
+        compiledQuery,
+        chunkSize
+      )) {
+        yield await this.#transformResult(result, queryId)
```
koskimas, line 94: If the iteration is ended prematurely in the user code, this loop will get stuck forever and it will hold the connection forever. Right?

I think you need to implement the `AsyncIterator` interface "manually". Here's the interface

```ts
interface AsyncIterator<T, TReturn = any, TNext = undefined> {
    // NOTE: 'next' is defined using a tuple to ensure we report the correct assignability errors in all places.
    next(...args: [] | [TNext]): Promise<IteratorResult<T, TReturn>>;
    return?(value?: TReturn | PromiseLike<TReturn>): Promise<IteratorResult<T, TReturn>>;
    throw?(e?: any): Promise<IteratorResult<T, TReturn>>;
}
```

There you can implement the `return` and `throw` methods to release the connection. They get called when the iteration is ended by the consumer.

Since this is a really critical thing (we don't want to leak connections) we need a test for this. One way to test this is to run a code that breaks the iteration midway multiple times. To be specific, the code needs to be run `connectionPoolMaxSize + 1` times. You could create a constant for that in the tests. The same constant would be used to create the kysely instances AND in this test.

koskimas: Also I think you need to reserve the connection only when the `next` method is called the first time to make sure has been started. Otherwise the case where user creates a stream but never iterates it will leak a connection.

koskimas: Or if this method returns an `AsyncIterable`, you can reserve the connection when the `[Symbol.asyncIterator]()` method is called. I don't think there's a need to use `AsyncIterableIterator`? Why is it used here? Doesn't this method actually return a `AsyncGenerator`?

elderapo: This doesn't seem to be the case. Added a test case for it though. Good idea!

README.md
```diff
@@ -31,16 +31,24 @@ You can find a more thorough introduction [here](https://www.jakso.me/blog/kysel
 
 # Table of contents
 
+- [Kysely](#kysely)
+- [Table of contents](#table-of-contents)
```
koskimas, line 35: These two don't need to be here.

koskimas:
Looking really good! Thank you so much for working on this!

I managed to create conflicts once again for you, sorry about that. Once those are resolved, this is good to go!

src/query-executor/query-executor-base.ts
koskimas: You're right! I had to write some test code to understand how generators created using `async function *func()` behave. It seems that the code after the last `yield` is never executed BUT the code in any `finally` blocks is executed. And the `return` calls "bubble in" when the outermost loop ends. I learn't something new today!

README.md
elderapo: It was added by prettier (I have it automatically run after each file save). Didn't even realize it added these 2.

koskimas:
This still has conflicts and I can't merge it. I'll merge as soon as the conflicts are fixed.

elderapo:
> This still has conflicts and I can't merge it. I'll merge as soon as the conflicts are fixed.

According to GH there are no conflicts 🤔

elderapo:
image
:thinking:

koskimas:
WTF 🤔

koskimas (approved):
Awesome! 💯 🥇

PR #116, opened 2022-07-15
adding `db.replaceInto(table)`.

igalklebanov:
Current `db.onDuplicateKeyUpdate(updates)` implementation isn't perfect for the "upsert many items" use case. You have to write your own wrappers that take some item from values and turn it into:
```typescript
{ [key: keyof TB]: sql`values(${sql.ref(key)})` }
```
tldr not great DX, complex for some consumers (let's shorten the learning curve), less readable code.

Instead of introducing a breaking change to the current implementation of `db.onDuplicateKeyUpdate(updates)`, let's provide an alternative.

[REPLACE statements](https://dev.mysql.com/doc/refman/8.0/en/replace.html) are part of the MySQL dialect, and behave like [INSERT ... ON DUPLICATE KEY UPDATE statements](https://dev.mysql.com/doc/refman/8.0/en/insert-on-duplicate.html), except they delete and insert on collisions instead of updating existing rows.

Deleting and inserting a row with the same primary key value won't hurt in a modern mysql environment, where there is no usage or support (planetscale, singlestore, etc.) of foreign key constraints (so no cascading deletes in case the row gets deleted before being inserted).

This is an easy win imho.

---

tldr

This pull request basically adds the following:

```typescript
await db.replaceInto('person').values(people)
```

that compiles to something like:

```typescript
{
    sql: 'replace into `person` (`id`, `first_name`, `last_name`, `age`) 
        values (?, ?, ?, ?), (?, ?, ?, ?);',
    parameters: [123, 'Jennifer', 'Aniston', 18, 456, 'Arnold', 'Schwarzenegger', 45]
}
```

and helps us upsert many items in MySQL without any extra work. 👍


scripts/check-esm-imports.js
```diff
@@ -40,11 +40,11 @@ function readLines(filePath) {
 }
 
 function isLocalImport(row) {
-  return row.includes("from '.") || row.includes('from ".')
+  return row.match(/from ['"]\./)
```
koskimas, line 43: Shouldn't this be `test` instead of `match`?

src/query-builder/replace-query-builder.ts
```diff
@@ -0,0 +1,373 @@
+import { OperationNodeSource } from '../operation-node/operation-node-source.js'
```
koskimas, line 1: Couldn't you use `InsertQueryBuilder` instead of creating this copy-pasted builder? It probably has features that are not supported by MySQL replace queries but I think that's less bad than having another query builder to maintain. You don't need to change the `InsertQueryBuilder's` documentation either. Simply an example in the `replaceInto` methods documentation is enough.

src/parser/replace-values-parser.ts
```diff
@@ -0,0 +1,40 @@
...
+  NonNullableInsertKeys,
+  NonNullableUpdateKeys,
+  NullableInsertKeys,
+  NullableUpdateKeys,
+  UpdateType,
+} from '../util/column-type.js'
+import { parseInsertObjectOrList } from './insert-values-parser.js'
+import { ValueExpression } from './value-parser.js'
+
+export type ReplaceObject<DB, TB extends keyof DB> = {
```
koskimas, line 14: Why is this type needed? Isn't the input for `replace into` exactly the same as for `insert into`? The only difference is that mysql deletes the old row if it exists. I should not affect the types.

src/query-builder/replace-query-builder.ts
koskimas: `replace into` is probably used rarely. We want to minimise the impact of features that are hardly ever used.

koskimas:
Really nice work! I think you could make this simpler by just using the `InsertQueryBuilder`, but other than that 💯

scripts/check-esm-imports.js
igalklebanov: nice catch!

changed to use RegExp#test.

src/parser/replace-values-parser.ts
igalklebanov: To my understanding, the argument of `replaceInto().values(arg)` differs from the argument of `insertInto().values(arg)` in that it should accept `ColumnType<any, never, any>` columns, while the latter should not. `replaceInto` is the permissive "everything goes" brother of `insertInto`, getting the values exactly where you expect them to be at all costs.

example:

```typescript
type Person {
  id: Generated<number> // auto-increment
  first_name: string
  middle_name: string | null
  last_name: string
  created_at: GeneratedAlways<string>
}

type Database {
  persons: Person
}

const person = {
  id: 15,
  first_name: 'Jennifer',
  last_name: 'Aniston',
}

db<Database>.insertInto('persons').values([person]) // typescript error at `[person]` (due to Generated<number>), which is fine, we want to insert new person and let the database generate id for us.

db<Database>.replaceInto('persons').values([person]) // should be OK! we want to insert a person we got, and if exists with id:15, to replace it.
```

igalklebanov: Not a big deal though, I'm OK with not having this difference between the two.

src/query-builder/replace-query-builder.ts
igalklebanov: I agree, this is a niche feature that is not worth all the additional complexity added in my first attempt.

igalklebanov:
Thanks for the great feedback! Did some changes following your comments.. 🤞

src/parser/replace-values-parser.ts
koskimas: Actually this is ok

```ts
const person = {
  id: 15,
  first_name: 'Jennifer',
  last_name: 'Aniston',
}

db.insertInto('persons').values([person]) 
```

You can insert values to auto incrementing rows. There's no typescript error in that case.

igalklebanov: You're right, my bad. 💯

PR #117, opened 2022-07-16
adding `between` & `not between` operators.

igalklebanov:
`BETWEEN ... AND ...` and `NOT BETWEEN ... AND ...` comparison operators exist in all three base dialects ([postgresql](https://www.postgresql.org/docs/current/functions-comparison.html#FUNCTIONS-COMPARISON-PRED-TABLE), [mysql](https://dev.mysql.com/doc/refman/8.0/en/comparison-operators.html#operator_between), [sqlite](https://sqlite.org/lang_expr.html#the_between_operator)).

Current implementation is not ideal as it forces consumers to either:

* Write `.where`/`.having` for `>=` and `<=` separately. 

   ```typescript
   db.selectFrom('person').where('age', '>=', 0).where('age', '<=', 18)
   ```

   This can be tedious and might force wrappers if you're, for example, constantly querying by date range. tldr not great DX.

* Write a raw query using `sql` component as follows:

   ```typescript
   db.selectFrom('person').where(sql`${sql.ref('age')} between ${0} and ${18}`)
   ```
  This can be complex to some consumers (we should shorten the learning curve), and makes the code less readable.

I believe a lot of people intuitively try to `where between` when picking up `kysely` and are surprised they don't have it in their autocomplete menu. Especially people coming from `knex`, where [this](http://knexjs.org/guide/query-builder.html#wherebetween) option exists.

---

tldr

This pull request basically adds the following:

```typescript
await db.selectFrom('person').where('age', 'between', [0, 18]).selectAll()
```

that compiles to something like:

```typescript
{
    sql: 'select * from `person` where `age` between ? and ?'
    parameters: [0, 18]
}
```

### DIFF
--- a/scripts/check-esm-imports.js
+++ b/scripts/check-esm-imports.js
@@ -40,11 +40,11 @@ function readLines(filePath) {
 }
 
 function isLocalImport(row) {
-  return row.includes("from '.") || row.includes('from ".')
+  return row.match(/from ['"]\./)
 }
 
 function isDotJsImport(row) {
-  return row.endsWith(".js'")
+  return row.match(/.js['"][\n\t\s]*$/)
 }
 
 checkDir(path.join(__dirname, '..', 'src'))
--- a/src/operation-node/operator-node.ts
+++ b/src/operation-node/operator-node.ts
@@ -39,6 +39,8 @@ export const OPERATORS = [
   '@@@',
   '!!',
   '<->',
+  'between',
+  'not between'
 ] as const
 
 export type Operator = ArrayItemType<typeof OPERATORS>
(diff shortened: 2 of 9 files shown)


koskimas:
Sorry those were left out on purpose.

igalklebanov:
> Sorry those were left out on purpose.

OK.. 😢

elderapo:
> Sorry those were left out on purpose.

Why?

koskimas:
We'd need to add 8 methods to at least 5 classes to have a good API. The API proposed here is not intuitive IMHO. Arrays are only used when you have a list of things. Here an array is used as a hack to pass two values.

Adding 40 methods for something that can be done using a simple `and where` is not a good tradeoff. Someone (me) needs to maintain the code once it's implemented.

igalklebanov:
> Arrays are only used when you have a list of things. Here an array is used as a hack to pass two values.

Could replace it with a `{ min, max }` or `{ lowerBound, upperBound }`..

But even then, not worth the added complexity. 👍

PR #119, opened 2022-07-18
adding fractional seconds precision to `db.schema.createTable(name).addColumn(...)`.

igalklebanov:
Fractional Seconds Precision (e.g. `timestamp(n)`) is supported in both [Postgresql](https://www.postgresql.org/docs/current/datatype-datetime.html) and [Mysql](https://dev.mysql.com/doc/refman/8.0/en/fractional-seconds.html) dialects, documented as optional *p* or *fsp* parameter.

Currently, kysely supports date and time data types, without stating *fsp*. Consumers are forced to use raw sql to achieve more accurate columns when creating tables using the schema module.

Database defaults and behavior on insert/update in case *fsp* is omitted are not good enough.

> A value of 0 signifies that there is no fractional part. If omitted, the default precision is 0. 
> 
> Inserting a TIME, DATE, or TIMESTAMP value with a fractional seconds part into a column of the same type but having fewer fractional digits results in **rounding**.  -mysql docs

Millisecond accuracy (at the very least) is crucial to some workflows (e.g. analytics, financial, event-driven systems).

----

tldr

This PR basically adds the ability to add date time columns as `data_type(fsp)`:
```typescript
`datetime(${number})` | `time(${number})` | `timetz(${number})` | `timestamp(${number})` | `timestamptz(${number})`
```

### DIFF
--- a/src/operation-node/data-type-node.ts
+++ b/src/operation-node/data-type-node.ts
@@ -22,10 +22,15 @@ export type ColumnDataType =
   | 'binary'
   | 'date'
   | 'datetime'
+  | `datetime(${number})`
   | 'time'
+  | `time(${number})`
   | 'timetz'
+  | `timetz(${number})`
   | 'timestamp'
+  | `timestamp(${number})`
   | 'timestamptz'
+  | `timestamptz(${number})`
   | 'serial'
   | 'bigserial'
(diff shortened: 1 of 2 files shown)


koskimas:
Thanks!

PR #124, opened 2022-07-22
add `.explain()` to supported statements.

igalklebanov:
`EXPLAIN` statement is supported in all 3 base dialects ([postgres](https://www.postgresql.org/docs/current/sql-explain.html), [mysql](https://dev.mysql.com/doc/refman/8.0/en/explain.html), [sqlite](https://www.sqlite.org/eqp.html)). Helpful when working on new designs and queries, or optimizing existing ones. 

When we abstract away our queries with query builders, such as Kysely, we are distancing ourselves from being able to debug and tinker with our designs and queries with ease. It suddenly requires us to either:

- copy-paste our Kysely code, replace `.execute()` with `.compile()`.
- add query logging in Kysely constructor, execute for real, copy-paste logged query.

and then add explain stuff manually, execute raw sql with `sql` component OR copy-paste logged sql into database client. Go back to migration code and modify or add stuff like indexes OR go back to query code and improve the query. And back and forth..

Feels like DX could be improved here.

What if we could just quickly comment out `.execute()` and replace it with `.explain()` whenever we want to have a look at the execution plan?

(while working on this, for a very long time `.explain()` was not an alternative to `.execute()`, but a part of the method chain before executing, but the typescript changes required to make its return type bubble up until the execute call, regardless of chain order [e.g. selectAll would extend it unless there's some generics flag telling it not to], are not worth it at all).

---

tldr

This PR adds the ability to add `explain` statement right before `select`/`insert`/`update`/`delete`/`replace` keywords.

This can be done by adding `.explain()` to your query (instead of `.execute()`):

```typescript
const explained = await db
    .selectFrom('person')
    .selectAll()
    .where('gender', '=', 'female')
    .explain('json')
```

```sql
explain format=json select * from `person` where `gender` = ?
```

You can also pass the expected row type `.explain<MysqlExplainJsonResult>()`


src/operation-node/operation-node-transformer.ts
```diff
@@ -351,6 +354,7 @@ export class OperationNodeTransformer {
       with: this.transformNode(node.with),
       ignore: node.ignore,
       replace: node.replace,
+      explain: node.explain,
```
koskimas, line 357: transformNode(node.explain)

src/operation-node/operation-node-transformer.ts
```diff
@@ -825,4 +831,9 @@ export class OperationNodeTransformer {
     // An Object.freezed leaf node. No need to clone.
     return node
   }
+
+  protected transformExplain(node: ExplainNode): ExplainNode {
+    // An Object.freezed leaf node. No need to clone.
+    return node
```
koskimas, line 837: `ExplainNode` isn't a leaf node as it has node children. You need to transform them.

src/plugin/explain/explain-plugin.ts
```diff
@@ -0,0 +1,37 @@
...
+export interface ExplainPluginOptions {
+  enabled?: boolean
+  format?: ExplainFormat
+  raw?: AnyRawBuilder
+}
+
+/**
+ * Plugin that adds `explain` statement before `select`/`insert`/`update`/`replace`/`delete` keywords.
+ */
+export class ExplainPlugin implements KyselyPlugin {
```
koskimas, line 21: I don't think this plugin is a good idea. It will lead to wrong output types and will basically break all queries that return data. I don't see this being useful.

src/query-builder/delete-query-builder.ts
```diff
@@ -642,6 +645,35 @@ export class DeleteQueryBuilder<DB, TB extends keyof DB, O>
...
+   *  .deleteFrom('person')
+   *  .where('id', '=', 123)
+   *  .explain('json')
+   * ```
+   *
+   * ```sql
+   * explain format=json delete from `person` where `id` = ?
+   * ```
+   */
+  async explain<ER extends Record<string, any> = Record<string, any>>(
```
koskimas, line 663: I think the default output type needs to be `unknown` right? Or do all dialects return a list of objects in all cases?

src/dialect/postgres/postgres-query-compiler.ts
```diff
@@ -5,6 +5,18 @@ export class PostgresQueryCompiler extends DefaultQueryCompiler {
     return '$' + this.numParameters
   }
 
+  protected override getLeftExplainOptionsWrapper(): string {
```
koskimas, line 8: The default query compiler is supposed to be for postgres. Use these as the defaults in `DefaultQueryCompiler`

koskimas: We should probably remove all these overrides here to make that more obvious.

koskimas:
This is a nice feature and a good addition. Thank you once again!

Ther separate `explain` method as an alternative to `execute` was a great design choice 💯 

I don't think the plugin is a good idea though. It only makes sense to use the plugin to globally enable explaining, but you can't actually run the code as all queries now return the explain data and not the actual data. Also all output types will be wrong.

src/operation-node/operation-node-transformer.ts
igalklebanov: Nice catch! done!

src/plugin/explain/explain-plugin.ts
igalklebanov: Yeah, on second thought, no one will ever investigate multiple queries at once. 👍 

r.i.p. ExplainPlugin. 💀

src/dialect/postgres/postgres-query-compiler.ts
igalklebanov: > We should probably remove all these overrides here to make that more obvious.

Good idea! done!

src/operation-node/operation-node-transformer.ts
igalklebanov: Nice catch! done!

igalklebanov:
Thanks! 😃

koskimas (approved):

PR #126, opened 2022-07-24
`FunctionBuilder` is too permissive in column argument. (issue #125)

### LINKED ISSUE #125: how to properly use "count" function builder with output type?
thanks a lot for this awesome library!

the "count" function builder seem to loose the "column" type-checking if an output type is being specified. 

```
import type { Generated } from "kysely";
import {
  DummyDriver,
  Kysely,
  SqliteAdapter,
  SqliteIntrospector,
  SqliteQueryCompiler,
} from "kysely";

interface Person {
  id: Generated<number>;
  age: number;
}

interface Database {
  person: Person;
}

const db = new Kysely<Database>({
  dialect: {
    createAdapter() {
      return new SqliteAdapter();
    },
    createDriver() {
      return new DummyDriver();
    },
    createIntrospector(db: Kysely<unknown>) {
      return new SqliteIntrospector(db);
    },
    createQueryCompiler() {
      return new SqliteQueryCompiler();
    },
  },
});

const qb = db
  .selectFrom("person")
  .select([
    b => b.fn.count("person.age").as("cnt1"), // compiles as expected, but output is string|bigint|int
    b => b.fn.count("person.age2").as("cnt2"), // does not compile as expected
    b => b.fn.count<number>("person.age2").as("cnt3"), // compiles NOT EXPECTED
    b => b.fn.count("person.age2").castTo<number>().as("cnt4"), // does not compile as expected
  ]);

```

i am not sure if "cnt3" usage is the intended approach - found it in some tests / docs.

the "cnt4" usage is more code but keeps the "column" type-checking.

please let me know which is the recommended intended approach.

ps:
the same happens for "avg" and "sum" function builder.

igalklebanov:
fixes #125

unless it was defaulting to `any` by design to not break/block some other behavior we want.. 🤷‍♂️


koskimas:
Nice catch! Thanks for fixing it 💯

PR #127, opened 2022-07-24
add `WithTablePrefixSuffixPlugin`.

igalklebanov:


koskimas:
I don't think this should be added to kysely core. I've never heard anyone using this kind of pre/suffixing, and I personally think it's a horrible idea to have different table names in different environments.

I disagree with the notion that we should add everything someone might use. I think the opposite is true. We should add the minimum amount of features that make most use cases possible.

It's not just the matter of implementing the feature. We need to maintain the code, usually "forever". Added code also makes the bundle larger.

It's not easy to know which features are worth implementing. Usually I wait until at least couple of people ask for it.

PR #151, opened 2022-08-30
Coerce file migration from esModuleInterop default prop.

waynebloss:
Add a check for a `default` prop when dynamically importing file migrations.

The `FileMigrationProvider` wasn't working in my project which is compiled to `commonjs` and has `esModuleInterop` enabled. So, I edited my local Kysely dependency to add some logging around the imported migration. I saw that the imports where wrapped like this: `{ default: { up: Function, down: Function } }`, so I changed the code to coerce the value from the default property like this:

```js
let migration = await Promise.resolve().then(() => require(
/* webpackIgnore: true */ this.#props.path.join(this.#props.migrationFolder, fileName)));
migration = migration?.default ?? migration;
```

All existing tests have passed for me. I did not write any new tests surrounding this change because I didn't think it was warranted, but if it's required for acceptance please let me know. Thanks!

(I was on the fence about the one comment line that I added. Do you think it's self explanatory?)

### DIFF
--- a/src/migration/file-migration-provider.ts
+++ b/src/migration/file-migration-provider.ts
@@ -33,12 +33,14 @@ export class FileMigrationProvider implements MigrationProvider {
         (fileName.endsWith('.js') || fileName.endsWith('.ts') || fileName.endsWith('.mjs')) &&
         !fileName.endsWith('.d.ts')
       ) {
-        const migration = await import(
+        let migration = await import(
           /* webpackIgnore: true */ this.#props.path.join(
             this.#props.migrationFolder,
             fileName
           )
         )
+        // Coerce an esModuleInterop export's `default` prop...
+        migration = migration?.default ?? migration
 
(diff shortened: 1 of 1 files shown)


waynebloss:
One mistake that I made when I initially created my Migration files was that I exported a default object with `up/down` function fields instead of just exporting 2 named functions `up/down`.

Now that I changed the export style of my files, the original `FileMigrationProvider` works just fine.

Perhaps you will still want this change though, so I'll leave it here.

koskimas:
To play it safe, you could check `isMigration(migration?.default)` before coercing. Someone could in theory export some unrelated stuff as default by accident or otherwise.

waynebloss:
OK, I changed it a bit to avoid a duplicate call to `isMigration`. If you don't want that, I'll do it the other way - I'm not too concerned about performance for migrations, but I thought it made more sense this way...

PR #153, opened 2022-09-03
AggregateFunctionBuilder, distinct & over clause support (window functions).

igalklebanov:
- [x] Implement `AggregateFunctionBuilder` & `OverBuilder`. Supporting `distinct` clause (e.g. `count(distinct name)`) and `over` clause (window functions) [should be supported in all 3 base dialects].
- [x] Rename `FunctionBuilder` to `FunctionModule` to align with `SchemaModule`.
- [x] Implement `AggregateFunctionNode`, `OverNode`, `PartitionByNode`, `PartitionByItemNode`. 
- [x] Implement parsing and compilation of new nodes.
- [x] Fix `select` test type errors.
- [X] Fix `having` test type errors.
- [X] Add unit tests for new `distinct` and `over` functionality.
- [X] Add typings tests.
- [X] Write tsdocs for new functionality.


src/query-builder/aggregate-function-builder.ts
```diff
@@ -0,0 +1,114 @@
...
+{
+  readonly #aggregateFunctionBuilder: AggregateFunctionBuilder<DB, TB, O>
+  readonly #alias: A
+
+  /**
+   * @private
+   *
+   * This needs to be here just so that the typings work. Without this
+   * the generated .d.ts file contains no reference to the type param A
+   * which causes this type to be equal to AliasedRawBuilder with any A
```
koskimas, line 80: Comment refers to `AliasedRawBuilder`

src/query-builder/expression-builder.ts
```diff
@@ -58,8 +58,8 @@ export class ExpressionBuilder<DB, TB extends keyof DB> {
    * having count("pet"."id") > $1
    * ```
    */
-  get fn(): FunctionBuilder<DB, TB> {
-    return new FunctionBuilder()
+  get fn(): FunctionModule<DB, TB> {
+    return new FunctionModule()
```
koskimas, line 62: Nice change!

src/query-builder/function-module.ts
```diff
@@ -39,52 +38,13 @@ import { createQueryId } from '../util/query-id.js'
  * having count("pet"."id") > $1
  * ```
  */
-export class FunctionBuilder<DB, TB extends keyof DB> {
+export class FunctionModule<DB, TB extends keyof DB> {
```
koskimas, line 41: I think you can now remove the `#oneArgFunction` method.

src/parser/select-parser.ts
```diff
@@ -76,6 +77,8 @@ type ExtractAliasFromSelectExpression<SE> = SE extends string
   ? QA
   : SE extends DynamicReferenceBuilder<infer RA>
   ? ExtractAliasFromStringSelectExpression<RA>
+  : SE extends AliasedAggregateFunctionBuilder<any, any, any, infer QA>
```
koskimas, line 80: Don't you also need the callback version here?

We need to be careful to not add too many types to these types. The next time a new type needs to be added here, maybe we can think of ways to combine some of these.

src/parser/select-parser.ts
```diff
@@ -118,6 +121,8 @@ type ExtractTypeFromSelectExpression<
   ? A extends ExtractAliasFromStringSelectExpression<RA>
     ? ExtractTypeFromStringSelectExpression<DB, TB, RA, A> | undefined
     : never
+  : SE extends AliasedAggregateFunctionBuilder<any, any, infer O>
```
koskimas, line 124: Same here, also needs the callback version.

src/query-builder/aggregate-function-builder.ts
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  toOperationNode(): AliasNode {
+    return AliasNode.create(
+      this.#aggregateFunctionBuilder.toOperationNode(),
+      IdentifierNode.create(this.#alias)
+    )
+  }
+
+  constructor(
```
koskimas, line 94: Move the constructor before getters and methods.

koskimas:
Really nice addition! Will definitely merge this. My only concern is adding too many types to the already complex `select-parser` types. If we keep getting these new builders we need to refactor the types somehow to keep them clean. But let's see how they evolve in the future.

src/query-builder/aggregate-function-builder.ts
```diff
@@ -0,0 +1,114 @@
...
+import { AggregateFunctionNode } from '../operation-node/aggregate-function-node.js'
+import { OperationNodeSource } from '../operation-node/operation-node-source.js'
+import { AliasNode } from '../operation-node/alias-node.js'
+import { IdentifierNode } from '../operation-node/identifier-node.js'
+import { preventAwait } from '../util/prevent-await.js'
+import { OverBuilder } from './over-builder.js'
+import { createOverBuilder } from '../parser/parse-utils.js'
+
+export class AggregateFunctionBuilder<DB, TB extends keyof DB, O = unknown>
+  implements OperationNodeSource
```
koskimas, line 11: Is the output type `O` actually needed here? Does something use it?

koskimas: It is, never mind.

src/query-builder/aggregate-function-builder.ts
igalklebanov: Good catch! done.

src/query-builder/aggregate-function-builder.ts
igalklebanov: Good catch! done.

src/query-builder/expression-builder.ts
igalklebanov: Thanks! Was a bit scared of this, since this is a breaking change (if anything ever used the type directly).

src/query-builder/function-module.ts
igalklebanov: I'm confused, I swear I did this.. it doesn't exist on my end. 😕

src/parser/select-parser.ts
igalklebanov: > Don't you also need the callback version here?

Good catch! done.

> We need to be careful to not add too many branches to these types. The next time a new type needs to be added here, maybe we can think of ways to combine some of these.

Agreed!

Is this a maintainability concern and/or typescript stability concern?

src/parser/select-parser.ts
igalklebanov: Good catch! done.

src/parser/select-parser.ts
koskimas: At least maintainability, maybe typescript speed/stability too. Haven't run any tests on that.

src/query-builder/aggregate-function-builder.ts
```diff
@@ -0,0 +1,114 @@
...
+      aggregateFunctionNode: AggregateFunctionNode.cloneWithDistinct(
+        this.#props.aggregateFunctionNode
+      ),
+    })
+  }
+
+  /**
+   * TODO: ...
+   */
+  over(
```
koskimas, line 43: Does `over()` without any arguments produce meaningful SQL?

test/node/src/aggregate-function.test.ts
```diff
@@ -0,0 +1,301 @@
...
+      await destroyTest(ctx)
+    })
+
+    for (const func of ['avg', 'count', 'max', 'min', 'sum']) {
+      describe(func, () => {
+        it(`should execute a query with ${func}(...) in select clause`, async () => {
+          const query = ctx.db
+            .selectFrom('person')
+            .select([
+              (ctx.db.fn[func as keyof FunctionModule<any, any>] as Function)(
```
koskimas, line 28: I don't think you need these casts if you iterate over

```ts
for (const func of ['avg', 'count', 'max', 'min', 'sum'] as const) {
```

src/query-builder/aggregate-function-builder.ts
finaloop-igal: This use case is mentioned in official documentations ([postgres](https://www.postgresql.org/docs/current/tutorial-window.html), [mysql](https://dev.mysql.com/doc/refman/8.0/en/window-functions-usage.html)).

Basically "let aggregate function do its thing, but duplicate it across all rows and return all rows instead of a single row".

> The first OVER clause is empty, which treats the entire set of query rows as a single partition. The window function thus produces a global sum, but does so for each row.

test/node/src/aggregate-function.test.ts
igalklebanov: Good catch! 

This removes the need for `as keyof FunctionModule`.
But ts still throws an error due to the key space not having a uniform function signature. 😐

koskimas: Don't know how to fix that one off the top of my head

PR #163, opened 2022-09-16
add `intersect` & `except` support.

igalklebanov:
- [x] generalize `UnionNode`, its parser, transforms and visits - all set operators share the same syntax of `query1 SET_OPERATOR [all | distinct] query2`. Also set operations can be combined, e.g. `query1 union query2 except query3`, so generalize `SelectQueryNode.union` field as well.
- [x] add `intersect`, `except`, ~`minus`~  (not supported in base 3 dialects - removed)  to `SelectQueryBuilder`.
- [x] add unit tests for `intersect`, `except`, ~`minus`~ and their combinations including `union` in the same query.
- [x] add typings tests. 

not sure about:

- [x] add `intersectAll` to `SelectQueryBuilder` - supported in postgres, keeps duplicates.
- [x] add `exceptAll` to `SelectQueryBuilder` - supported in postgres, keeps duplicates.


src/operation-node/set-operator-node.ts
```diff
@@ -0,0 +1,37 @@
+import { freeze } from '../util/object-utils.js'
+import { OperationNode } from './operation-node.js'
+import { RawNode } from './raw-node.js'
+import { SelectQueryNode } from './select-query-node.js'
+
+export type SetOperatorExpressionNode = SelectQueryNode | RawNode
+
+export type SetOperator = 'union' | 'interect' | 'except' | 'minus'
+
+export interface SetOperatorNode extends OperationNode {
```
koskimas, line 10: I'd rename this to `SetOperationNode` instead since it describes the whole operation and not just the operator. Otherwise all this looks great!

igalklebanov: Good point!

renamed.

igalklebanov:
- removed minus, as its not supported in mysql (my bad).
- added intersectAll & exceptAll - supported in postgres and low hanging fruit.
- simplified typings folder structure to work with defaults.

koskimas (approved):
LGTM! Awesome work once again ❤️

PR #169, opened 2022-09-23
add `DynamicReferenceBuilder` support @ `FunctionModule`.

igalklebanov:
- [x] allow `DynamicReferenceBuilder` to be passed at `FunctionModule`'s `avg`, `count`, `max`, `min`, `sum`.
- [x] allow passing a default value type to reference expression extract type helper, so it doesn't default to `unknown` all the time. why? aggregate functions return `string | number | bigint`.
- [x] allow passing a return type generic at `FunctionModule`'s `max` & `min`, but only when column is a dynamic reference. regular reference exists in `db` so we can extract type with ease - allowing consumer to control this will contradict the `SelectType` she passed to `Kysely` constructor generic.
- [x] typings tests
- [x] unit tests


koskimas (approved):
Excellent! 💯

igalklebanov:
ready to go! 🥳

koskimas:
Sorry, this now needs a rebase. I'll merge this after that ASAP.

igalklebanov:
@koskimas let me know if its still the case, cause UI stopped warning about conflicts for me.

Issue #170, opened 2022-09-23
Updateable<T> allow undefined?


bestickley:
Why cannot Updateable<T> allow `undefined`?

I'm using zod for input validation. For an update, all properties besides an id should be optional. However, this seems to not allow me to use `.set` within Kysely because it does not allow undefined.

Use case:
```ts
import { Kysely, KyselyConfig, Updateable } from "kysely";
import { z } from "zod";

interface UserTable {
  id: number;
  firstName: string | null;
  lastName: string | null;
}
interface Database {
  user: UserTable;
}
const db = new Kysely<Database>({} as KyselyConfig);

const userSchema = z.object({
  id: z.number(),
  firstName: z.string().optional(),
  lastName: z.string().optional(),
});

const user = userSchema.parse({
  id: 1,
  firstName: "Ben",
});

// doesn't work
db.updateTable("user").set(user).where("id", "=", 1).execute();
// works
db.updateTable("user")
  .set(user as Updateable<UserTable>)
  .where("id", "=", 1)
  .execute();

```

Error:
<img width="905" alt="image" src="https://user-images.githubusercontent.com/35735118/191946482-c0c0b356-1479-46af-841b-96aaaf01a85f.png">

bestickley:
Found a work around updating my tsconfig to:
```json
{
  "extends": "@tsconfig/node16-strictest-esm/tsconfig.json",
  "compilerOptions": {
    "exactOptionalPropertyTypes": false // UPDATE HERE
  },
}
```
but I'd prefer to keep all strict TS checks on if possible. I'd still like this library to consider allowing `Updateable` to accept `undefined`

igalklebanov:
Try adjusting `UserTable` as follows:

```ts
interface UserTable {
  id: GeneratedAlways<number>;
  firstName: ColumnType<string | null, string | undefined, string | undefined>;
  lastName: ColumnType<string | null, string | undefined, string | undefined>;
}
```

bestickley:
@igalklebanov, so in the docs when it says:

// If the column is nullable in the database, make its type nullable.
// Don't use optional properties. Optionality is always determined
// automatically by Kysely.

it's saying don't use `{ key?: string }` but you **can** use `{ key: string | undefined }`?

igalklebanov:
`Kysely` is not developed with such strict tsconfig, which is another topic of discussion.

I've found a fix for this issue, and it doesn't seem to break other things, so I'm pushing it.

bestickley:
[...]

PR #171, opened 2022-09-23
allow `undefined` when inserting/updating if column is nullable, even in stricter environments.

igalklebanov:
relevant to #170


koskimas:
This is already handled in some place since it works when with typical typescript settings. This adds the handing in another place. I'll see if we can combine the two places to keep things simpler.

PR #173, opened 2022-09-24
pass `queryId` to `executeQuery` to allow easy async comms between connections & plugins

igalklebanov:
Currently the only components enjoying comms by `queryId` are plugins between their `transformX` methods and compilers. Seems a bit odd that `connection.executeQuery` was left behind.

While playing around with casting results to js land types in some data api implementation, I couldn't communicate some elements from a query response to a plugin's `transformResult` method (connection and plugin belong to the same package so it's fine). This restriction forced me to implement an overly complex hack that leaks. Having queryId in the connection could allow an easy, garbage collectable, WeakMap flow.

---

Wanted flow that's currently impossible:

connection:

```ts
//...

executeQuery(compiledQuery, queryId) {
  const result = ...

  someWeakMap.set(queryId, result.someUncommonFieldInSomeJsonWeWontStandardize)

  return {
    rows: result.rows
  }
}

// ...
```

plugin:

```ts
transformResult(args) {
  const someUncommonFieldInSomeJsonWeWontStandardize = someWeakMap.get(args.queryId)

  // do stuff..
}
```

### DIFF
--- a/src/dialect/mysql/mysql-driver.ts
+++ b/src/dialect/mysql/mysql-driver.ts
@@ -5,6 +5,7 @@ import {
 import { Driver, TransactionSettings } from '../../driver/driver.js'
 import { CompiledQuery } from '../../query-compiler/compiled-query.js'
 import { isFunction, isObject, freeze } from '../../util/object-utils.js'
+import { createQueryId, QueryId } from '../../util/query-id.js'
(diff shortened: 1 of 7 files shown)


koskimas:
Could the queryId be a property in the compiledQuery interface? I didn't think about this too much, so it could be a stupid idea.

igalklebanov:
Should work, trying it out now.

PR #176, opened 2022-09-27
add `queryId` to `CompiledQuery` to allow async communication between more components.

igalklebanov:
does what #173 tried to achieve, but in a simpler way as suggested by @koskimas.

- adds `queryId` to `CompileQuery` - now `DatabaseConnection` has access to it in `executeQuery` and `streamQuery` methods in a non-breaking way.
- adds `queryId` as a 2nd optional (for now) argument in all `transform` methods. Now plugins can enjoy even more flexibility - not just in root `transformQuery` and `transformResults` methods.
- exposes `createQueryId` to ease dialect/plugin implementation by 3rd parties.
- also removes `numUpdatedOrDeletedRows` usage as it has been deprecated for 2 years now.

### DIFF
--- a/src/dialect/mssql/mssql-driver.ts
+++ b/src/dialect/mssql/mssql-driver.ts
@@ -339,7 +339,6 @@ class MssqlRequest<O> {
     (event: 'completed' | 'chunkReady' | 'error', error?: unknown) => void
   >
   readonly #tedious: Tedious
-  #error: Error | any[] | undefined
   #rowCount: number | undefined
 
   constructor(props: MssqlRequestProps<O>) {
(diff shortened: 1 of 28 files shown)


alexgleason:
Thank you for continuing to update this branch @igalklebanov 

What do you think of this PR @koskimas? Are you going to merge it? This issue has been open for 2 years. I understand not wanting to break things, because it will be a pain to deal with. I am an open source maintainer, I get it. But serious projects really need this. It is essentially broken already to me, in its current state. Because my services are doing this constantly and I don't really have a good way to deal with it in Kysely. I feel backed into a corner.

image

I have spent the past two weeks gathering data to try to fix this. I came up with this graph (query duation):

image

And it is very symetrical with this graph (available connections):

image

That causes downtime. I need to cancel queries after a timeout.

I am planning to try some postgres workarounds, like using `set statement_timeout` in a transaction. But that is a hack. I would really like a real solution. Thank you.

PR #179, opened 2022-09-30
add `db.fn.coalesce(...)`

igalklebanov:
- [x] implement coalesce type extract.
- [x] implement coalesce parser.
- [x] implement FunctionModule.coalesce(value, ...otherValues).
- [x] unit tests.
- [x] typings tests.
- [x] FunctionModule.coalesce ts docs.


igalklebanov:
Don't know if we should go for a new builder and node, so went with raw for now.

If we go for new builder and node.. not sure what terminology to use..

In mysql docs it is grouped with [comparison operators & functions](https://dev.mysql.com/doc/refman/8.0/en/comparison-operators.html#function_coalesce).
In postgres docs it is grouped with [conditional expressions](https://www.postgresql.org/docs/current/functions-conditional.html#FUNCTIONS-COALESCE-NVL-IFNULL)
In sqlite docs it is grouped with [built-in scalar functions](https://www.sqlite.org/lang_corefunc.html#coalesce)

koskimas:
This is good to go once you rebase! Nice work once again 👍

igalklebanov:
Thanks! Seems ready now.. also went over 5c2c190 (👏 btw ) just to be sure.

PR #183, opened 2022-10-06
make `FunctionModule`'s `avg`, `max`, `min` & `sum` functions nullable (consumer opt-in).

igalklebanov:
These aggregate functions are nullable, in some edge cases.

Current types:

a. are a lie, if some edge cases are possible in consumer's queries / state of database.
b. do not allow our consumers to state `null` as a possible return type.

Let's help our consumers write robust code, by allowing them to opt-in for the stricter nullable form.

Should fit smooth like 🧈 with recently merged `.coalesce(...)` function.

----

- [x] implement type changes.
- [x] fix typings tests.


src/query-builder/function-module.ts
```diff
@@ -142,14 +145,19 @@ export class FunctionModule<DB, TB extends keyof DB> {
...
+    >
   >(
     column: C
   ): AggregateFunctionBuilder<
     DB,
     TB,
-    ExtractTypeFromReferenceExpression<DB, TB, C, O>
+    ExtractTypeFromReferenceExpression<DB, TB, C, O> | null extends O
+      ? null
+      : never
```
igalklebanov, line 160: Current types allow consumers to state `O` only when column is a `DynamicReference`. Let's open this up, so the consumer can tell us whether `O` is nullable, even when column is a `StringReference`, but let's still restrict return type to the type provided for that column in `Database`.

src/query-builder/function-module.ts
```diff
@@ -211,9 +225,12 @@ export class FunctionModule<DB, TB extends keyof DB> {
    *   .select(sum<number>('price').as('total_price'))
    *   .execute()
    * ```
+   *
+   * It is highly recommended to include null in the output type union and handle
+   * null values in post-execute code, or wrap the function with a coalesce function.
    */
   sum<
-    O extends number | string | bigint,
+    O extends number | string | bigint | null = number | string | bigint,
```
igalklebanov, line 233: User can opt-in for nullable `O`.

src/query-builder/function-module.ts
```diff
@@ -60,14 +60,15 @@ export class FunctionModule<DB, TB extends keyof DB> {
    * Calls the `avg` function for the column given as the argument.
    *
    * If this is used in a `select` statement the type of the selected expression
-   * will be `number | string` by default. This is because Kysely can't know the
-   * type the db driver outputs. Sometimes the output can be larger than the
-   * largest javascript number and a string is returned instead. Most drivers
-   * allow you to configure the output type of large numbers and Kysely can't
-   * know if you've done so.
+   * will be `number | string | null` by default. This is because Kysely can't know
```
koskimas, line 63: This isn't true is it? The default is now `number | string` with a user opt-in of including the `null`?

src/query-builder/function-module.ts
```diff
@@ -143,11 +155,11 @@ export class FunctionModule<DB, TB extends keyof DB> {
    * Calls the `count` function for the column given as the argument.
    *
    * If this is used in a `select` statement the type of the selected expression
-   * will be `number | string | bigint` by default. This is because Kysely can't
-   * know the type the db driver outputs. Sometimes the output can be larger than
-   * the largest javascript number and a string is returned instead. Most drivers
-   * allow you to configure the output type of large numbers and Kysely can't
-   * know if you've done so.
+   * will be `number | string | bigint | null` by default. This is because Kysely
```
koskimas, line 158: Same here? Default is actually `number | string | bigint`.

src/query-builder/function-module.ts
```diff
@@ -250,11 +285,12 @@ export class FunctionModule<DB, TB extends keyof DB> {
    * Calls the `sum` function for the column given as the argument.
    *
    * If this is used in a `select` statement the type of the selected expression
-   * will be `number | string | bigint` by default. This is because Kysely can't
-   * know the type the db driver outputs. Sometimes the output can be larger than
-   * the largest javascript number and a string is returned instead. Most drivers
-   * allow you to configure the output type of large numbers and Kysely can't
-   * know if you've done so.
+   * will be `number | string | bigint | null` by default. This is because Kysely
```
koskimas, line 288: Here too, null is not in the default.

src/query-builder/function-module.ts
igalklebanov: Good catch! forgot about aligning the docs.. 😢

src/query-builder/function-module.ts
igalklebanov: Good catch! forgot about aligning the docs.. 😢

src/query-builder/function-module.ts
igalklebanov: Good catch! forgot about aligning the docs.. 😢

igalklebanov:
Sorry about docs, hope its fine now.. 😅

koskimas:
Thanks!

PR #188, opened 2022-10-09
add `numInsertedOrUpdatedRows` to `InsertResult`.

igalklebanov:
Also:

- Deprecate `QueryResult.numUpdatedOrDeletedRows` and add `QueryResult.numAffectedRows`.

---

- [x] add to `InsertResult`.
- [x] unit tests.
~~- [ ] typings tests.~~ tests for `InsertResult` already in place.

---

On MySQL, this field lets consumers know whether an `insert...on duplicate key update` query inserted OR updated some records. Real world scenario, publishing an event following the query, "entity {created | updated}".

> With ON DUPLICATE KEY UPDATE, the affected-rows value per row is 1 if the row is inserted as a new row, 2 if an existing row is updated, and 0 if an existing row is set to its current values. If you specify the CLIENT_FOUND_ROWS flag to the mysql_real_connect() C API function when connecting to mysqld, the affected-rows value is 1 (not 0) if an existing row is set to its current values.
https://dev.mysql.com/doc/refman/8.0/en/insert-on-duplicate.html

The field is also relevant to inserts in Postgres..

> The rowCount is populated from the command tag supplied by the PostgreSQL server. It's generally of the form: COMMAND [OID] [ROWS]. For DML commands (INSERT, UPDATE, etc), it reflects how many rows the server modified to process the command.
https://node-postgres.com/api/result

---

Things to consider:

Feels like `QueryResult`'s `numUpdatedOrDeletedRows` should be renamed to `numAffectedRows`, which is more readable, neutral and aligned with MySQL terminology.

~~Stopped myself from doing this change as it'll break 3rd party drivers.~~

### DIFF
--- a/src/dialect/mysql/mysql-driver.ts
+++ b/src/dialect/mysql/mysql-driver.ts
@@ -129,7 +129,7 @@ class MysqlConnection implements DatabaseConnection {
               ? BigInt(insertId)
               : undefined,
           numUpdatedOrDeletedRows:
-            affectedRows !== undefined && insertId !== null
+            affectedRows !== undefined && affectedRows !== null
               ? BigInt(affectedRows)
               : undefined,
           rows: [],
--- a/src/dialect/mysql/mysql-driver.ts
+++ b/src/dialect/mysql/mysql-driver.ts
@@ -121,6 +121,11 @@ class MysqlConnection implements DatabaseConnection {
       if (isOkPacket(result)) {
         const { insertId, affectedRows } = result
 
+        const numAffectedRows =
+          affectedRows !== undefined && affectedRows !== null
+            ? BigInt(affectedRows)
+            : undefined
+
         return {
           insertId:
             insertId !== undefined &&
@@ -128,10 +133,9 @@ class MysqlConnection implements DatabaseConnection {
             insertId.toString() !== '0'
               ? BigInt(insertId)
               : undefined,
-          numUpdatedOrDeletedRows:
-            affectedRows !== undefined && affectedRows !== null
-              ? BigInt(affectedRows)
-              : undefined,
+          // TODO: remove.
+          numUpdatedOrDeletedRows: numAffectedRows,
+          numAffectedRows,
           rows: [],
         }
       } else if (Array.isArray(result)) {
--- a/src/driver/database-connection.ts
+++ b/src/driver/database-connection.ts
@@ -15,10 +15,16 @@ export interface DatabaseConnection {
 
 export interface QueryResult<O> {
   /**
+   * @deprecated use {@link QueryResult.numAffectedRows} instead.
+   */
+  // TODO: remove.
+  readonly numUpdatedOrDeletedRows?: bigint
+
+  /**
    * This is defined for insert, update and delete queries and contains
    * the number of rows the query inserted/updated/deleted.
    */
-  readonly numUpdatedOrDeletedRows?: bigint
+  readonly numAffectedRows?: bigint
 
   /**
    * This is defined for insert queries on dialects that return
--- a/src/driver/database-connection.ts
+++ b/src/driver/database-connection.ts
@@ -15,8 +15,8 @@ export interface DatabaseConnection {
 
 export interface QueryResult<O> {
   /**
-   * This is defined for update and delete queries and contains
-   * the number of rows the query updated/deleted.
+   * This is defined for insert, update and delete queries and contains
+   * the number of rows the query inserted/updated/deleted.
    */
   readonly numUpdatedOrDeletedRows?: bigint
 
--- a/src/query-executor/query-executor-base.ts
+++ b/src/query-executor/query-executor-base.ts
@@ -14,6 +14,9 @@ import { Deferred } from '../util/deferred.js'
 
 const NO_PLUGINS: ReadonlyArray<KyselyPlugin> = freeze([])
 
+// TODO: remove.
+let warnedOfDeprecation: boolean
+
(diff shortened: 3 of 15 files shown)


koskimas:
`numAffectedRows` needs a comment that it's only defined for update and delete queries. People might expect it to contain the number of selected and inserted rows too.

src/driver/database-connection.ts
```diff
@@ -15,8 +15,8 @@ export interface DatabaseConnection {
 
 export interface QueryResult<O> {
   /**
-   * This is defined for update and delete queries and contains
-   * the number of rows the query updated/deleted.
+   * This is defined for insert, update and delete queries and contains
+   * the number of rows the query inserted/updated/deleted.
    */
   readonly numUpdatedOrDeletedRows?: bigint
```
koskimas, line 21: You're right, this should now be renamed to `numAffectedRows` now. I sucks that this will break 3rd party dialects, but we could give a heads up for the biggest one (SST and data api). 

We could even go so far as to handle the old property `numUpdatedOrDeletedRows` internally using ugly `any` casts. Then the old dialects would _mostly_ work but the maintainers would get type errors about the missing `numUpdatedOrDeletedRows` when they update their dialects. What do you think? Or is it better to just break it and expedite the update? We'll bump the version to indicate a breaking change of course.

test/node/src/insert.test.ts
```diff
@@ -485,12 +527,72 @@ for (const dialect of BUILT_IN_DIALECTS) {
...
+      }
+
+      const inserted = await ctx.db
+        .selectFrom('person')
+        .selectAll()
+        .orderBy('id', 'desc')
+        .limit(2)
+        .execute()
+
+      expect(inserted).to.have.length(2)
```
koskimas, line 587: Surely the length is two since you use `limit(2)` 😄

src/driver/database-connection.ts
koskimas: We could even open a PR [here](https://github.com/serverless-stack/kysely-data-api)

koskimas: If we detect the usage of `numUpdatedOrDeletedRows` we could have a one-time warning log indicating that the dialects is outdated?

igalklebanov: > We could even go so far as to handle the old property numUpdatedOrDeletedRows internally using ugly any casts. Then the old dialects would mostly work but the maintainers would get type errors about the missing numUpdatedOrDeletedRows when they update their dialects. What do you think? Or is it better to just break it and expedite the update? We'll bump the version to indicate a breaking change of course.

Our heaviest dependant is SST. Kysely is a hard dependency @^0.21.3 in their core package, and a peer in data-api @0.x.
Another potential heavy dependant is kysely-planetscale (100+ stars, used by an influencer with a big community). Kysely is a peer @*.

We could break things in the next major, or deprecate+warn and postpone break to major after next. Both are fine.
Either way, I'd go the extra mile and immediately open a pull request @ kysely-data-api and kysely-planetscale.

test/node/src/insert.test.ts
igalklebanov: Whoops! 🤣

src/dialect/mysql/mysql-driver.ts
```diff
@@ -129,7 +129,7 @@ class MysqlConnection implements DatabaseConnection {
               ? BigInt(insertId)
               : undefined,
           numUpdatedOrDeletedRows:
-            affectedRows !== undefined && insertId !== null
```
koskimas, line 132: Oh shit, this has been broken hasn't it 😬

igalklebanov: I suspect it was. Had a "wtf" moment when I went over it for this PR. 😟

igalklebanov:
- Added `numAffectedRows` to `QueryResult`. Added `@deprecated` to `numUpdatedOrDeletedRows` jsdoc.
- Added one-time outdated driver / plugin detection and "deprecation warning".

src/query-executor/query-executor-base.ts
```diff
@@ -65,7 +68,13 @@ export abstract class QueryExecutorBase implements QueryExecutor {
   ): Promise<QueryResult<R>> {
     return await this.provideConnection(async (connection) => {
       const result = await connection.executeQuery(compiledQuery)
-      return this.#transformResult(result, queryId)
+
+      const transformedResult = await this.#transformResult(result, queryId)
+
+      // TODO: remove.
+      warnOfDeprecatedDriverOrPlugins(result, transformedResult)
```
koskimas, line 75: I think we should add a helper function like `logDeprected` (or something) in its own module that keeps a `Set<string>` of deprection messages that have been shown.

src/query-executor/query-executor-base.ts
```diff
@@ -14,6 +14,9 @@ import { Deferred } from '../util/deferred.js'
 
 const NO_PLUGINS: ReadonlyArray<KyselyPlugin> = freeze([])
 
+// TODO: remove.
+let warnedOfDeprecation: boolean
```
koskimas, line 18: I think we should add a helper function like logDeprected (or something) in its own module that keeps a Set<string> of deprection messages that have been shown.

koskimas: Or even more generic `logOnce`

igalklebanov: Roger!

koskimas (approved):

PR #200, opened 2022-10-15
fix(types): make `sql.ids(...)` & `sql.join(...)` arguments readonly

ethanresnick:
`sql.join` and `sql.ids` don't mutate their arguments, so those arguments should be `readonly`, so that readonly arrays can be passed in without TS complaining.


koskimas:
Thanks!

PR #202, opened 2022-10-22
fix no separator between default insert values.

igalklebanov:
Currently, having more than 1 missing column value in `insertInto` value compiles to:

`("hello", "world", nullnull)`

instead of

`("hello", "world", null, null)`


koskimas (approved):
Nice find!

PR #206, opened 2022-10-26
add `Infer<typeof query>`, `Infer<typeof compiledQuery>`

igalklebanov:
Attempting a cheaper approach to #186

### DIFF
--- a/src/query-builder/delete-query-builder.ts
+++ b/src/query-builder/delete-query-builder.ts
@@ -53,7 +53,7 @@ export class DeleteQueryBuilder<DB, TB extends keyof DB, O>
     WhereInterface<DB, TB>,
     ReturningInterface<DB, TB, O>,
     OperationNodeSource,
-    Compilable,
+    Compilable<O>,
     Explainable
 {
   readonly #props: DeleteQueryBuilderProps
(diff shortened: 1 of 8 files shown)


koskimas:
This seems like a good addition, but I'm not sure about the name `Infer`. It seems too generic. It's not at all clear what it will infer without reading the docs. How about `InferResult` or `InferResultType`?

igalklebanov:
I totally agree.

The `Infer` naming comes from packages like `zod`/`runtypes`/`typebox`. There it makes sense since the consumer defines a structure and immediately extract its type. It's not ambiguous in a data definition context.

Here, its not obvious to the consumer, so we better have a more precise name. I find `InferResult` to be best candidate.

src/util/infer-result.ts
```diff
@@ -0,0 +1,54 @@
...
+ *   .innerJoin('pet', 'pet.owner_id', 'person.id')
+ *   .select(['person.first_name', 'pet.name'])
+ *
+ * type QueryResult = InferResult<typeof query> // { first_name: string; name: string; }[]
+ * ```
+ *
+ * Infer a compiled query's result type:
+ *
+ * ```ts
+ * import { Infer } from 'kysely'
```
koskimas, line 29: `Infer` --> `InferType`

koskimas:
Found usage of `Infer` in a comment. Once fixed, I'll merge this. Thanks again! You're awesome!

igalklebanov:
Let's go! 🚀

PR #208, opened 2022-10-28
add `filter` clause support at `AggregateFunctionBuilder`.

igalklebanov:
brought up by @thelinuxlich on discord.

Postgres [[1]](https://www.postgresql.org/docs/current/sql-expressions.html#SYNTAX-AGGREGATES) [[2]](https://www.postgresql.org/docs/current/sql-expressions.html#SYNTAX-WINDOW-FUNCTIONS) & [SQLite](https://www.sqlite.org/windowfunctions.html#the_filter_clause) support `filter` clause when using aggregate functions, with syntax as follows:
```
aggregate_function(...) [filter(where ...)] [over(...)]
```

Its a simple alternative to using `case...when...then...end` inside the aggregate function.

This PR adds `filterWhere`, `filterWhereExists`, `filterWhereNotExists`, `filterWhereRef`, `orFilterWhere`, `orFilterWhereExists`, `orFilterWhereNotExists` & `orFiltereWhereRef` to `AggregateFunctionBuilder`.

---

- [x] implement.
- [x] unit tests.
- [x] typings tests.
- [x] ts docs.

---

Some things to consider..

- since `where` is nested in `filter`, maybe its more intuitive and future-proof to make a `FilterBuilder` that implements `WhereInterface`, and make `filter` accept a `FilterBuilderCallback` instead of acting as the `where` itself?
- ~maybe we should rename `filter` to `filterWhere` to not hide `where` from caller?~ went with it, feels more aligned with Kysely's API philosophy.
- since `filter` clause is a thing, maybe we should rename the existing `FilterNode`, parsers, etc and introduce a new `FilterNode`?

### DIFF
--- a/src/operation-node/aggregate-function-node.ts
+++ b/src/operation-node/aggregate-function-node.ts
@@ -50,7 +50,11 @@ export const AggregateFunctionNode = freeze({
     return freeze({
       ...aggregateFunctionNode,
       filter: aggregateFunctionNode.filter
-        ? WhereNode.cloneWithFilter(aggregateFunctionNode.filter, 'And', filter)
+        ? WhereNode.cloneWithOperation(
+            aggregateFunctionNode.filter,
+            'And',
+            filter
+          )
         : WhereNode.create(filter),
     })
   },
@@ -62,7 +66,11 @@ export const AggregateFunctionNode = freeze({
     return freeze({
       ...aggregateFunctionNode,
       filter: aggregateFunctionNode.filter
-        ? WhereNode.cloneWithFilter(aggregateFunctionNode.filter, 'Or', filter)
+        ? WhereNode.cloneWithOperation(
+            aggregateFunctionNode.filter,
+            'Or',
+            filter
+          )
         : WhereNode.create(filter),
     })
   },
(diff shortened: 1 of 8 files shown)


koskimas:
> since where is nested in filter, maybe its more intuitive and future-proof to make a FilterBuilder that implements WhereInterface, and make filter accept a FilterBuilderCallback instead of acting as the where itself?

🤔 Maybe. On the other hand, too many nested callbacks starts to look messy.

> since filter clause is a thing, maybe we should rename the existing FilterNode, parsers, etc and introduce a new FilterNode?

We could go towards more generic `AST`ish nodes and rename it to `BinaryExpressionNode` and also have `UnaryExpressionNode` for `exists` etc. We'll need these anyway if we implement the full expression builder at some point. What do you think?

I originally called the node tree `AST` but renamed it to `operation nodes` since they are not actually SQL AST, but kinda like "kysely query builder AST" or something 😄

igalklebanov:
> We could go towards more generic ASTish nodes and rename it to BinaryOperationNode and also have UnaryOperationNode for exists etc. unary operations. We'll need these anyway if we implement the full expression builder at some point. What do you think?
>
> We already have these "semantic" parent nodes like WhereNode and OnNode so we wouldn't lose any "transformability".

Sounds good!

> I originally called the node tree AST but renamed it to "operation nodes" since they are not actually SQL AST, but kinda like "kysely query builder AST" or something 😄

😆 KyselyAST ™️ has a nice sound to it!

koskimas (approved):

PR #218, opened 2022-11-04
feat: add `ReadonlyKysely<DB>` helper.

igalklebanov:
Hey 👋 

This PR introduces `ReadonlyKysely<DB>`. A type you can use to expose a type-level read-only Kysely instance that denies anything but reads at compile time. This covers usage of the main instance, transactions, CTEs, etc.

```ts
import { Kysely } from 'kysely'
import type { ReadonlyKysely } from 'kysely/readonly'

export const db = new Kysely<Database>({...}) as never as ReadonlyKysely<Database>

db.selectFrom('person').selectAll() // no problem.
db.selectNoFrom(sql`now()`.as('now')) // no problem.

db.deleteFrom('person') // compilation error + deprecation!
db.insertInto('person').values({...}) // compilation error + deprecation!
db.mergeInto('person')...  // compilation error + deprecation!
db.updateTable('person').set('first_name', 'Timmy') // compilation error + deprecation!
sql`...`.execute(db) // compilation error!
```

---

original post:

as requested by @ethanresnick on discord.

Real-world use case: knowing you're exposing a Kysely instance that's connecting to a read replica, or uses a user that only has read permissions, you want to catch data mutating queries during compilation.

This obviously doesn't provide safety for raw queries. We could omit `getExecutor` so it can't be used as executor provider for `sql` template tag.

Downside is we'll have to maintain it, keep it aligned with changes to `Kysely`, `QueryCreator`, `Transaction`. `Pick` ensures it keeps its promise, even if it falls behind.

---

- [x] implement
- [x] typings tests


ethanresnick:
Wow, thanks @igalklebanov! I'm impressed you got this up so fast, and really appreciate it!

Now that you've written the implementation, I'll be curious to hear what you and @koskimas think about the value vs. maintenance effort tradeoff.

koskimas:
I don't know about this 🤔 This will easily get out of sync with the RW versions and there's no way to create tests that make sure it doesn't. Each time we a new method to the RW versions we need to remember to add a read-only version (if applicable). Expecially those `ReadonlyCommonTableExpression` types seem like a maintenance nightmare.

tonivj5:
Adding some context, we did something similar a week ago in our project to protect an instance of kysely over a slave connection database

```ts
import { Kysely } from 'kysely';

import { Database } from './database';

export class ReadonlyDatabaseConnection extends Kysely<Database> {
  /**
   * Don't use me, please. I'm a readonly database connection
   */
  declare replaceInto: never;

  /**
   * Don't use me, please. I'm a readonly database connection
   */
  declare insertInto: never;

  /**
   * Don't use me, please. I'm a readonly database connection
   */
  declare deleteFrom: never;

  /**
   * Don't use me, please. I'm a readonly database connection
   */
  declare updateTable: never;
}
```

Don't cover every situation (using `with*` & other methods) but it's mostly perfect to avoid certain actions. So it was a surprise to see this PR!

> This will easily get out of sync with the RW version and there's no way to create tests that make sure it doesn't. 

What about make it a class/subclass? `ReadWriteKysely extends ReadOnlyKysely` I don't know if it's technically possible or it's simply a bad idea... 😆 


Good work @koskimas & @igalklebanov, I really enjoy using Kysely 👍🏻

igalklebanov:
@koskimas 

> I don't know about this

Same.. the overhead for future contributions/reviews makes this really questionable. Thought the implementation would be not so hard (right now) and wanted to make this idea more presentable for this discussion.

Maybe this should be a separate package, with tight peer dependency? would be a hell to maintain, but won't affect Kysely.

koskimas:
> Same.. the overhead for future contributions/reviews makes this really questionable. Thought the implementation would be not so hard (right now) and wanted to make this idea more presentable for this discussion.

This could still be worth it if enough people find this useful. I simply don't know. I've never found myself wanting something like this.

igalklebanov:
> This could still be worth it if enough people find this useful. I simply don't know. I've never found myself wanting something like this.

If things like [this](https://planetscale.com/blog/introducing-planetscale-portals-read-only-regions) gain more traction, we might see more interest from the edge function crowd.

igalklebanov:
Reopened this and merged it after almost 4 years. We didn't drift that much, which means, the maintenance burden described in the comments is exaggerated. Most of it still worked.

I chose a different strategy for writes - include them as errorneous/deprecated overrides. The explicit error is a lot better for humans and agents. It also enables typings tests to verify `keyof` parity between source and readonly variant. This'll help us correct the readonly variants over time.

We'll ship it in version 0.29 under `kysely/readonly`. I'm looking forward to getting feedback on this one. It's been one of those things that kept popping up in issues. I had a production use case for this 2 years ago - so I'm personally a believer in its usefulness.

ethanresnick:
@igalklebanov Thank you! I'm gonna start using this ASAP.

PR #226, opened 2022-11-11
Add additional column in PostgreSQL migration example to showcase defaultTo

kharann:
I thought it would be nice to showcase `defaultTo` in the PostgreSQL migration example. Personally, I had to look into the discord to figure it out, so I think that an additional line won't hurt.


koskimas:
Thanks I'll merge this 👍 

I'm curious, where did you search for the method if you had trouble finding it? Just trying to type in `default` or anything similar would have autocompleted the method for you and the example is right there in the autocompletion dialog:

Screenshot 2022-11-12 at 16 19 47

PR #236, opened 2022-11-24
fix: trim sql row to remove leading whitespace or newline

### LINKED ISSUE #235: Bug: sqlite introspection does not detect autoincrement column when sql contains newlines
```typescript
const autoIncrementCol = createSql[0]?.sql
      ?.split(/[\(\),]/)
      ?.find((it) => it.toLowerCase().includes('autoincrement'))
      ?.split(/\s+/)?.[0]
      ?.replace(/["`]/g, '')
```
https://github.com/koskimas/kysely/blob/2c49bae555505d347e73ded4cafc0f5dc7279426/src/dialect/sqlite/sqlite-introspector.ts#L68:L72

given the following sql:
```sql
CREATE TABLE "users"
(
    "id"         INTEGER  NOT NULL PRIMARY KEY AUTOINCREMENT,
    "created_at" DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
    "updated_at" DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
    "name"       TEXT     NOT NULL
)
```

the result of find the autoincrement row is:   `\n    "id"         INTEGER  NOT NULL PRIMARY KEY AUTOINCREMENT` 
after split on `whitespace` the first item inside the array is `""` instead of `""id""`

tobiasfoerg:
Fixes #235

### DIFF
--- a/src/dialect/sqlite/sqlite-introspector.ts
+++ b/src/dialect/sqlite/sqlite-introspector.ts
@@ -68,7 +68,7 @@ export class SqliteIntrospector implements DatabaseIntrospector {
     const autoIncrementCol = createSql[0]?.sql
       ?.split(/[\(\),]/)
       ?.find((it) => it.toLowerCase().includes('autoincrement'))
-        ?.trimStart()
+      ?.trimStart()
       ?.split(/\s+/)?.[0]
       ?.replace(/["`]/g, '')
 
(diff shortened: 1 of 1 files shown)


src/dialect/sqlite/sqlite-introspector.ts
```diff
@@ -68,6 +68,7 @@ export class SqliteIntrospector implements DatabaseIntrospector {
     const autoIncrementCol = createSql[0]?.sql
       ?.split(/[\(\),]/)
       ?.find((it) => it.toLowerCase().includes('autoincrement'))
+        ?.trimStart()
```
koskimas, line 71: Wrong indentation level

koskimas:
Thanks! Fix the indentation and this is good to go

src/dialect/sqlite/sqlite-introspector.ts
tobiasfoerg: done

PR #240, opened 2022-11-26
add `executeTakeFirstRow(...)` & `executeTakeFirstRowOrThrow(...)` @ `RawBuilder`.

### LINKED ISSUE #212: add `executeTakeFirst(...)` & `executeTakeFirstOrThrow(...)` to `sql` template tag.
raised by @kristiannotari on discord.

Currently only `.execute(...)` is supported when using `sql` template tag to execute raw queries.

Adding `.executeTakeFirst(...)` & `.executeTakeFirstOrThrow(...)` would be a nice QoL enhancement.

igalklebanov:
closes #212.

- [x] implement.
- [x] unit tests.

small QoL improvement for anyone executing result set returning queries using `sql` template tag or `RawBuilder`.

I'm not 100% sure about the naming..

On one hand, `executeTakeFirst(...)` & `executeTakeFirstOrThrow(...)` do align with our query builders, so it'll feel like home for our consumers.

On the other, the lack of context in `RawBuilder.execute(...)` does not allow us to return `InsertResult`/`UpdateResult`/`DeleteResult` (unless we go for sql string compilation - which is probably something we don't want for kysely), so adding `Row` to these methods removes any false expectations a consumer might have.

---

This PR introduces a breaking change for anyone who's using `errorConstructor` arguments in `executeTakeFirstOrThrow(...)` - it's underlying node's type is now `QueryNode | RawNode`.

---

Maybe we should also have something like `executeTakeRows(...)` (returning `O[]`)?

### DIFF
--- a/src/query-builder/no-result-error.ts
+++ b/src/query-builder/no-result-error.ts
@@ -1,14 +1,15 @@
 import { QueryNode } from '../operation-node/query-node.js'
+import { RawNode } from '../operation-node/raw-node.js'
 
-export type NoResultErrorConstructor = new (node: QueryNode) => Error
+export type NoResultErrorConstructor = new (node: QueryNode | RawNode) => Error
 
 export class NoResultError extends Error {
   /**
    * The operation node tree of the query that was executed.
    */
-  readonly node: QueryNode
+  readonly node: QueryNode | RawNode
 
-  constructor(node: QueryNode) {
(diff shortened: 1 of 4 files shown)


koskimas:
Sorry, I think this is a bad idea and a weird API. Let's not add these methods to the raw builder.

PR #241, opened 2022-11-26
add `using` clause support @ `DeleteQueryBuilder`.

### LINKED ISSUE #177: Postgres deleteFrom(table) does not support `using` expression
Postgres allows specifying other columns when deleting rows from a table using columns from another. Unfortunately there is no corresponding method on [DeleteQueryBuilder](https://koskimas.github.io/kysely/classes/DeleteQueryBuilder.html).

e.g. 

```sql
DELETE FROM table 
USING other_table ot, another_table at 
WHERE table.id = ot.tableId OR table.id = at.tableId
```

Work around: 
```typescript
sql`write your full delete statement here`.execute(db)
```

Version: 0.21.6

igalklebanov:
Adding `using` to `DeleteQueryBuilder`:

1. allows `mysql` consumers to use additional tables to filter what gets deleted from TDF* (it also allows using joins now, as they were not usable previously). Caveat in spec is you have to repeat yourself and add TDF* again in `using` clause in order to have additional tables (not in TDF*) as part of the query (via `using` or in joins) - `delete from t1 using t1, t2` or `delete from t1 using t1 inner join t2`.

2. allows `postgres` consumers to use additional tables to filter what gets deleted from TDF*.

\* TDF = Table/s deleted from.

closes #177.

- [x] implement.
- [x] unit tests.
- [x] typings tests.
- [x] ts docs.

---

Takeaways:

1. A follow-up PR needs to address `DeleteQueryBuilder.returningAll(...)`. It should now allow passing a table argument (just like `SelectQueryBuilder.selectAll(...)`) so we could have `t2.*`.

2. A follow-up PR should add `QueryCreator.deleteFrom([t0, t1])` support. `mysql` supports deleting from multiple tables, and current PR is a real tease to our consumers since it allows to filter by additional tables but still delete only from one.


src/operation-node/delete-query-node.ts
```diff
@@ -69,4 +71,17 @@ export const DeleteQueryNode = freeze({
       explain,
     })
   },
+
+  cloneWithUsing(
+    deleteNode: DeleteQueryNode,
+    from: OperationNode[]
```
naorpeled, line 77:
```suggestion
    table: OperationNode[]
```

naorpeled (approved):

src/query-builder/delete-query-builder.ts
```diff
@@ -172,6 +176,98 @@ export class DeleteQueryBuilder<DB, TB extends keyof DB, O>
...
+   * ```sql
+   * delete from "toy"
+   * using "pet", "person"
+   * where "toy"."pet_id" = "pet"."id"
+   *   and "pet"."owner_id" = "person"."id"
+   *   and "person"."first_name" = $1
+   * returning "pet"."name"
+   * ```
+   */
+  using<TE extends [TableExpression<DB, TB>, ...TableExpression<DB, TB>[]]>(
```
koskimas, line 253: Isn't this exactly like `selectFrom` type-wise? I think you could just copy-paste `selectFrom` types from `query-creator` instead of creating a new set of types. You don't need to rename the types either IMO.

igalklebanov: I remember it wasn't for some reason, will try again.

igalklebanov:
Reverted all the renames and added type bloat in `DeleteQueryBuilder`. Thanks for feedback!

koskimas:
Awesome! 💯

MoSheikh:
Hi, thanks a lot for this functionality! 

I was just using it and did notice however that table names provided in the "using" clause are not picked up by the CamelCasePlugin, leading to an error. Using snake case in the using clause does work, but ends up causing type errors since the table name is not recognized.

MoSheikh:
I [created a PR](https://github.com/koskimas/kysely/pull/315) to address the issue: feedback is appreciated!

PR #242, opened 2022-11-29
Add `call` to create/alter table statement builder

jacobpgn:
This is a copy of the approach used on the query builders (e.g. [`SelectQueryBuilder`](https://github.com/koskimas/kysely/blob/9c9acc6a4dbaf90b178397d3ceafdc22d3cb2b46/src/query-builder/select-query-builder.ts#L1471-L1495)), applied to the builders used for creating and altering tables. This addition should make it easier to write reusable functions that operate on those builders!

### DIFF
--- a/src/schema/alter-table-builder.ts
+++ b/src/schema/alter-table-builder.ts
@@ -208,6 +208,15 @@ export class AlterTableBuilder {
       }),
     })
   }
+
+  /**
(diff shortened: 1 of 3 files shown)


koskimas (approved):

koskimas:
Perfect! Thank you!

Issue #146, opened 2022-08-22
introspection.getTables sorts column array alphabetically instead of by ordinal position


jaym910:
Is this intentional? It is a bit counter intuitive to me and makes the introspection feature slightly less useful

would be happy to do a PR if you think this change would make sense.

koskimas:
Why is it counter intuitive? And what is the ordinal position anyway in case of columns?

wirekang:
This issue should be reopen. Columns have a clear order regardless of dialect. 

#### Mysql
`information_schema.columns.ordinal_position`

#### Postgres
`inforamtion_schema.columns.ordinal_position` or `pg_catalog.pg_attribute.attnum`

#### Sqlite
`pragma_table_info(table).cid`

## Suggestion

Reverting part of f6588f8f could solve this problem. However, since there is no criterion for ordering the tables, sorting the tables by name seems necessary.

This line
https://github.com/koskimas/kysely/blob/71507c4102ca1ae86fa18940c53310aa6e15ac1b/src/dialect/mysql/mysql-introspector.ts#L49

should be `orderBy('ordinal_position')`

This line
https://github.com/koskimas/kysely/blob/71507c4102ca1ae86fa18940c53310aa6e15ac1b/src/dialect/postgres/postgres-introspector.ts#L80

should be `orderBy('a.attnum')`

This line
https://github.com/koskimas/kysely/blob/71507c4102ca1ae86fa18940c53310aa6e15ac1b/src/dialect/sqlite/sqlite-introspector.ts#L85

should be `orderBy('cid')`

PR #262, opened 2022-12-29
Keep column order same as ordinal_position.

wirekang:
related #146


wirekang:
@koskimas
I don't know what you would think about this change. If you think this is not worth it, I can bring you more reasons.

koskimas:
I think this is fine. The user can always sort the list by name.

PR #267, opened 2022-12-30
add `FunctionModule.countAll(...)`.

### LINKED ISSUE #250: add `count(*)` & `count(t1.*)` support.
Currently the following syntax is not supported:

```sql
select count(*) as count from t1;
select count(t1.*) as count from t1 inner join t2 on t1.id = t2.t1_id;
```

This should probably be added as `FunctionModule.countAll(table?)`.

igalklebanov:
Closes #250.

- [x] implement.
- [x] unit test.
- [x] typings test.
- [x] document.

---

Also went over `FunctionModule`'s docs.

### DIFF
--- a/src/query-builder/function-module.ts
+++ b/src/query-builder/function-module.ts
@@ -84,6 +84,14 @@ export class FunctionModule<DB, TB extends keyof DB> {
    * select avg("price") as "avg_price" from "toy"
    * ```
    *
+   * You can limit column range to only columns participating in current query:
+   *
+   * ```ts
+   * db.selectFrom('toy')
+   *   .select(qb => qb.fn.avg('price').as('avg_price'))
+   *   .execute()
+   * ```
+   *
    * If this function is used in a `select` statement, the type of the selected
    * expression will be `number | string` by default. This is because Kysely can't know the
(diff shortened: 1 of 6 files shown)


koskimas:
Awesomness!

src/query-builder/function-module.ts
```diff
@@ -198,11 +285,112 @@ export class FunctionModule<DB, TB extends keyof DB> {
...
+   * see {@link count}.
+   *
+   * For additional functionality such as filtering and window functions, refer
+   * to {@link AggregateFunctionBuilder}. An instance of this builder is returned
+   * when calling this function.
+   *
+   * ### Examples
+   *
+   * ```ts
+   * const { count } = db.fn
```
zmilonas, line 304: I think you meant `countAll` here otherwise it might not work 😅

igalklebanov: @zmilonas Nice catch! Wanna submit a PR fixing this?

zmilonas: Sure thing! Here you go: https://github.com/koskimas/kysely/pull/303

PR #275, opened 2023-01-06
Implement CockroachDB support

lbguilherme:
[CockroachDB](https://www.cockroachlabs.com/) is a scalable database built for the cloud and compatible with PostgreSQL.

This PR adds a new dialect for it. Mostly, everything is the same as Postgres, except that `acquireMigrationLock` had to be modified because it doesn't support advisory locks.

This is my first contribution here, I hope I did everything right. This is a feature I need for some projects where we plan to migrate from Knex to Kysely.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,3 +18,11 @@ services:
       POSTGRES_HOST_AUTH_METHOD: trust
     ports:
       - '5434:5432'
+  cockroachdb:
+    image: cockroachdb/cockroach
+    command: start-single-node --insecure
+    environment:
+      COCKROACH_DATABASE: kysely_test
+      COCKROACH_USER: kysely
+    ports:
+      - '26257:26257'
--- a/src/dialect/cockroach/cockroach-adapter.ts
+++ b/src/dialect/cockroach/cockroach-adapter.ts
@@ -1,0 +1,17 @@
+import { Kysely } from '../../kysely.js'
+import { KyselyPlugin } from '../../plugin/kysely-plugin.js'
+import { PostgresAdapter } from '../postgres/postgres-adapter.js'
+
+export class CockroachAdapter extends PostgresAdapter {
+  async acquireMigrationLock(db: Kysely<any>, schemaPlugin: KyselyPlugin, migrationLockTable: string): Promise<void> {
+    // We don't need to actually write to a row on this table, only selecting it "for update"
+    // will ensure no other transaction can get past this point. They will wait in a queue.
+    await db.withPlugin(schemaPlugin).selectFrom(migrationLockTable).selectAll().forUpdate().execute();
+  }
+
+  async releaseMigrationLock(): Promise<void> {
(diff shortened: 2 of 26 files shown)


igalklebanov:
Thank you! 💪 

> This is my first contribution here, I hope I did everything right.

You did everything right. A very good and thorough PR!

> This is a feature I need for some projects where we plan to migrate from Knex to Kysely.

This is exciting! Thanks for believing in the project. 🍾 

> This PR adds a new dialect for it. Mostly, everything is the same as Postgres, except that acquireMigrationLock had to be modified because it doesn't support advisory locks.

We aim to support the main staples, and allow for 3rd party libraries to provide additional dialects.
See [3rd party dialects](https://github.com/koskimas/kysely#3rd-party-dialects) section.

I'm not sure we want to add CockroachDB to the core.
Its an exciting offering but not quite a household name. We'll have to maintain it, and take it into account with every new feature or change. ...and speaking for myself, I'm not well versed in this offering.

@koskimas WDYT?

lbguilherme:
The fact that it's very similar to Postgres makes it easier to maintain, but it isn't identical and I understand it can be a burden in the future. I'm willing to help. If it's not accepted in the Core, I can maintain as a 3rd party dialect in a npm module, no problem.

Regardless, the change to pass the schema and table name to the `acquireMigrationLock` function is quite important for this to work. Otherwise it is impossible to accept a custom lock table name or schema. (https://github.com/koskimas/kysely/pull/275/files#diff-f9fbf3cc67ca7a23e40c5a27ef6e462511a37d74d5c035894545e3008b98fed4L19-R21)

koskimas:
Thank you for the contribution @lbguilherme! But I have to agree with @igalklebanov. This shouldn't be added to the core, even though it's currently very similar to Postgres. If we have CockroachDB in the core, we'd have to consider adding all kinds of Cockroach specific features to the API, just as we have some Postgres and MySQL specific special methods. Over time the maintenance burden will grow.

> Regardless, the change to pass the schema and table name to the acquireMigrationLock function is quite important for this to work. Otherwise it is impossible to accept a custom lock table name or schema.

That's a good point. I'll figure something out for that. I'll keep this PR open and post here once I have released a solution.

koskimas:
The lock table, schema and row id are now passed to the lock methods

https://github.com/koskimas/kysely/blob/fbcf3e3853e6e466fc7e93ce0294d6da32554a14/src/dialect/dialect-adapter.ts#L25

I'll release a new version today.

koskimas:
0.23.4 is now released.

lbguilherme:
Thanks @koskimas! Totally understand the decision and we can move forward now with 0.23.4.

zhamid-src:
If it helps, I can't do migrations with kysley without this being in the core. I know @lbguilherme created a plugin but it doesn't work. It's also not necessarily as maintained to be used in production. This will sadly become a dealbreaker for me.

PR #276, opened 2023-01-07
Fix FileMigrationProvider example

gimenete:
It seems like the `FileMigrationProvider` constructor changed at some point and the jsdoc is not in sync.


koskimas:
Thank you!

PR #284, opened 2023-01-18
CamelCasePlugin option to maintainJson data format

### LINKED ISSUE #281: [CamelCasePlugin] Add option to disable transforming of JSON content
When using the CamelCasePlugin, it also transforms the content of JSON fields.
It would be nice to have an option that disables that, so the contents remain in snake_case.

Discord thread for reference:
https://discord.com/channels/890118421587578920/890118421587578925/1063741659545882625

otaviosoares:
closes #281

### DIFF
--- a/src/plugin/camel-case/camel-case-plugin.ts
+++ b/src/plugin/camel-case/camel-case-plugin.ts
@@ -45,6 +45,13 @@ export interface CamelCasePluginOptions {
    * Defaults to false.
    */
   underscoreBetweenUppercaseLetters?: boolean
+
+  /**
+   * If true, nested object's keys will not be converted to camel case.
+   *
+   * Defaults to false.
+   */
+  maintainNestedObjectKeys?: boolean
 }
 
 /**
@@ -116,7 +123,7 @@ export class CamelCasePlugin implements KyselyPlugin {
   readonly #snakeCase: StringMapper
   readonly #snakeCaseTransformer: SnakeCaseTransformer
 
-  constructor(opt: CamelCasePluginOptions = {}) {
+  constructor(readonly opt: CamelCasePluginOptions = {}) {
     this.#camelCase = createCamelCaseMapper(opt)
     this.#snakeCase = createSnakeCaseMapper(opt)
 
@@ -147,8 +154,8 @@ export class CamelCasePlugin implements KyselyPlugin {
       let value = row[key]
 
       if (Array.isArray(value)) {
-        value = value.map((it) => (canMap(it) ? this.mapRow(it) : it))
-      } else if (canMap(value)) {
+        value = value.map((it) => (canMap(it, this.opt) ? this.mapRow(it) : it))
+      } else if (canMap(value, this.opt)) {
         value = this.mapRow(value)
       }
 
@@ -166,8 +173,15 @@ export class CamelCasePlugin implements KyselyPlugin {
   }
 }
 
-function canMap(obj: unknown): obj is Record<string, unknown> {
+function canMap(
+  obj: unknown,
+  opt: CamelCasePluginOptions
+): obj is Record<string, unknown> {
   return (
-    isObject(obj) && !isDate(obj) && !isBuffer(obj) && !isArrayBufferOrView(obj)
+    isObject(obj) &&
+    !isDate(obj) &&
+    !isBuffer(obj) &&
+    !isArrayBufferOrView(obj) &&
+    !opt?.maintainNestedObjectKeys
   )
 }
--- a/test/node/src/camel-case.test.ts
+++ b/test/node/src/camel-case.test.ts
@@ -1,4 +1,4 @@
-import { CamelCasePlugin, Generated, Kysely, sql } from '../../../'
+import { CamelCasePlugin, Generated, Kysely, RawBuilder, sql } from '../../../'
 
 import {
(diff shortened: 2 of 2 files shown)


test/node/src/camel-case.test.ts
```diff
@@ -203,5 +209,33 @@ for (const dialect of BUILT_IN_DIALECTS) {
...
+    if (dialect !== 'sqlite') {
+      it('should respect maintainJsonFormat', async () => {
+        const dbWithOption = new Kysely<CamelDatabase>({
+          ...ctx.config,
+          plugins: [
+            new CamelCasePlugin({
+              maintainJsonFormat: true,
+            }),
+          ],
+        })
```
igalklebanov, line 223: let's reuse `camelDb`..

```ts
camelDb.withoutPlugins().withPlugin(new CamelCasePlugin({ maintainJsonFormat: true }))
```

test/node/src/camel-case.test.ts
```diff
@@ -203,5 +209,33 @@ for (const dialect of BUILT_IN_DIALECTS) {
         },
       })
     })
+
+    // Can't run this test on SQLite because json field returns string
+    if (dialect !== 'sqlite') {
```
igalklebanov, line 214: could we cover sqlite as well? it's a `JSON.parse` or a an assertion against a string away..

test/node/src/camel-case.test.ts
```diff
@@ -203,5 +209,33 @@ for (const dialect of BUILT_IN_DIALECTS) {
...
+              maintainJsonFormat: true,
+            }),
+          ],
+        })
+
+        const data = await dbWithOption
+          .selectFrom('camelPerson')
+          .selectAll()
+          .executeTakeFirst()
+        expect(data?.preferences).to.eql({
```
igalklebanov, line 229:
```suggestion
          .executeTakeFirstOrThrow()
          
        expect(data.preferences).to.eql({
```

test/node/src/camel-case.test.ts
```diff
@@ -203,5 +209,33 @@ for (const dialect of BUILT_IN_DIALECTS) {
...
+          disable_emails: true,
+        })
+
+        await dbWithOption.destroy()
```
igalklebanov, line 233: no need if reusing `camelDb`.

src/plugin/camel-case/camel-case-plugin.ts
```diff
@@ -166,8 +173,15 @@ export class CamelCasePlugin implements KyselyPlugin {
   }
 }
 
-function canMap(obj: unknown): obj is Record<string, unknown> {
+function canMap(
+  obj: unknown,
+  opt?: CamelCasePluginOptions
```
igalklebanov, line 178:
```suggestion
  opt: CamelCasePluginOptions
```

src/plugin/camel-case/camel-case-plugin.ts
```diff
@@ -166,8 +173,15 @@ export class CamelCasePlugin implements KyselyPlugin {
...
+): obj is Record<string, unknown> {
   return (
-    isObject(obj) && !isDate(obj) && !isBuffer(obj) && !isArrayBufferOrView(obj)
+    isObject(obj) &&
+    !isDate(obj) &&
+    !isBuffer(obj) &&
+    !isArrayBufferOrView(obj) &&
+    !opt?.maintainJsonFormat
```
igalklebanov, line 185:
```suggestion
    !opt.maintainJsonFormat
```

igalklebanov (changes requested):
Hey 👋 

Thank you! 💪

igalklebanov:
@koskimas wondering if transformer should apply naming strategy recursively to json parameters, seems not aligned with output transformation logic.. WDYT?

test/node/src/camel-case.test.ts
otaviosoares: Thank, didn't know about that one

igalklebanov (approved):
LGTM! 🚀

koskimas:
`maintainJsonFormat` is not the best name for the option IMO. This doesn't just apply to JSON but any nested objects. A better name would be `maintainNestedObjectKeys`

koskimas:
> @koskimas wondering if transformer should apply naming strategy recursively to json parameters, seems not aligned with output transformation logic.. WDYT?

You mean essentially making this option `true` by default? The use case I was thinking when I implemented this wasn't only json object columns but `json_agg` etc. nested objects that would by default have snake_cased properties you'd also want to map.

I'm fine by changing the default, but It could be a nasty breaking change.

koskimas:
LGTM! Thank you @otaviosoares ❤️

igalklebanov:
> > @koskimas wondering if transformer should apply naming strategy recursively to json parameters, seems not aligned with output transformation logic.. WDYT?
> 
> You mean essentially making this option `true` by default? The use case I was thinking when I implemented this wasn't only json object columns but `json_agg` etc. nested subqueries that would by default have snake_cased properties you'd also want to map.
> 
> I'm fine with changing the default, but It could be a nasty breaking change.

Sorry, I meant that we seem to only alter identifiers on input, doing nothing to json parameters in value nodes. On output, we touch everything.

Makes you wonder, if for predictability, consumers would want the option that nested json keys have same naming strategy as columns.

Issue #224, opened 2022-11-11
Allow `DEFERRABLE` or `modifyEnd/modifyStart` to constraints


rubenferreira97:
Currently we can't set a constraint (Unique, FK, etc...) as `[DEFERRABLE | NOT DEFERRABLE] [INITIALLY DEFERRED | INITIALLY IMMEDIATE]` on `ALTER TABLE` nor `CREATE TABLE`.

Example:
```sql
ALTER TABLE tablename
   DROP CONSTRAINT constraintname,
   ADD  CONSTRAINT constraintname UNIQUE(columnname) DEFERRABLE;
```
```ts
db.schema.alterTable('tablename').dropConstraint('constraintname').execute();
db.schema.alterTable('tablename').addUniqueConstraint('constraintname', ['columnname']).execute(); // No DEFERRABLE nor .modifyEnd 
```

koskimas:
We should add `deferrable()` and `initiallyDeferred()` methods. This applies to all constraints, including the primary key.

rubenferreira97:
@koskimas I think `initially immediate` makes sense if a constraint can be changed (not all dbs allow this). An example would be a constraint created with `initially deferred` and now we want to change it to `initially immediate` with `ALTER CONSTRAINT`. As far as I know `initially` here means the default transaction "deferred mode". If we don't change it (in Postgres via `SET CONSTRAINTS { ALL | name [, ...] } { DEFERRED | IMMEDIATE }`) it will mantain the value declared after the `initially` keyword.

For the same reason I think if `initially immediate` is added to Kysely `NOT DEFERRABLE`/`notDeferrable()` should also be taken to consideration. We could change a constraint that was `DEFERRABLE` to `NOT DEFERRABLE`.

naorpeled:
@igalklebanov @koskimas 
can you assign me to this issue?

Thanks 😎 🙏

naorpeled:
Sorry guys,
been super swamped with stuff to do,
it's better if someone else will take this 🙏

anirudh1713:
Hi @igalklebanov @koskimas I'd love to take this on if its okay. Let me know.

Thanks.

igalklebanov:
It's yours. Have fun! 🙂

anirudh1713:
Hey all! I apologize  for the delay here 🙏, I'm unassigning this for now as I've been busy and will be busy till mid August due to some unexpected things (In good way).

viraxslot:
[...]

PR #291, opened 2023-01-20
wip(constraints): add deferrable support

naorpeled:


drewbitt:
This would be great; as a workaround until this is merged I am doing the following as SQLite doesn't support the ADD CONSTRAINT variant of the ALTER TABLE, so columns + references have to be added at the same time
```
await sql`ALTER TABLE mytable ADD COLUMN goalId INTEGER REFERENCES goals(id) DEFERRABLE INITIALLY DEFERRED;`.execute(db);
```

JDvorak:
Why was this closed?

drewbitt:
> Why was this closed?

Because https://github.com/kysely-org/kysely/issues/224#issuecomment-1485937688

PR #293, opened 2023-01-21
Allow to have additional properties for KyselyPlugin.

KostyaTretyak:
This change allow create plugin like this:

```ts
const plugin: KyselyPlugin = {
  data: new WeakMap<QueryId, SomeData>(),

  transformQuery(args: PluginTransformQueryArgs) {
    this.data.set(args.queryId, something);
    return args.node;
  },

  async transformResult(args: PluginTransformResultArgs) {
    const data = this.data.get(args.queryId);
    return args.result;
  },
};
```

For now, `data` property not allowed in `KyselyPlugin`.

### DIFF
--- a/src/plugin/kysely-plugin.ts
+++ b/src/plugin/kysely-plugin.ts
@@ -43,6 +43,8 @@ export interface KyselyPlugin {
   transformResult(
     args: PluginTransformResultArgs
   ): Promise<QueryResult<UnknownRow>>
+
+  [property: string | number | symbol]: any
 }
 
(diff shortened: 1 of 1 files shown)


KostyaTretyak:
By the way, should `KyselyPlugin` properties really be required? It seems to me that most often the user passes either `transformQuery` or `transformResult`, but not both.

koskimas:
KyselyPlugin is just an interface. Simply leave out the explicit type in your example and the object will implement KyselyPlugin AND have extra properties.

Issue #286, opened 2023-01-19
docs(website): some links don't parses and other issues with docs


KostyaTretyak:
I found three cases where the `{@link tagName}` entry does not parse:

- `{@link selectAll}`
- `{@link sql}`
- `{@link $if}`

https://koskimas.github.io/kysely/classes/SelectQueryBuilder.html#select

KostyaTretyak:
Also broken this example: https://koskimas.github.io/kysely/classes/SelectQueryBuilder.html#innerJoinLateral

~~~
```ts db.selectFrom('person') .innerJoinLateral( (eb) => eb.selectFrom('pet') .select('name') .whereRef('pet.owner_id', '=', 'person.id') .as('p'), (join) => join.onTrue() ) .select(['first_name', 'p.name']) .orderBy('first_name')
~~~

**Comment by igalklebanov** [comment]
Hey 👋 

Thank you! 💪 

Wanna submit a PR for these?

**Comment by KostyaTretyak** [comment]
[My project](https://github.com/ditsmod/realworld) is currently migrating to kysely. Maybe I'll find more bugs. If no one fixes the bugs described in this issue by the end of this migration, I will try to fix it.

**Comment by KostyaTretyak** [comment]
Incorrect example with plugins: https://koskimas.github.io/kysely/interfaces/KyselyPlugin.html#transformQuery

Need fix, like this:

```ts
const plugin: KyselyPlugin & { data: WeakMap<any, any> } = {
  data: new WeakMap<QueryId, SomeData>(),

  transformQuery(args: PluginTransformQueryArgs) {
    this.data.set(args.queryId, something);
    return args.node;
  },

  async transformResult(args: PluginTransformResultArgs) {
    const data = this.data.get(args.queryId);
    return args.result;
  },
};
```

By the way, I think `KyselyPlugin` need extend the to allow for other properties besides `transformQuery` and `transformResult`.

PR #290, opened 2023-01-20
Fix missing quotes in typedoc

yardenshoham:
Fix #286

PR #296, opened 2023-01-22
Fix docs for #286

KostyaTretyak:


igalklebanov:
Hey 👋 

Thank you!

Does #290 solve #286 ?

KostyaTretyak:
@igalklebanov, #290 solves only one of the issues in #286.

src/plugin/kysely-plugin.ts
```diff
@@ -13,21 +13,25 @@ export interface KyselyPlugin {
...
+  }
+
+  async transformResult(args: PluginTransformResultArgs) {
+    const data = this.data.get(args.queryId);
+    return args.result;
+  }
+}
+
+const myPlugin = new MyPlugin();
+```
```
igalklebanov, line 34: Let's keep it focused. 

The intention is to show how to pass data from `transformQuery` to `transformResult`.
Things like imports or how to instantiate the class are noise here.

It is also missing some `*`s at the start of each line..

```suggestion
```ts
class MyPlugin implements KyselyPlugin {
  private data = new WeakMap<QueryId, SomeData>();

  transformQuery(args: PluginTransformQueryArgs) {
    this.data.set(args.queryId, something);
    return args.node;
  }

  async transformResult(args: PluginTransformResultArgs) {
    const data = this.data.get(args.queryId);
    return args.result;
  }
}
```

src/query-builder/select-query-builder.ts
```diff
@@ -319,9 +319,9 @@ export class SelectQueryBuilder<DB, TB extends keyof DB, O>
    * same as calling `select(['id', 'first_name']).
    *
    * To select all columns of the query or specific tables see the
-   * {@link selectAll} method.
+   * {@link SelectQueryBuilder.selectAll} method.
```
igalklebanov, line 322: Revert please.

```suggestion
   * {@link selectAll} method.
```

src/query-builder/select-query-builder.ts
```diff
@@ -319,9 +319,9 @@ export class SelectQueryBuilder<DB, TB extends keyof DB, O>
...
    *
-   * See the {@link $if} method if you are looking for a way to add selections
+   * See the {@link SelectQueryBuilder.$if} method if you are looking for a way to add selections
```
igalklebanov, line 324: Revert please.

```suggestion
   * See the {@link $if} method if you are looking for a way to add selections
```

src/query-builder/select-query-builder.ts
```diff
@@ -436,7 +436,7 @@ export class SelectQueryBuilder<DB, TB extends keyof DB, O>
    * from "person"
    * ```
    *
-   * In case you use the {@link sql} tag you need to specify the type of the expression
+   * In case you use the {@link Sql} tag you need to specify the type of the expression
```
igalklebanov, line 439: Revert please.

```suggestion
   * In case you use the {@link sql} tag you need to specify the type of the expression
```

src/query-builder/select-query-builder.ts
```diff
@@ -987,6 +987,7 @@ export class SelectQueryBuilder<DB, TB extends keyof DB, O>
    *   )
    *   .select(['first_name', 'p.name'])
    *   .orderBy('first_name')
+   * ```
```
igalklebanov, line 990: Nice catch! 💪

igalklebanov (changes requested):

igalklebanov:
@KostyaTretyak This is why having a separate issue for each problem works. 😄 

Let's remove the things #290 does, and focus on what it doesn't OK?

src/query-builder/select-query-builder.ts
KostyaTretyak: Why?

igalklebanov: It works without the change.

PR #304, opened 2023-01-28
use cache in test.yml

jongwooo:
Signed-off-by: jongwooo <jongwooo.han@gmail.com>

## Description

Updated `test.yml` to cache dependencies using [actions/setup-node](https://github.com/actions/setup-node#caching-global-packages-data). `setup-node@v3` or newer has caching **built-in**.

### About caching workflow dependencies
Jobs on GitHub-hosted runners start in a clean virtual environment and **must download dependencies each time**, causing increased network utilization, longer runtime, and increased cost. To help speed up the time it takes to recreate files like dependencies, GitHub can cache files that frequently use in workflows.

### Solutions
Node.js projects can run faster on GitHub Actions by enabling dependency caching on the [setup-node](https://github.com/actions/setup-node) action. 

### AS-IS

```yml
- name: Use Node.js ${{ matrix.node-version }}
  uses: actions/setup-node@v3
  with:
    node-version: ${{ matrix.node-version }}
```

### TO-BE

```yml
- name: Use Node.js ${{ matrix.node-version }}
  uses: actions/setup-node@v3
  with:
    node-version: ${{ matrix.node-version }}
    cache: npm
```

> It’s literally a one line change to pass the `cache: npm` input parameter.

## References

- [https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,9 +19,10 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node-version }}
+          cache: npm
 
       - name: Use Deno ${{ matrix.deno-version }}
         uses: denolib/setup-deno@v2


igalklebanov:
Hey 👋 

Thank you!

How significant is this for a small library project?

I re-ran node v18 workflow.. and the difference from v16 run (no cache) seems to be:

- use node step 1s -> 3s (+2s downloading 40mb cache contents)
- install dependencies step 18s -> 15s (-3s)

jongwooo:
> I re-ran node v18 workflow.. and the difference from v16 run (no cache) seems to be:
use node step 1s -> 3s (+2s downloading 40mb cache contents)
install dependencies step 17s -> 15s (-2s)

@igalklebanov I agree with your opinion. If this PR is not needed, please close it.

PR #308, opened 2023-01-30
code hints/docs: Remove duplicate .execute from insert onConflict with constraint example.

rvdende:
I noticed theres a duplicate .execute command, must be a typo?


koskimas:
Thanks!

PR #315, opened 2023-02-06
Fix using node transformation

MoSheikh:
Fixes bug in `transformUsing` method which resulting in table references in a `using` clause not being transformed by the CamelCasePlugin.


igalklebanov (approved):
LGTM! Nice catch. 💪

koskimas:
Awesome! Thanks ❤️

PR #324, opened 2023-02-10
Update `executeTakeFirstOrThrow` to allow throwing custom error objects

mikewuu:
Currently, `executeTakeFirstOrThrow` only accepts a custom error **class**. This would allow users to pass a fully formed error **object** to throw if they don't care about the `node`.

- [x] implement
- [x] document
- [x] test


igalklebanov:
Hey 👋 

Thanks! 💪 

I'm not sure about this one..

Currently you can provide a custom error class that extends `NoResultError`. 
You receive the query syntax tree at runtime, which enables more accurate error messages.
You can also define a hard-coded a hard coded error message.

I don't see any benefit to passing an instance of an error instead of the class itself. What am I missing?

mikewuu:
I agree. Functionally, this offers very little benefit, though if the user doesn't care about the underlying error, it would be nice to reduce the amount of lines of code required.

Here's a real example that I'm currently facing:

```ts
const building = await req.db
    .selectFrom("building")
    .where("building_id", "=", building_id)
    .executeTakeFirst()

  if (!building) {
    throw new NotFoundException({
      type: "not_found",
      message: "Building not found",
    })
  }
```

After this PR:

```ts
const building = await req.db
    .selectFrom("building")
    .where("building_id", "=", building_id)
    .executeTakeFirstOrThrow(
      new NotFoundException({
        type: "not_found",
        message: "Building not found",
      })
    )
```

The project above avoids introducing new types where possible to reduce complexity. Creating a new error class would violate this.

igalklebanov:
If a user doesn't care about the error, she can simply call `.executeTakeFirstOrThrow()` without any arguments. This will throw the basic `NoResultError`.

igalklebanov:
If a user doesn't care about the error, she can simply call `.executeTakeFirstOrThrow()` without any arguments. This will throw the basic `NoResultError`.

> The project above avoids introducing new types where possible to reduce complexity. Creating a new error class would violate this.

But it instantiates an error prematurely for every invocation.

mikewuu:
> But it instantiates an error prematurely for every invocation.

Fair. Will this be better? 

```ts
const building = await req.db
    .selectFrom("building")
    .where("building_id", "=", building_id)
    .executeTakeFirstOrThrow(
    () =>  new NotFoundException({
        type: "not_found",
        message: "Building not found",
      })
    )
```

igalklebanov:
Yeah, that makes sense.

bjornlll:
Thanks @mikewuu and @igalklebanov! 

Would it make sense for the error creation function to also receive `OperationNode` as an argument? This way we could have a custom error class hierarchy and still get the same context as `NoResultError` and its extensions.

```typescript
const building = await req.db
    .selectFrom("building")
    .where("building_id", "=", building_id)
    .executeTakeFirstOrThrow(
    // 👇 Added argument
    (node: OperationNode) =>  new NotFoundException({
        type: "not_found",
        message: "Building not found",
      })
    )
```

mikewuu:
I was thinking the same thing literally minutes ago. Will push an update later today, thanks!

mikewuu:
@igalklebanov , @bjornlll , just updated this.

koskimas (approved):
This is good to go once rebased! Sorry for the massive delay 😞

bjornlll:
No reason to be sorry @koskimas ! I'm just excited we're getting this in. It's one of those things that won't fundamentally change life but will make me feel a tad bit happier when looking at my code.

Thanks again @mikewuu for spearheading this!

PR #355, opened 2023-03-08
add compiled query and timing to error logs

### LINKED ISSUE #342: Log compiled query in error
Per this comment, I would like to suggest an improvement to the logging flow:

https://github.com/koskimas/kysely/issues/24#issuecomment-994893505

It would be useful to expose the compiled query on the error log too. This can help when debugging timeouts. I'm happy to open a pr implementing this if it seems reasonable. Thanks for the great library!

fwojciec:
This PR is a follow up to [this issue](https://github.com/koskimas/kysely/issues/342).

The change makes it possible to log the compiled query along with its timing in case of error logs. As the issue mentioned, this makes it easier to debug failing queries.

closes #342

### DIFF
--- a/src/driver/runtime-driver.ts
+++ b/src/driver/runtime-driver.ts
@@ -105,7 +105,7 @@ export class RuntimeDriver implements Driver {
       try {
         return await executeQuery.call(connection, compiledQuery)
       } catch (error) {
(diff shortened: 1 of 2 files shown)


koskimas:
Nice, thank you!

PR #364, opened 2023-03-18
add `Kysely.executeQuery`.

igalklebanov:
This PR introduces a shortcut for `Kysely.getExecutor().executeQuery(compiledQuery, queryId)`.

When splitting build, compile and execute code, consumers no longer have to:

- access Kysely's internal `.getExecutor()` method.
- provide a queryId - which is internal.

Additionally: 
- consumer can pass a query builder, it's get compiled under the hood.
- QueryResult gets the correct R type from compiled query when using QueryExecutor's execute query.


koskimas (approved):
LGTM

PR #365, opened 2023-03-18
add `RawBuilder.compile`.

igalklebanov:
This PR allows consumers to compile `RawBuilder`s and pass around their `CompiledQuery`s when splitting build, compile and execute code.


koskimas:
Nice!

PR #371, opened 2023-03-20
add `.where` & `.orWhere` @ `CreateIndexBuilder`.

### LINKED ISSUE #302: Creating a Partial Index
Sorry for creating that many issues/questions.

While doing #301, I tried to create a partial index. Namely, I tried to achieve this:
```sql
CREATE INDEX post_expiration
ON post (expiration)
WHERE expiration IS NOT NULL
```

[SQLite](https://sqlite.org/partialindex.html) and [Postgres](https://www.postgresql.org/docs/current/indexes-partial.html) support partial indices. I'm not quite sure about the situation with MySQL/MariaDB.

I could not find a builder method for this, but I managed to [abuse `expression` for this](https://wirekang.github.io/kysely-playground/?p=f&i=-NMaBRrRK8T7IOeb7ICF):

```ts
db.schema.createIndex("post_created_at")
    .on("post")
    .expression(sql`expiration DESC) WHERE (expiration IS NOT NULL`)
```

Which results in this SQL:
```sql
CREATE INDEX `post_created_at` ON `post` (expiration DESC)
WHERE
  (expiration IS NOT NULL)
```

Well, that's obviously not the way to use `createIndex`. Is the last resort falling back to raw SQL here? Would this be something worth considering for a solution on a query builder level?

igalklebanov:
closes #302.

Adds the ability to create partial indexes in postgres / sqlite.

- only columns known in compile-time (used in `.column`/`.columns`) provide autocompletion,
unknown columns require `sql` template tag to be referred to.
- parameter types are unknown.
- parameters are not allowed (engine throws), so they're turned to literals before parsing.

### DIFF
--- a/src/parser/binary-operation-parser.ts
+++ b/src/parser/binary-operation-parser.ts
@@ -32,7 +32,6 @@ import { HavingInterface } from '../query-builder/having-interface.js'
 import { createJoinBuilder, createSelectQueryBuilder } from './parse-utils.js'
 import { OperationNode } from '../operation-node/operation-node.js'
 import { Expression } from '../expression/expression.js'
-import { sql } from '../raw-builder/sql.js'
 
 export type OperandValueExpression<
   DB,
@@ -101,9 +100,9 @@ export function parseFilterExpression(
   throw createFilterExpressionError(type, args)
 }
 
-export function parseWhereWithParametersAsLiterals(args: any[]): OperationNode {
+export function parseWhereWithImmediateParameters(args: any[]): OperationNode {
   if (args.length === 3) {
-    args = [args[0], args[1], sql.literal(args[2])]
+    args = [args[0], args[1], ValueNode.createImmediate(args[2])]
   }
 
   return parseWhere(args)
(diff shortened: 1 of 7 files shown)


src/schema/create-index-builder.ts
```diff
@@ -126,6 +133,101 @@ export class CreateIndexBuilder implements OperationNodeSource, Compilable {
...
+  where(
+    grouper: (qb: CreateIndexWhereInterface<C>) => CreateIndexWhereInterface<C>
+  ): CreateIndexBuilder<C>
+  where(expression: Expression<any>): CreateIndexBuilder<C>
+
+  where(...args: any[]): any {
+    return new CreateIndexBuilder({
+      ...this.#props,
+      node: QueryNode.cloneWithWhere(
+        this.#props.node as any,
```
koskimas, line 183: Instead of casting to `any` change the `cloneWithWhere` interface from:

```ts
  cloneWithWhere<T extends FilterableQueryNode>(
    node: T,
    operation: OperationNode
  ): T
```

to something like this:

```ts
  cloneWithWhere<T extends { where?: WhereNode }>(
    node: T,
    operation: OperationNode
  ): T
```

src/parser/binary-operation-parser.ts
```diff
@@ -100,6 +101,14 @@ export function parseFilterExpression(
   throw createFilterExpressionError(type, args)
 }
 
+export function parseWhereWithParametersAsLiterals(args: any[]): OperationNode {
+  if (args.length === 3) {
+    args = [args[0], args[1], sql.literal(args[2])]
```
koskimas, line 106: Instead of `sql.literal` you can use `ValueNode.createImmediate`.

igalklebanov: Good point! forgot this thing exists.

src/schema/create-index-builder.ts
igalklebanov: Good idea!

PR #375, opened 2023-03-21
add column sort order @ `CreateIndexBuilder`.

### LINKED ISSUE #301: Creating a Sorted Index
In a migration, I write
```ts
await db.schema.createIndex("user_created_at")
    .on("user")
    .columns(["createdAt"])
    .execute();
```

To build a simple index on `createdAt`.

This doesn't help when doing a range query on `createdAt`, so I want to create a sorted index to help with that.

Looking at the source, I _think_ the following is not supported (because `"createdAt DESC"` is not a column name):
```ts
await db.schema.createIndex("user_created_at")
    .on("user")
    .columns(["createdAt DESC"])
    .execute();
```

From the docs, there is `expression`, which might help with that:
```ts
await db.schema.createIndex('user_created_at')
    .on("user")
    .expression(sql`createdAt DESC`)
    .execute();
```

Is this the way to go or is there a better solution? I'm trying to be compatible with PG, SQLite and MySQL/MariaDB, so I try to avoid raw SQL.

igalklebanov:
closes #301.

This PR adds the ability to create sorted indexes in all built-in dialects, by simply adding `asc`/`desc` after column names.

Extra enhancement: can now chain multiple `.column`, `.columns` & `.expression` calls to append additional columns/expressions to index.


src/parser/reference-parser.ts
```diff
@@ -147,6 +162,18 @@ export function parseColumnName(column: AnyColumn<any, any>): ColumnNode {
   return ColumnNode.create(column as string)
 }
 
+export function parseOrderedColumnName(column: string): OperationNode {
+  const ORDER_SEPARATOR = ' '
+
+  if (column.includes(ORDER_SEPARATOR)) {
+    const [columnName, order] = column.split(ORDER_SEPARATOR).map(trim)
+
+    return parseOrderBy(columnName, order as any)
```
koskimas, line 171: You could check the `order` argument instead of casting to any:

```ts
function isOrder(order: string): order is 'asc' | 'desc'
```

src/parser/reference-parser.ts
```diff
@@ -147,6 +162,18 @@ export function parseColumnName(column: AnyColumn<any, any>): ColumnNode {
...
+  } else {
+    return parseColumnName(column as any)
```
koskimas, line 173: Doesn't `parseColumnName` accept a string? No cast needed here?

src/parser/reference-parser.ts
```diff
@@ -76,6 +77,20 @@ export type ExtractTypeFromStringReference<
   ? ExtractColumnType<DB, TB, RE>
   : DV
 
+export type OrderedColumnName<C extends string> =
```
koskimas, line 80: This is identical to `ExtractColumnNameFromOrderedColumnName`. Shouldn't this be just

```ts
export type OrderedColumnName<C extends string> = C | `${C} ${OrderByDirection}`
```

?

koskimas (approved):
Couple of comments, LGTM otherwise.

src/parser/reference-parser.ts
igalklebanov: Good point, Added the guard to `order-by-parser` next to `OrderByDirection` type.

src/parser/reference-parser.ts
igalklebanov: True, I don't know why I used `as` there. Maybe the language server was f'ing up.

src/parser/reference-parser.ts
igalklebanov: They're very close. 
`OrderedColumnName` type-guards and returns as-is.
`ExtractColumnNameFromOrderedColumnName` extracts the column name.

The suggestion behaves differently, check [this](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcDyUAmwrEwYQgBsBXAWwDsA5AQzOAB585QZgLMBnOTmKASwoBzAHxwAvAFgAUHDjNW7LnAAGAEgDevAcIC+cTYIBmOdLpUy5cgPzoWINh24ZsUAEKIAIv1wBjGPwQFJZWNvIhoQBccBTAAG44EXDR+DIyoJCwCMioLjh4hKSUtPQATEz2jsragqIS8nAAPqqa+Pqaee5ePsD+gRTmadJIKOhYOB7efgFB9QDkNJy+c01wc9hLc0NGJBR9s0YUTKWVStw1wgA0DYpOPHy1IgAU+KXRnQXE5NR0wOWvImu+He41wBC+xV+TBEAEoNLpthQnnMAFYQAAWFDm11RGKxMIA3DJDsi0Zi4BtljiyRQKcBNjDiUjceTKXTNtS8ezfNy5ozpCSWbTOGR+DB0bzOeSRWKJZS+UA) out.

koskimas:
Feel free to rebase and merge.

PR #378, opened 2023-03-23
DRY up `explain` docs and clone logic.

igalklebanov:


src/query-builder/delete-query-builder.ts
```diff
@@ -956,7 +956,7 @@ export class DeleteQueryBuilder<DB, TB extends keyof DB, O>
 
     return result as SimplifyResult<O>
   }
-
+
```
koskimas, line 959: Some extra whitespace here

src/operation-node/query-node.ts
```diff
@@ -19,6 +22,8 @@ type HasJoins = { joins?: ReadonlyArray<JoinNode> }
 type HasWhere = { where?: WhereNode }
 type HasReturning = { returning?: ReturningNode }
 
+export type HasExplain = { explain?: ExplainNode }
```
koskimas, line 25: Why export?

igalklebanov: Haven't updated after the other types were un-exported.

src/query-builder/delete-query-builder.ts
igalklebanov: Nice catch!

PR #388, opened 2023-03-26
align `sql` template tag with `ExpressionBuilder`.

igalklebanov:
- deprecate `Sql.value`, for `Sql.val`.
- add `.literal` @ `ExpressionBuilder`.
- add generics to various `Sql` methods.

### DIFF
--- a/src/expression/expression-builder.ts
+++ b/src/expression/expression-builder.ts
@@ -223,6 +223,35 @@ export interface ExpressionBuilder<DB, TB extends keyof DB> {
   ): ExpressionWrapper<ExtractTypeFromValueExpressionOrList<VE>>
 
   /**
+   * This can be used to add literal values to SQL snippets.
+   *
+   * WARNING! Using this with unchecked inputs WILL lead to SQL injection
+   * vulnerabilities. The input is not checked or escaped by Kysely in any way.
+   * You almost always want to use normal substitutions that get sent to the
+   * db as parameters.
+   *
+   * ```ts
+   * const firstName = 'first_name'
+   *
+   * db.selectFrom('dela')
+   *   .where(({ cmp, literal, ref }) => cmp(literal(3), '=', ref('magic_number')))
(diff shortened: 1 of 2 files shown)


koskimas:
Good catches! Should `literal` be `lit` to use the same naming?

igalklebanov:
> Should `literal` be `lit` to use the same naming? 

Yeah, let's go with `lit`.

> I'm not sure we should have `lit` or `literal` in the `ExpressionBuilder` though. It's a dangerous method.

We could take a page out of react's playbook and call it `dangerouslyLit` 😃

koskimas:
Why add it at all if it's deprecated from the get go?

koskimas:
By the way, there's a [discussion](https://discord.com/channels/890118421587578920/1073948476590792786/threads/1085891226546470953) where people are already asking for a better, longed name for `cmp` 😮‍💨

igalklebanov:
> Why add it at all if it's deprecated from the get go? Or dangerous. The user always has `sql` for dangerous stuff.

Just thought it would have some QoL benefits (not having to import sql). 🤷‍♂️

koskimas:
> Just thought it would have some QoL benefits (not having to import sql). 🤷‍♂️

That's a good point. But I don't think users will need to use `lit` that often. It's only useful in migrations and some other very specific places. I think it's ok to import `sql` in those cases. That way we'll keep the dangerous stuff in one place.

igalklebanov:
> > Just thought it would have some QoL benefits (not having to import sql). 🤷‍♂️
> 
> That's a good point. But I don't think users will need to use `lit` that often. It's only useful in migrations and some other very specific places. I think it's ok to import `sql` in those cases. That way we'll keep the dangerous stuff in one place.

Yeah, its rare with the built-in dialects, but from what I've seen, 3rd party dialects might require using it more.

koskimas:
Noice! Thanks!

PR #389, opened 2023-03-27
Fix playground iframe of site.

wirekang:
## Minor changes
Pass `allow="clipboard-write" to playground iframe to solve the permission error:
image

Default kysely version: 0.23.3 -> 0.23.5


fhur:
Thanks for the fix @wirekang. And by the way, thanks for implementing the playground :) 

LGTM.

fhur:
We'll need a better mechanism for versioning in the docs, so this doesn't happen again in the future, but that's a bigger topic.

PR #399, opened 2023-03-29
fix explicitly `undefined` values not falling back to default @ `InsertQueryBuilder.values`.

### LINKED ISSUE #398: SQL parser bug
First of all, thank you for writing this lightweight TS safe library. We're migrating our website from MongoDB to PostgreSQL. We tried Prisma and it was a hot pile of disaster in terms on performance. 

However, while setting up a few scripts, we found a bug inside this library. Our hands our currently full otherwise would have loved to jump into source and send a PR fix too. Here's how you can reproduce the bug: 

1. Go to https://codedamn.com/playground/4rv-jRRmU-BdpcEtjKm-0
2. Click on "Run code" button on top right
3. Go through the comments in the codebase

igalklebanov:
closes #398.

- opened multi-row inserts with defaults test cases to mysql as well.
- also removes some unused imports.

### DIFF
--- a/src/parser/insert-values-parser.ts
+++ b/src/parser/insert-values-parser.ts
@@ -116,14 +116,11 @@ function parseRowValues(
   const hasMissingColumns = rowColumns.length < columns.size
 
   if (hasMissingColumns || hasUndefinedOrComplexColumns) {
-    let defaultValueNode: DefaultInsertValueNode
+    const defaultValue = DefaultInsertValueNode.create()
 
     return ValueListNode.create(
-      rowValues.map((value) =>
-        isUndefined(value)
-          ? (defaultValueNode =
-              defaultValueNode || DefaultInsertValueNode.create())
-          : parseValueExpression(value)
+      rowValues.map((it) =>
+        isUndefined(it) ? defaultValue : parseValueExpression(it)
       )
     )
   }


src/parser/insert-values-parser.ts
```diff
@@ -91,34 +91,39 @@ function parseRowValues(
...
-    const defaultValue = DefaultInsertValueNode.create()
+  if (hasMissingColumns || hasUndefinedOrComplexColumns) {
+    let defaultValueNode: DefaultInsertValueNode
 
     return ValueListNode.create(
-      rowValues.map((it) =>
-        isUndefined(it) ? defaultValue : parseValueExpression(it)
+      rowValues.map((value) =>
+        isUndefined(value)
+          ? (defaultValueNode =
```
koskimas, line 124: Move the assignment away from the ternary. Do we really need this optimization? I don't think the old way caused any performance problems. We could even create a new instance each time and ditch the `defaultValueNode` variable.

src/parser/insert-values-parser.ts
```diff
@@ -91,34 +91,39 @@ function parseRowValues(
...
 
-  if (columnMissing || complexColumn) {
-    const defaultValue = DefaultInsertValueNode.create()
+  if (hasMissingColumns || hasUndefinedOrComplexColumns) {
+    let defaultValueNode: DefaultInsertValueNode
 
     return ValueListNode.create(
-      rowValues.map((it) =>
-        isUndefined(it) ? defaultValue : parseValueExpression(it)
+      rowValues.map((value) =>
```
koskimas, line 122: Quite a lot of opinionated changes style changes. You'll have a lot of work ahead of you if you plan to get rid of all `it` variables. It has been used has the shorthand for `iterator` pretty much everywhere.

koskimas (approved):

src/parser/insert-values-parser.ts
igalklebanov: Sorry, habits. 😢

src/parser/insert-values-parser.ts
igalklebanov: I'll revert to the old way.

PR #401, opened 2023-03-30
Fix playground iframe.

wirekang:
Very little changes.

1. Remove fixed default value of `kyselyVersion` to use latest version.
2. Trim example codes to prevent annoying(at least for me) empty line.
image

By the way, the StackBlitz version of playground in [kysely.dev](https://kysely.dev/) looks nice and I'll do what I can do until all playground replaced with it :(


wirekang:
It seems something wrong, maybe because I reuse merged branch?

igalklebanov (approved):

igalklebanov:
> It seems something wrong, maybe because I reuse merged branch?

Is it good to go now?

koskimas:
Thank you!

> By the way, the StackBlitz version of playground in [kysely.dev](https://kysely.dev/) looks nice and I'll do what I can do until all playground replaced with it

There's no plan to start using it everywhere. It was added by @fhur, probably because its look fits the front page better.

It's important to me that the playground we use provides the same experience as vscode. Using your version might just be the way to have that control.

wirekang:
> > It seems something wrong, maybe because I reuse merged branch?
> 
> Is it good to go now?

Yes.

PR #404, opened 2023-03-31
add `case` statements/operators.

### LINKED ISSUE #214: `CASE...WHEN...THEN...END` Support
This is a bigger one, but, any plans on implementing CASE support?

https://modern-sql.com/feature/case

Looks like it is supported by: BigQuery, Db2, MariaDB, MySQL, Oracle DB, PostgreSQL, SQL Server, SQLite, so basically all major engines.

Syntax:

```
CASE
    WHEN condition1 THEN result1
    WHEN condition2 THEN result2
    WHEN conditionN THEN resultN
    ELSE result
END;
```

Probably would fit well as another method on ExpressionBuilder, ie:

```
db.selectFrom("table")
 .select(eb => {
  return eb.case(cb => {
   return cb.when("expression1", "expression2")
    .when("expression3", "expression4")
    .else("expression5")
  })
 })
```

igalklebanov:
closes #214.

- [x] implement
- [x] test
- [x] typings test
- [x] document


src/query-builder/case-builder.ts
```diff
@@ -0,0 +1,162 @@
...
+export class CaseBuilder<DB, TB extends keyof DB, W = unknown, O = never>
+  implements Whenable<DB, TB, W, O>
+{
+  readonly #props: CaseBuilderProps
+
+  constructor(props: CaseBuilderProps) {
+    this.#props = freeze(props)
+  }
+
+  when(expression: Expression<W>): CaseThenBuilder<DB, TB, W, O> {
```
koskimas, line 22: Wondering if we should also have a shortcut for `cmpr`: `when('a', 'op', 'b')`? Thoughts?

src/expression/expression-builder.ts
```diff
@@ -150,6 +155,64 @@ export interface ExpressionBuilder<DB, TB extends keyof DB> {
...
+   *
+   * const { title, name } = await db
+   *   .selectFrom('person')
+   *   .where('id', '=', '123')
+   *   .select((eb) => [
+   *     'first_name',
+   *     eb
+   *       .case()
+   *       .when(eb.cmpr('gender', '=', 'male'))
+   *       .then(sql.lit('Mr.'))
```
koskimas, line 176: The examples are full of `sql.lit` which feels wrong for some reason. I still don't feel good about adding `eb.lit` though 🤔 

I'm guessing `eb.val` doesn't work here without SQL casts?

src/operation-node/case-node.ts
```diff
@@ -0,0 +1,59 @@
...
+    return freeze({
+      kind: 'CaseNode',
+      value,
+    })
+  },
+
+  cloneWithWhen(caseNode: CaseNode, when: WhenNode): CaseNode {
+    return freeze({
+      ...caseNode,
+      when: caseNode.when ? [...caseNode.when, when] : [when],
```
koskimas, line 31: `freeze` the lists, since that's what's done in other nodes. I'm been thinking that maybe we could live without freezing stuff. But until we drop it, let's be consistent.

src/operation-node/case-node.ts
```diff
@@ -0,0 +1,59 @@
...
+    })
+  },
+
+  cloneWithThen(caseNode: CaseNode, then: OperationNode): CaseNode {
+    return freeze({
+      ...caseNode,
+      when: caseNode.when
+        ? [
```
koskimas, line 39: freeze list here too

src/query-builder/case-builder.ts
igalklebanov: Great idea!

I'm also adding the single value variant (case...value...when...then...end).

src/expression/expression-builder.ts
igalklebanov: I tried to mimmick how I'd write this case example in SQL without a query builder. 
I'd use literals in WHEN and THEN unless its user input, because paramters here feel odd.

FYI, I actually wrote `'male'` because right now, `cmpr` doesn't accept `sql.lit` (which is an issue for another PR).

src/operation-node/case-node.ts
igalklebanov: Good catch!

We could drop these and add a strong disclaimer in the docs - if you hack it, we don't bail you out when you f things up.

koskimas (approved):
Simply amazing work! 💯

PR #409, opened 2023-04-03
optional `constraintName` in CreateTableBuilder/AlterTableBuilder methods

kraftwerk28:
It is anyway optional in `*ConstraintNode` constructors, so I'd not need to pass an ambiguous empty string in case I want Postgres to automatically name a constraint for me:

```sql
create table "tbl" (
  "id" bigint not null,
  unique ("id")
)
```

### DIFF
--- a/src/schema/alter-table-builder.ts
+++ b/src/schema/alter-table-builder.ts
@@ -155,10 +155,16 @@ export class AlterTableBuilder implements ColumnAlteringInterface {
   /**
    * See {@link CreateTableBuilder.addUniqueConstraint}
    */
+  addUniqueConstraint(columns: string[]): AlterTableExecutor
   addUniqueConstraint(
     constraintName: string,
     columns: string[]
+  ): AlterTableExecutor
+  addUniqueConstraint(
+    ...args: [string[]] | [string, string[]]
   ): AlterTableExecutor {
+    const constraintName = args.length === 1 ? undefined : args[0]
+    const columns = args.length === 1 ? args[0] : args[1]
     return new AlterTableExecutor({
       ...this.#props,
       node: AlterTableNode.cloneWithTableProps(this.#props.node, {
@@ -173,9 +179,17 @@ export class AlterTableBuilder implements ColumnAlteringInterface {
    * See {@link CreateTableBuilder.addCheckConstraint}
    */
   addCheckConstraint(
+    checkExpression: Expression<any>
+  ): AlterTableExecutor
+  addCheckConstraint(
     constraintName: string,
     checkExpression: Expression<any>
+  ): AlterTableExecutor
+  addCheckConstraint(
+    ...args: [string, Expression<any>] | [Expression<any>]
   ): AlterTableExecutor {
+    const constraintName = args.length === 1 ? undefined : args[0]
+    const checkExpression = args.length === 1 ? args[0] : args[1]
     return new AlterTableExecutor({
       ...this.#props,
       node: AlterTableNode.cloneWithTableProps(this.#props.node, {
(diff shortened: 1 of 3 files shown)


src/schema/create-table-builder.ts
```diff
@@ -162,7 +162,7 @@ export class CreateTableBuilder<TB extends string, C extends string = never>
    * ```
    */
   addPrimaryKeyConstraint(
-    constraintName: string,
+    constraintName?: string,
     columns: C[]
   ): CreateTableBuilder<TB, C> {
```
igalklebanov, line 167: This is not valid TypeScript.

You should split to multiple overloads as follow:

```ts
addPrimaryKeyConstraint(constraintName: string, columns: C[]): CreateTableBuilder<TB, C>
addPrimaryKeyConstraint(columns: C[]): CreateTableBuilder<TB, C>

addPrimaryKeyConstraint(...args: any[]): any {
  // implementation
}
```

igalklebanov (changes requested):
Left some comments.

Missing handling of same functionality in `AlterTableBuilder`.
Missing tests.

src/schema/alter-table-builder.ts
```diff
@@ -201,7 +215,19 @@ export class AlterTableBuilder implements ColumnAlteringInterface {
...
+    targetTable: string,
+    targetColumns: string[]
+  ): AlterTableAddForeignKeyConstraintBuilder
+  addForeignKeyConstraint(
+    ...args: [string, string[], string, string[]] | [string[], string, string[]]
   ): AlterTableAddForeignKeyConstraintBuilder {
+    const constraintName = args.length === 4 ? args[0] : undefined
+    const columns = args.length === 4 ? args[1] : args[0]
+    const targetTable = args.length === 4 ? args[2] : args[1]
+    const targetColumns = args.length === 4 ? args[3] : args[2]
```
igalklebanov, line 230: This is painful to read, not sure what can be done here.

src/schema/create-table-builder.ts
```diff
@@ -263,8 +283,22 @@ export class CreateTableBuilder<TB extends string, C extends string = never>
...
+    build?: ForeignKeyConstraintBuilderCallback
+  ): CreateTableBuilder<TB, C>
+  addForeignKeyConstraint(...args: any[]): CreateTableBuilder<TB, C> {
+    const constraintName = typeof args[0] === 'string' ? args[0] : undefined
+    const columns: C[] = typeof args[0] === 'string' ? args[1] : args[0]
+    const targetTable: string = typeof args[0] === 'string' ? args[2] : args[1]
+    const targetColumns: string[] =
+      typeof args[0] === 'string' ? args[3] : args[2]
+    const build: ForeignKeyConstraintBuilderCallback =
+      (typeof args[0] === 'string' ? args[4] : args[3]) ?? noop
```
igalklebanov, line 301: Even more painful.

kraftwerk28: How about changing the signature completely (the PR is anyway marked as label:breaking)? For example,

```typescript
addForeignConstraint(
  columns: C[],
  targetTable: string,
  targetColumns: string[],
  build: ForeignKeyConstraintBuilderCallback = noop,
  constraintName?: string,
)
```

or even

```typescript
addForeignConstraint(opts: {
  columns: C[]
  targetTable: string
  targetColumns: string[]
  build?: ForeignKeyConstraintBuilderCallback
  constraintName?: string
})
```
?

igalklebanov: We're trying to be as close as possible to SQL, JSON inputs are distant from that.
I believe we should add `.references(table: string, columns: string[])` to `ForeignKeyConstraintBuilderInterface`,
and make the FK constraint builder callback a required argument.

So the problematic functions look like:

```ts
db.schema.createTable('table1').addForeignKeyConstraint(
  'constraintName',
  ['col1'],
  (qb) => qb.references('table2', ['col1'])
)
```

```ts
db.schema.createTable('table1').addForeignKeyConstraint(
  ['col1'],
  (qb) => qb.references('table2', ['col1'])
)
```

@koskimas WDYT?

kraftwerk28: The `references` clause is required in SQL (unlike `on delete`/`on update`), so when using it in a builder, we'll need some additional types for builder where **not** calling `.references` on it, the compilation shall fail. I'd lean towards optional builder and...
```ts
db.schema.alterTable('table1').addForeignKeyConstraint(
  ['table1_col'],
  'table2',
  ['table2_col'],
  qb => qb.constraintName('constraintName'), /* .onDelete(...) .onUpdate(...) etc */
)
```

igalklebanov: > The references clause is required in SQL, unlike on delete/on update (at least in postgres), so when using it in a builder, we'll need some additional types on its methods where not calling .references shall fail the compilation.

We've done it before, most common is requiring aliased builders.
The callback's argument is a builder with nothing but a `.references` method.
The callback's expected return type is a different builder that is returned from `.references` invocation.

The API would be much more readable this way.

> I'd lean towards optional builder and...

Yeah losing the overload is a plus.
But the order of SQL is not aligned with the order of reading/writing this invocation, which is not intuitive.

src/schema/alter-table-builder.ts
Loskir: Maybe `const [columns, targetTable, targetColumns] = args.length === 4 ? args.slice(1) : args`

koskimas:
This PR seems to be dead. Closing.

PR #416, opened 2023-04-06
feat: support `char` datatype

### LINKED ISSUE #339: CHAR datatype
Is support for `CHAR` datatype (fixed-length, blank padded) omitted on purpose? Both PostgreSQL and MySQL support that type.

Is it planned or will you accept PR on that?

anirudh1713:
closes #339 

### CONTEXT
This PR adds support for `char` datatype for -
- mysql
- postgresql

### Implementation
To achieve that we've added the `char` type to `ColumnDataType` type. This will allow consumers to use type `char`.
- Existing test have been updated to test the functionality.

### DIFF
--- a/src/operation-node/data-type-node.ts
+++ b/src/operation-node/data-type-node.ts
@@ -4,6 +4,7 @@ import { OperationNode } from './operation-node.js'
 export type ColumnDataType =
   | 'varchar'
   | `varchar(${number})`
+  | `char(${number})`
   | 'text'
   | 'integer'
   | 'int2'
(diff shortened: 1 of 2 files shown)


src/operation-node/data-type-node.ts
```diff
@@ -4,6 +4,7 @@ import { OperationNode } from './operation-node.js'
 export type ColumnDataType =
   | 'varchar'
   | `varchar(${number})`
+  | `char(${number})`
```
igalklebanov, line 7: What about just `char`?

igalklebanov (changes requested):
Looks good! 🚀 

Minor comment.

src/operation-node/data-type-node.ts
anirudh1713: I assumed `char` without size didn't existed as its fixed size. I apologize for that, added it now, and read the docs for both mysql and postgres and found that `char` is equivalent to `char(1)`.

Thanks for pointing this out @igalklebanov :+1:

igalklebanov (approved):
LGTM!

Issue #397, opened 2023-03-29
Support `ROW_NUMBER()`


mplibunao:
Hi

I'm trying to create the following window function

```sql
ROW_NUMBER() OVER (PARTITION BY t.id ORDER BY num_common_tags DESC, ct.is_primary DESC) AS tag_rank,
```

I've noticed that `OVER` is supported [here](https://kysely-org.github.io/kysely/classes/AggregateFunctionBuilder.html#over) but I can't find `ROW_NUMBER()` anywhere

I've also tried using raw sql but I'm getting the following type error

```
No overload matches this call.
  Overload 1 of 2, '(selections: readonly SelectExpression<From<DB, "communityTags"> & Record<"commonTags", Selection<From<DB, "communityTags">, "communityTags" | "userInterests", "communityTags.communityId" | AliasedAggregateFunctionBuilder<...>>>, FromTables<...> | "commonTags">[]): QueryBuilderWithSelection<...>', gave the following error.
    Type 'RawBuilder<number>' is not assignable to type 'SelectExpression<From<DB, "communityTags"> & Record<"commonTags", Selection<From<DB, "communityTags">, "communityTags" | "userInterests", "communityTags.communityId" | AliasedAggregateFunctionBuilder<...>>>, FromTables<...> | "commonTags">'.
      Type 'RawBuilder<number>' is missing the following properties from type 'AliasedExpression<any, any>': expression, alias
  Overload 2 of 2, '(selection: SelectExpression<From<DB, "communityTags"> & Record<"commonTags", Selection<From<DB, "communityTags">, "communityTags" | "userInterests", "communityTags.communityId" | AliasedAggregateFunctionBuilder<...>>>, FromTables<...> | "commonTags">): QueryBuilderWithSelection<...>', gave the following error.
    Argument of type '(string | RawBuilder<number>)[]' is not assignable to parameter of type 'SelectExpression<From<DB, "communityTags"> & Record<"commonTags", Selection<From<DB, "communityTags">, "communityTags" | "userInterests", "communityTags.communityId" | AliasedAggregateFunctionBuilder<...>>>, FromTables<...> | "commonTags">'. (tsserver 2769)
```

koskimas:
Kinda difficult to help without seeing the code that produced that error 😄

koskimas:
You need to use the `as` method to give an alias for the raw selection. The raw SQL is not parsed as a type:

```ts
[...]

PR #417, opened 2023-04-08
add `agg` method @ `FunctionModule` to allow `AggregateFunctionBuilder` usage outside of built-ins.

igalklebanov:
as discussed in #397.

This PR adds the ability to use `AggregateFunctionBuilder`'s methods without a built-in aggregate function method (e.g. sum), using `eg.fn.agg(name, [args])` which is inspired by `eb.fn(name, [args])`.

```ts
db.selectFrom('person')
  .select(({ fn }) => [
    fn.agg<number>('rank').over().as('rank'),
    fn.agg<string>('group_concat', ['first_name']).distinct().as('first_names')
  ])
```

breaking changes:

- internal: aggregate function node structure.

### DIFF
--- a/src/operation-node/aggregate-function-node.ts
+++ b/src/operation-node/aggregate-function-node.ts
@@ -1,17 +1,12 @@
 import { freeze } from '../util/object-utils.js'
 import { OperationNode } from './operation-node.js'
 import { OverNode } from './over-node.js'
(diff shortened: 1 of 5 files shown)


koskimas (approved):
Flawless yet again! You, sir, are a legend! 🎩

PR #431, opened 2023-04-19
Add UpdateResult.numChangedRows

### LINKED ISSUE #428: Update result only returning number of updated rows
Is there anyway I can get the number of changed rows and matched rows? [UpdateResult](https://kysely-org.github.io/kysely/classes/UpdateResult.html) only returns `numUpdatedRows` which is the number of changed rows, I want to have the number of matched but unchanged rows as well.

wirekang:
MySQL-only feature, `changedRows` is returned from [node-mysql2](https://github.com/sidorares/node-mysql2/blob/01cbaafc97f62742d294d682048e2baaec828462/typings/mysql/lib/protocol/packets/OkPacket.d.ts#L8)

There are some non-related file changes because some files in master branch does not prettified well.

Closes #428


wirekang:
Can I ignore the vercel deployment?

src/dialect/database-introspector.ts
```diff
@@ -35,7 +35,7 @@ export interface SchemaMetadata {
 export interface DatabaseMetadata {
   /**
    * The tables and views found in the database.
-*  * The propery isView can be used to tell them apart.
+   *  * The propery isView can be used to tell them apart.
```
koskimas, line 38: The second asterisk can be removed. It's a typo.

src/query-builder/update-query-builder.ts
```diff
@@ -881,7 +881,8 @@ export class UpdateQueryBuilder<DB, UT extends keyof DB, TB extends keyof DB, O>
     return [
       new UpdateResult(
         // TODO: remove numUpdatedOrDeletedRows.
-        result.numAffectedRows ?? result.numUpdatedOrDeletedRows ?? BigInt(0)
+        result.numAffectedRows ?? result.numUpdatedOrDeletedRows ?? BigInt(0),
+        result.numChangedRows ?? BigInt(0)
```
koskimas, line 885: Since `numChangedRows` is an optional field, it should default to `undefined` instead of 0. 0 value would confuse the users that don't happen to read the comment that this is a MySQL-only field.

src/query-builder/update-result.ts
```diff
@@ -1,15 +1,24 @@
...
   }
 
   get numUpdatedRows(): bigint {
     return this.#numUpdatedRows
   }
+
+  /**
+   * MySQL-only
+   */
+  get numChangedRows(): bigint {
```
koskimas, line 19: Make this optional and only give value on MySQL.

src/query-builder/update-result.ts
```diff
@@ -1,15 +1,24 @@
...
     this.#numUpdatedRows = numUpdatedRows
+    this.#numChangedRows = numChangedRows
   }
 
   get numUpdatedRows(): bigint {
     return this.#numUpdatedRows
   }
+
+  /**
+   * MySQL-only
```
koskimas, line 17: Instead of documenting this as `MySQL-only` document it as optional and that some dialects may not provide a value. Some 3rd party dialects might have this and then the documentation would be incorrect for those ones.

src/query-builder/update-result.ts
```diff
@@ -1,15 +1,24 @@
...
+   */
+  get numChangedRows(): bigint {
+    return this.#numChangedRows
+  }
 }
 
-Object.defineProperty(UpdateResult.prototype, 'numUpdatedRows', {
-  enumerable: true,
-})
+setEnumerableProperties(UpdateResult, 'numUpdatedRows', 'numChangedRows')
```
koskimas, line 24: The name of this functions sounds like

```
set the value of these enumerable properties
```

A better name would be `makePropertiesEnumerable`

src/driver/database-connection.ts
```diff
@@ -26,6 +26,11 @@ export interface QueryResult<O> {
    */
   readonly numAffectedRows?: bigint
 
+  /**
+   * A MySQL-only field contains the number of rows the query changed.
+   */
```
koskimas, line 31: Document as optional field instead of MySQL-only field. Some 3rd party dialects might set this.

koskimas (changes requested):
Thank you! Some comments, then this is good to go.

src/query-builder/update-query-builder.ts
wirekang: I did this for consistency. `numAffectedRows` and `numChangedRows` both are optional in `QueryResult` but not in `UpdateResult`.


`driver/database-connection.ts`
```ts
export interface QueryResult<O> {
  readonly numAffectedRows?: bigint
  readonly numChangedRows?: bigint
}
```

`query-builder/update-result.ts`
```ts
export class UpdateResult {
  readonly #numUpdatedRows: bigint
  readonly #numChangedRows: bigint

  constructor(numUpdatedRows: bigint, numChangedRows: bigint) {
    this.#numUpdatedRows = numUpdatedRows
    this.#numChangedRows = numChangedRows
  }
}
```

src/util/type-utils.ts
```diff
@@ -191,3 +191,7 @@ export type NarrowPartial<S, T> = {
 }
 
 export type SqlBool = boolean | 0 | 1
+
+export interface Type<T> extends Function {
```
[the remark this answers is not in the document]
wirekang: This is well-know utility type used by many typescript projects such as angular/nestjs. Since I don't want to be stubborn, I will change it to Constructor if you want, but please reconsider.

[Sourcegraph](https://sourcegraph.com/search?q=context:global+interface+Type%3CT%3E+lang:typescript+&patternType=standard&sm=1&groupBy=repo)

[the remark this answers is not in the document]
koskimas: https://sourcegraph.com/search?q=context:global+interface+Constructor%3CT%3E+lang:typescript+&patternType=standard&sm=1&groupBy=repo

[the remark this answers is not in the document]
wirekang: Haha the winner is Typescript

src/driver/database-connection.ts
koskimas: The documentation doesn't say it's optional now. It says the opposite. Optionality of the type does not imply optionality of implementation in this interface.

src/query-builder/update-result.ts
koskimas: You didn't make this optional.

koskimas (changes requested):

wirekang (commented):
reviewed

src/dialect/database-introspector.ts
wirekang, line 1: Unformatted file

src/parser/update-set-parser.ts
wirekang, line 1: Unformatted

wirekang:
@koskimas

I updated codes to fit with master branch. Changes simplified due to https://github.com/kysely-org/kysely/commit/f171a905bae8649ad54460da3b14bc9b5f32e6d5

igalklebanov (approved):
Went over @koskimas' comments, and changes made in response. Looks good to go on my end.

koskimas (approved):

drewbitt:
This is a breaking change for anyone who ever JSONifys an UpdateResult, as JSONify does not serialize BigInts. This is likely a common pattern and we had been doing this in our backend.

For anyone looking for a fix, you need to only JSONify numChangedRows if you were using that and not the whole UpdateResult.

PR #440, opened 2023-04-22
add json operator (`->`, `->>`) support.

igalklebanov:


koskimas:
I like the idea of a JSON path builder, but I'm not the biggest fan of the `jxp('column', '->>', callback)` syntax. Seems overly complex. Did you consider something like

```ts
jsref('column').key('foo').key('bar').at('baz')
```

you wouldn't be able to specify the operator, but could that be given as an optional second argument?

```ts
jsref('column', '->').key('foo').key('bar').at('baz')
```

igalklebanov:
> I like the idea of a JSON path builder, but I'm not the biggest fan of the `jxp('column', '->>', callback)` syntax. Seems overly complex. Did you consider something like
> 
> ```ts
> jsref('column').key('foo').key('bar').at('baz')
> ```
> 
> you wouldn't be able to specify the operator, but could that be given as an optional second argument?
> 
> ```ts
> jsref('column', '->').key('foo').key('bar').at('baz')
> ```
> 
> I'm probably missing some future use cases you had in mind for the `jxp` function...

Yeah I wasn't feeling `jxp` because of PostgreSQL. It would force wrapping hell to achieve something like `col ->> key ->> key`.

I like your suggestion, will play around with it more.

igalklebanov:
New idea pushed, still rough:

```ts
eb.ref('column', '->>').at(0).key('key').as('alias')
```

Should compile to:

- PostgreSQL

```sql
"column"->>0->>'key' as "alias"
```

- MySQL

```sql
`column`->>'$[0].key' as `alias`
```

- SQLite

```sql
"column"->>'$[0].key' as "alias"
```

igalklebanov:
Thinking about enhancing WYSIWYG by adding a `->>$` operator option for MySQL/SQLite, and removing dialect specific compilation. 🤔

igalklebanov:
Looking good thus far, need "more nested" test cases, and typing tests. 😌

igalklebanov:
Some updates:

- unit tests are pretty much done.
- array traversal should always be nullable. there's no guarantee an item `i` exists. we won't support tuples and readonly array literals - that just adds unnecessary type complexity right now.
- added `.at('last')` (supported in MySQL) and `.at('#-1')` (supported in SQLite) as ways to access items in the end of an array. we also support `.at(-1)` (supported in PostgreSQL).
- in case there's nested traversal in "multi-arrow" syntax, PostgreSQL forces that only the last operator can be `->>`, so we'll go with that by default, even though SQLite supports both forms.

igalklebanov:
Ready for review. 😮‍💨 

Will add more test cases, mainly for usage without `ColumnType` and usage in non-select clauses.

src/operation-node/json-path-reference-node.ts
```diff
@@ -0,0 +1,46 @@
+import { freeze } from '../util/object-utils.js'
+import { JSONPathNode } from './json-path-node.js'
+import { OperationNode } from './operation-node.js'
+import { JSONOperator } from './operator-node.js'
+
+type JSONOperatorNo$ = Exclude<JSONOperator, '->>$'>
```
koskimas, line 6: Should also exclude `->$` right?

src/query-compiler/default-query-compiler.ts
```diff
@@ -1327,6 +1337,53 @@ export class DefaultQueryCompiler
     }
   }
 
+  protected override visitJSONPathReference(node: JSONPathReferenceNode): void {
+    if (node.is$) {
+      this.append(node.operator)
+      this.visitNode(node.jsonPath)
+    } else {
```
koskimas, line 1344: This feels like a hack: `visitJSONPath` is not called at all if its parent doesn't have `is$: true`. Same for `visitJSONPathLeg`. 

I think we need to have a separate node completely for json paths and postgres-style nested references. The `visit` functions should always be called if there's a node of that kind in the tree.

If you're modelling this after some other similar case in the codebase, we should fix that too (I can do that).

src/expression/expression-builder.ts
```diff
@@ -353,6 +399,7 @@ export interface ExpressionBuilder<DB, TB extends keyof DB> {
    * Creates a binary expression.
    *
    * See {@link cmpr} for creating binary comparison operations.
+   * See {@link jxp} for creating JSON traversal operations.
```
koskimas, line 402: Outdated comment.

src/index.ts
```diff
@@ -183,6 +185,8 @@ export * from './operation-node/set-operation-node.js'
 export * from './operation-node/binary-operation-node.js'
 export * from './operation-node/unary-operation-node.js'
 export * from './operation-node/using-node.js'
+export * from './operation-node/json-path-leg-node.js'
+export * from './operation-node/json-path-node.js'
```
koskimas, line 189: Path reference node missing

src/plugin/parse-json-results/parse-json-results-plugin.ts
```diff
@@ -0,0 +1,41 @@
...
+    }
+  }
+}
+
+function parseRow(row: UnknownRow): UnknownRow {
+  return Object.entries(row).reduce<UnknownRow>((row, [key, value]) => {
+    if (isString(value) && value.match(/^[\[\{]/) != null) {
+      try {
+        value = JSON.parse(value)
+      } catch (err) {}
```
koskimas, line 34: Add a comment that the catch block is empty on purpose.

src/query-builder/json-path-builder.ts
```diff
@@ -0,0 +1,276 @@
...
+} from '../operation-node/json-path-leg-node.js'
+import { JSONPathNode } from '../operation-node/json-path-node.js'
+import { JSONPathReferenceNode } from '../operation-node/json-path-reference-node.js'
+import { isOperationNodeSource } from '../operation-node/operation-node-source.js'
+import { OperationNode } from '../operation-node/operation-node.js'
+import { RawNode } from '../operation-node/raw-node.js'
+import { ReferenceNode } from '../operation-node/reference-node.js'
+
+export class JSONPathBuilder<S, O = S> {
+  readonly #node: ReferenceNode | JSONPathNode
```
koskimas, line 16: Isn't this currently always a `ReferenceNode`? What's the use case for `JSONPathNode`?

`#createBuilderWithPathLeg` creates a new `ReferenceNode` if `#node` is a `ReferenceNode` and since it's a `ReferenceNode` initially --> it's always `ReferenceNode`.

src/query-builder/json-path-builder.ts
```diff
@@ -0,0 +1,276 @@
...
+        : JSONPathNode.cloneWithLeg(this.#node, legNode)
+    )
+  }
+}
+
+export class TraversedJSONPathBuilder<S, O>
+  extends JSONPathBuilder<S, O>
+  implements Expression<O>
+{
+  readonly #node: ReferenceNode | JSONPathNode
```
koskimas, line 186: Same here. Always a `ReferenceNode`?

koskimas (changes requested):
Amazing work!

The operation node structure needs some more work. Currently it's a little bit of hack.

Should we add tests that this works correctly with the `CamelCasePlugin`? The nested keys should be mapped depending on the `maintainNestedObjectKeys` config.

src/query-builder/json-path-builder.ts
```diff
@@ -0,0 +1,276 @@
...
+    return this.#createBuilderWithPathLeg('Member', key)
+  }
+
+  #createBuilderWithPathLeg(
+    legType: JSONPathLegType,
+    value: string | number
+  ): TraversedJSONPathBuilder<any, any> {
+    const legNode = JSONPathLegNode.create(
+      legType,
+      RawNode.createWithSql(String(value))
```
koskimas, line 165: I think this should be a `ValueNode.createImmediate` instead. That way it's somewhat transformable. Or am I missing something here?

src/operation-node/json-path-reference-node.ts
igalklebanov: Nice catch!

Fixed @ 713e97e5e955a333d12970f508d4cf85dfc54925.

src/expression/expression-builder.ts
igalklebanov: Damn, nice catch!

Removed @ 3f13b507a3fdb93e4902b39e5458a31c51c3b9ca.

src/index.ts
igalklebanov: Nice catch! damn missed so many of these. 😕

Added @ af8aae26acbe0916b7d8c5715c3702cf29bc22e2.

src/plugin/parse-json-results/parse-json-results-plugin.ts
igalklebanov: Good idea!

Added @ ec4a26d225e10df175d059a0e56ea82a7f69ded7.

src/query-builder/json-path-builder.ts
igalklebanov: I thought so too, but `ValueNode.createImmediate` wraps strings with `'`s and JSON path should be `->'$.key[last]'`.
Maybe we should create a `JSONPathLegValueNode`?

src/query-builder/json-path-builder.ts
igalklebanov: You're right.

The PR started off with just a `JSONPathBuilder` that could be used to generate `$.key[0]`, since this is a thing that can be used in many ways outside of `column->`, as a JSON function argument. Once I added the `column->` stuff, it became less important to support as a standalone, and eventually, I dropped the ability to use it that way for now - to focus the PR around `column->` which is way more common.

src/query-builder/json-path-builder.ts
igalklebanov: Simplified for now @ 4ab875d1587b027ee1090fa6fa9c5439a85fbc02.

src/query-compiler/default-query-compiler.ts
igalklebanov: I didn't like it either. Glad you noticed that as well.
I don't remember it being inspired by other things in Kysely's tree.
It just got changed a lot over time as I found out new things and new gotchas between the dialects.

Went back to the drawing board with d12e4478ac0c7164cbc51fa0649255b1213c5101.
Yeah, seriously went to a drawing board 😄 
image

igalklebanov:
> Should we add tests that this works correctly with the CamelCasePlugin? The nested keys should be mapped depending on the maintainNestedObjectKeys config.

Haven't thought about that one, will try adding these now.

src/query-compiler/default-query-compiler.ts
koskimas: Would this make all references produce a `JSONReferenceNode`? Or would `ref('foo.bar')` produce just a `ReferenceNode` and `ref('foo.bar', '->').key('baz')` would produce a JSONReferenceNode?

koskimas: Perfect!

igalklebanov: Currently if `op` is undefined, it produces a  `ReferenceNode`, otherwise it produces a  `JSONReferenceNode`.

koskimas:
> > Should we add tests that this works correctly with the CamelCasePlugin? The nested keys should be mapped depending on the maintainNestedObjectKeys config.
> 
> Haven't thought about that one, will try adding these now.

Maybe we should merge this first and implement the camelCase stuff as a separate PR? We could release a new version after we get this merged?

koskimas (approved):

igalklebanov:
> Maybe we should merge this first and implement the camelCase stuff as a separate PR? We could release a new version after we get this merged?

Yeah, let's go! 🚀

bombillazo:
Documentation in the Kysely docs site for this would be phenomenal!

PR #442, opened 2023-04-23
feat: fetch-driver

Andersgee:
The popular planetscale dialect relies on fetch where the driver sends post requests to a url that handles the query and responds with the result thus "edge-compatible".

This is the same idea but relies on superjson and not limited to mysql

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,6 +31,7 @@
         "playwright": "^1.22.2",
         "prettier": "^2.7.1",
         "sinon": "^14.0.0",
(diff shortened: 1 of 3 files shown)


Andersgee:
made a [package](https://github.com/andersgee/kysely-fetch-driver for it in case this does not belong in this repo

koskimas:
Thank you for the contribution! This makes a great 3rd party package, but it shouldn't be included in the core. Let's add a link to the package to the README (and the upcoming site).

PR #446, opened 2023-04-24
docs: use methods from expression builder callback instead of global instance

rafaumlemos:
Since Kysely is able to infer context correctly from a callback and provide increased type-safety (as explained [here](https://github.com/kysely-org/kysely/blob/master/site/docs/recipes/expressions.md)), I have decided to switch several use cases of expression builders to utilize callbacks instead of a global function.


igalklebanov:
Hey 👋 

Thanks!

Let's make sure it is aligned with the project's prettier configuration:

iirc having parenthesis around callback arguments.

```ts
(eb) => ...
```

koskimas:
There are also other issues with this PR. Please don't merge until I've reviewed it.

igalklebanov:
@koskimas 

> We should still have tests for the global module usage too. Now all those tests were removed.

Yeah, aggregate function tests explicitly tested both modes. I thought it was pretty obvious. 🤷

rafaumlemos:
Hi,

1. The parenthesis around the callback arguments have been fixed.
2. Regarding the tests, what do you guys suggest? Should I revert all test changes and create specific ones for testing the callback usage?

Thanks!

koskimas:
Yeah, revert the tests changes and this is good to go.

rafaumlemos:
@koskimas @igalklebanov 

fixes applied

igalklebanov (approved):
LGTM! 🚀

PR #462, opened 2023-05-04
make `SchemaModule` optional and injectable.

igalklebanov:
This PR makes `SchemaModule` optional and injectable when instantiating Kysely. 
If not injected and consumer tries accessing it, a runtime error is thrown.

99% of the time, DDL queries are executed in migrations.

A considerable chunk of our consumers use Kysely in serverless/edge/browser environments. 99% of the time only DML queries are being executed there. In such environments bundle sizes affect cold starts / load times. The bigger the bundle, the worse our consumers' user experience.

Builder pattern is great for DX, but is bad for tree-shaking. By making `SchemaModule` injectable and optionally supported, we help decrease our consumers bundle sizes and help them ship better user experiences.

Things considered but not done:

- export `SchemaModule` & builders from `kysely/schema` instead of `kysely`. Unlike helpers, `SchemaModule` belongs in "main" Kysely.
- make `db.schema` return type `never` if no `SchemaModule` was passed @ config. Compilation performance hit that's not worth it.

Todo:

- [ ] check bundle size without `SchemaModule` usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -427,7 +427,8 @@ import {
   Kysely,
   Migrator,
   PostgresDialect,
-  FileMigrationProvider
+  FileMigrationProvider,
+  SchemaModule,
 } from 'kysely'
 
 async function migrateToLatest() {
@@ -438,6 +439,7 @@ async function migrateToLatest() {
         database: 'kysely_test',
       })
     }),
+    SchemaModule,
   })
 
   const migrator = new Migrator({
--- a/site/docs/migrations.mdx
+++ b/site/docs/migrations.mdx
@@ -94,6 +94,7 @@ import {
   Migrator,
   PostgresDialect,
   FileMigrationProvider,
+  SchemaModule,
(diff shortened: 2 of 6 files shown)


koskimas:
This feels kinda weird. Should we instead make it completely separate. Something like:

```ts
const schema = new Schema(db)
```

?

igalklebanov:
This definitely weird, and your suggestion definitely removes that weirdness. 👏  
At the same time, we need to make sure our consumers, who might already have a bunch of migration files, don't get hurt by this.

koskimas:
Yeah we'd definitely need to first deprecate the current API and and so on. Not just break everything. But I think your suggestion is just too much of a hack to be a permanent part of the API. Is the saving in bundle size significant enough to justify this change in the first place?

igalklebanov:
Closing this as it smells by now. Might re-attempt something in the future.

PR #464, opened 2023-05-06
fix `fn.max`/`fn.min` not usable as `fn(...)` arguments.

igalklebanov:
Spotted by ddanielcruzz on discord.

```ts
 fn("round", [fn.min("ld.price_btc"), sql.lit(5)]).as("floor_price"), // TS does complain ❌
 fn("round", [fn.sum("s.volume"), sql.lit(2)]).as("volume"), // TS does not complain ✅
 fn("round", [fn.avg("s.avg_price"), sql.lit(5)]).as("avg_price"),// TS does not complain✅
```

When `.min`/`.max` are used as `fn(...)` arguments, their `O` generic defaults to `any` instead of `never` and doesn't allow to use `StringReference` as `.min`/`.max` argument.

The fix replaces `never` with `any` as the "consumer didn't explictly pick an `O` type" default `O` value and everything is fine again.

### DIFF
--- a/src/query-builder/function-module.ts
+++ b/src/query-builder/function-module.ts
@@ -13,7 +13,7 @@ import {
   parseReferenceExpressionOrList,
 } from '../parser/reference-parser.js'
 import { parseSelectAll } from '../parser/select-parser.js'
-import { Equals, IsNever } from '../util/type-utils.js'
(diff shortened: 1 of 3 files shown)


koskimas (approved):
LGTM

Issue #467, opened 2023-05-08
AlteredColumnBuilder incorrectly joins operations


soanvig:
Having

```ts
db
    .schema
    .alterTable('xxx')
    .alterColumn('id', c => c.dropDefault().setNotNull())
```

Produces SQL:

```ts
alter table "xxx" alter column "id" drop defaultset not null
```

Resulting in error:

```ts
syntax error at or near "defaultset"
```

igalklebanov:
Hey 👋 

Nice catch! 💪 

Do you want to submit a PR?

soanvig:
Yep, as it is confirmed, I will create PR

PR #468, opened 2023-05-08
fix: disallow AlterColumn method chaining

soanvig:
Fixes: #467

It was slightly tricky, but I decided that "unwinding" looks the best in terms of clarity of what is happening.

This effectively produces query:

```
ALTER TABLE "x" ALTER COLUMN "y" SET DEFAULT true, ALTER COLUMN "z" DROP NOT NULL, [...]
```

as in Postgres docs. MySQL has exact syntax for chaining ALTER COLUMN, but supports only `drop default` and `set default` operations, so probably nobody will chain those two together, plus alter column usage for MySQL is not properly handled in whole Kysely anyway. I tested MySQL case manually.

---

EDIT: now it just disallows chaining

### DIFF
--- a/src/operation-node/alter-column-node.ts
+++ b/src/operation-node/alter-column-node.ts
@@ -1,8 +1,6 @@
 import { OperationNode } from './operation-node.js'
 import { freeze } from '../util/object-utils.js'
 import { ColumnNode } from './column-node.js'
-import { DataTypeNode } from './data-type-node.js'
-import { ValueNode } from './value-node.js'
 import { RawNode } from './raw-node.js'
 
 export type AlterColumnNodeProps = Omit<AlterColumnNode, 'kind' | 'column'>
@@ -26,20 +24,11 @@ export const AlterColumnNode = freeze({
     return node.kind === 'AlterColumnNode'
   },
 
-  create(column: string): AlterColumnNode {
+  create<T extends keyof AlterColumnNode>(column: string, prop: T, value: Required<AlterColumnNode>[T]): AlterColumnNode {
     return freeze({
       kind: 'AlterColumnNode',
       column: ColumnNode.create(column),
-    })
-  },
-
-  cloneWith(
-    node: AlterColumnNode,
-    props: AlterColumnNodeProps
-  ): AlterColumnNode {
-    return freeze({
-      ...node,
-      ...props,
+      [prop]: value
     })
   },
 })
--- a/src/operation-node/alter-column-node.ts
+++ b/src/operation-node/alter-column-node.ts
@@ -24,7 +24,7 @@ export const AlterColumnNode = freeze({
     return node.kind === 'AlterColumnNode'
   },
 
-  create<T extends keyof AlterColumnNode>(column: string, prop: T, value: Required<AlterColumnNode>[T]): AlterColumnNode {
+  create<T extends keyof AlterColumnNodeProps>(column: string, prop: T, value: Required<AlterColumnNodeProps>[T]): AlterColumnNode {
     return freeze({
       kind: 'AlterColumnNode',
       column: ColumnNode.create(column),
--- a/src/schema/alter-column-builder.ts
+++ b/src/schema/alter-column-builder.ts
@@ -10,39 +10,39 @@ import {
 } from '../parser/default-value-parser.js'
 
 export class AlterColumnBuilder {
-  protected column: string
+  readonly #column: string
 
   constructor(column: string) {
-    this.column = column
+    this.#column = column
   }
 
   setDataType(dataType: DataTypeExpression): AlteredColumnBuilder {
     return new AlteredColumnBuilder(
-      AlterColumnNode.create(this.column, 'dataType', parseDataTypeExpression(dataType))
+      AlterColumnNode.create(this.#column, 'dataType', parseDataTypeExpression(dataType))
     )
   }
 
   setDefault(value: DefaultValueExpression): AlteredColumnBuilder {
     return new AlteredColumnBuilder(
-      AlterColumnNode.create(this.column, 'setDefault', parseDefaultValueExpression(value))
+      AlterColumnNode.create(this.#column, 'setDefault', parseDefaultValueExpression(value))
     )
   }
 
   dropDefault(): AlteredColumnBuilder {
     return new AlteredColumnBuilder(
-      AlterColumnNode.create(this.column, 'dropDefault', true)
+      AlterColumnNode.create(this.#column, 'dropDefault', true)
     )
   }
 
   setNotNull(): AlteredColumnBuilder {
     return new AlteredColumnBuilder(
-      AlterColumnNode.create(this.column, 'setNotNull', true)
+      AlterColumnNode.create(this.#column, 'setNotNull', true)
     )
   }
 
   dropNotNull(): AlteredColumnBuilder {
     return new AlteredColumnBuilder(
-      AlterColumnNode.create(this.column, 'dropNotNull', true)
+      AlterColumnNode.create(this.#column, 'dropNotNull', true)
     )
   }
 
@@ -56,9 +56,7 @@ export class AlterColumnBuilder {
 }
 
 /**
- * Allows us to force consumers to do something, anything, when altering a column,
- * and also disallows chaining more methods on AlterColumnBuilder, because
- * SQL syntax for chaining ALTER COLUMN operations would be not WYSIWYG
+ * Allows us to force consumers to do exactly one alteration to a column.
  *
  * Basically, deny the following:
  *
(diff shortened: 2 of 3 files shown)


igalklebanov:
Hey 👋 

I understand what you were going for with this and it does seem convenient, but it is not aligned with Kysely's WYSIWYG design principle. Not sure about these changes.

We should consider either:

1. each method returns a builder that cannot chain further and `.alterColumn` expects it as argument. So the original issue is blocked by more precise API.

2. make compiler just add a whitespace between each operation, and keep it a single `alter column` clause. This'll cause a runtime error obviously with PostgreSQL/MySQL - but consumer should know SQL. This might provide future-proofing in case some dialect supports this.

igalklebanov:
> plus alter column usage for MySQL is not properly handled in whole Kysely anyway. I tested MySQL case manually.

Can you explain this? does it require an issue?

soanvig:
@igalklebanov  that's rather new to me that Kysely pushes WYSIWYG to such extent.
I can wait for other people decision on that topic, sure.

---

> Can you explain this? does it require an issue?

No. I was just saying, that you can call `AlterTableBuilder.dropNotNull` even if it is not supported in MySQL (and you will get runtime error)

igalklebanov:
> No. I was just saying, that you can call AlterTableBuilder.dropNotNull even if it is not supported in MySQL (and you will get runtime error)

Oh, gotcha. The API is dialect agnostic by design.

koskimas:
> each method returns a builder that cannot chain further and .alterColumn expects it as argument. So the original issue is blocked by more precise API.

@igalklebanov Yes, I agree. This is the way to proceed.

soanvig:
@koskimas @igalklebanov 
So for fun I'll implement that part, and we'll see. I really like this library, and we heavily use it at work, so I'm glad to contribute something.

soanvig:
@igalklebanov 

I updated the PR. Above you can see intelisense suggestion (type compilation result).
Besides restricting the result, I also narrowed API of `AlterColumnNode`, so it cannot contain more than one operation at once (since it won't be handled properly anyway): I removed `cloneWith`, because it was not needed, and creating the node now requires passing the property to be set.

src/operation-node/alter-column-node.ts
```diff
@@ -26,20 +24,11 @@ export const AlterColumnNode = freeze({
     return node.kind === 'AlterColumnNode'
   },
 
-  create(column: string): AlterColumnNode {
+  create<T extends keyof AlterColumnNode>(column: string, prop: T, value: Required<AlterColumnNode>[T]): AlterColumnNode {
```
igalklebanov, line 27: You're allowing setting `'kind'` and `'column'` as prop.

```suggestion
  create<P extends keyof AlterColumnNodeProps>(column: string, prop: P, value: Required<AlterColumnNode>[P]): AlterColumnNode {
```

src/schema/alter-column-builder.ts
```diff
@@ -10,49 +10,39 @@ import {
 } from '../parser/default-value-parser.js'
 
 export class AlterColumnBuilder {
-  protected readonly alterColumnNode: AlterColumnNode
+  protected column: string
```
igalklebanov, line 13:
```suggestion
  readonly #column: string
```

src/schema/alter-column-builder.ts
```diff
@@ -10,49 +10,39 @@ import {
...
 
-  constructor(alterColumnNode: AlterColumnNode) {
-    this.alterColumnNode = alterColumnNode
+  constructor(column: string) {
+    this.column = column
```
igalklebanov, line 16:
```suggestion
    this.#column = column
```

src/schema/alter-column-builder.ts
```diff
@@ -10,49 +10,39 @@ import {
...
   }
 
   setDataType(dataType: DataTypeExpression): AlteredColumnBuilder {
     return new AlteredColumnBuilder(
-      AlterColumnNode.cloneWith(this.alterColumnNode, {
-        dataType: parseDataTypeExpression(dataType),
-      })
+      AlterColumnNode.create(this.column, 'dataType', parseDataTypeExpression(dataType))
```
igalklebanov, line 21:
```suggestion
      AlterColumnNode.create(this.#column, 'dataType', parseDataTypeExpression(dataType))
```

src/schema/alter-column-builder.ts
```diff
@@ -10,49 +10,39 @@ import {
...
     )
   }
 
   setDefault(value: DefaultValueExpression): AlteredColumnBuilder {
     return new AlteredColumnBuilder(
-      AlterColumnNode.cloneWith(this.alterColumnNode, {
-        setDefault: parseDefaultValueExpression(value),
-      })
+      AlterColumnNode.create(this.column, 'setDefault', parseDefaultValueExpression(value))
```
igalklebanov, line 27:
```suggestion
      AlterColumnNode.create(this.#column, 'setDefault', parseDefaultValueExpression(value))
```

src/schema/alter-column-builder.ts
```diff
@@ -10,49 +10,39 @@ import {
...
     )
   }
 
   dropDefault(): AlteredColumnBuilder {
     return new AlteredColumnBuilder(
-      AlterColumnNode.cloneWith(this.alterColumnNode, {
-        dropDefault: true,
-      })
+      AlterColumnNode.create(this.column, 'dropDefault', true)
```
igalklebanov, line 33:
```suggestion
      AlterColumnNode.create(this.#column, 'dropDefault', true)
```

src/schema/alter-column-builder.ts
```diff
@@ -10,49 +10,39 @@ import {
...
     )
   }
 
   setNotNull(): AlteredColumnBuilder {
     return new AlteredColumnBuilder(
-      AlterColumnNode.cloneWith(this.alterColumnNode, {
-        setNotNull: true,
-      })
+      AlterColumnNode.create(this.column, 'setNotNull', true)
```
igalklebanov, line 39:
```suggestion
      AlterColumnNode.create(this.#column, 'setNotNull', true)
```

src/schema/alter-column-builder.ts
```diff
@@ -10,49 +10,39 @@ import {
...
     )
   }
 
   dropNotNull(): AlteredColumnBuilder {
     return new AlteredColumnBuilder(
-      AlterColumnNode.cloneWith(this.alterColumnNode, {
-        dropNotNull: true,
-      })
+      AlterColumnNode.create(this.column, 'dropNotNull', true)
```
igalklebanov, line 45:
```suggestion
      AlterColumnNode.create(this.#column, 'dropNotNull', true)
```

src/schema/alter-column-builder.ts
```diff
@@ -66,20 +56,30 @@ export class AlterColumnBuilder {
 }
 
 /**
- * Allows us to force consumers to do something, anything, when altering a column.
+ * Allows us to force consumers to do something, anything, when altering a column,
+ * and also disallows chaining more methods on AlterColumnBuilder, because
+ * SQL syntax for chaining ALTER COLUMN operations would be not WYSIWYG
```
igalklebanov, line 61: I think the addition is TMI.

```suggestion
 * Allows us to force consumers to do exactly one alteration to a column.
```

src/schema/alter-column-builder.ts
```diff
@@ -66,20 +56,30 @@ export class AlterColumnBuilder {
...
+ * ```
  *
  * Which would now throw a compilation error, instead of a runtime error.
  */
-export class AlteredColumnBuilder
-  extends AlterColumnBuilder
-  implements OperationNodeSource
+export class AlteredColumnBuilder implements OperationNodeSource
 {
+  protected readonly alterColumnNode: AlterColumnNode
```
igalklebanov, line 77:
```suggestion
  readonly #alterColumnNode: AlterColumnNode
```

src/schema/alter-column-builder.ts
```diff
@@ -66,20 +56,30 @@ export class AlterColumnBuilder {
...
+
+  constructor(alterColumnNode: AlterColumnNode) {
+    this.alterColumnNode = alterColumnNode
```
igalklebanov, line 80:
```suggestion
    this.#alterColumnNode = alterColumnNode
```

src/schema/alter-column-builder.ts
```diff
@@ -66,20 +56,30 @@ export class AlterColumnBuilder {
...
+  }
+
   toOperationNode(): AlterColumnNode {
     return this.alterColumnNode
```
igalklebanov, line 84:
```suggestion
    return this.#alterColumnNode
```

igalklebanov (commented):
Looks good! 💪 

Left a few minor suggestions..

soanvig:
@igalklebanov thanks for the suggestions. I've implemented them all in one commit.

igalklebanov (approved):
LGTM! 🚀

PR #471, opened 2023-05-09
Fix alterTable table name parsing

### LINKED ISSUE #470: Unable to specify PostgreSQL schema when calling schema.alterTable
When calling `db.schema.alterTable`, it is not possible to give it a table name with a postgres schema name. This is possible in `db.schema.createTable` and `db.schema.dropTable`, as they call `parseTable`, when creating the node,  which returns the table identifier separated in schema and table names 

typescript:
```
db.schema
    .alterTable("schema.table")
    .addColumn("started_at", "timestamptz")
    .execute()
```

sql:
```
alter table "schema.table" add column "started_at" timestamptz
```

sql should be:
```
alter table "schema"."table" add column "started_at" timestamptz
```

hannesj:
Fixes #470

### DIFF
--- a/test/node/src/schema.test.ts
+++ b/test/node/src/schema.test.ts
@@ -2446,6 +2446,41 @@ for (const dialect of DIALECTS) {
         })
       }
 
+      if (dialect !== 'sqlite') {
+        describe('parse schema name', () => {
+          beforeEach(cleanup)
+          afterEach(cleanup)
+
+          it('should parse the schema from table name', async () => {
+            await ctx.db.schema.createSchema('test_schema').ifNotExists().execute()
+            await ctx.db.schema.createTable('test_schema.test').addColumn('id', 'serial').execute();
+
+            const builder = ctx.db.schema
+              .alterTable('test_schema.test')
+              .addColumn('second_column', 'text')
+
+            testSql(builder, dialect, {
+              postgres: {
+                sql: `alter table "test_schema"."test" add column "second_column" text`,
+                parameters: [],
+              },
+              mysql: {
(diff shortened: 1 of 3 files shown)


koskimas:
Thank you! The fix is correct, but the test fails because there's no `pets` schema or `name` table in it (nor should there be).

hannesj:
Yeah, noticed it. Had some issues yesterday with installing better-sqlite3, but got it working now, so had no chance of running the tests then. Everything should be green now.

koskimas:
The cleanup is not called and other tests fail now. Use `before` and `after` hooks inside your new `describe` block for creating and cleaning up the schema.

test/node/src/schema.test.ts
```diff
@@ -2446,6 +2446,39 @@ for (const dialect of DIALECTS) {
...
       }
 
+      if (dialect !== 'sqlite') {
+        describe('parse schema name', () => {
+          beforeEach(cleanup)
+          afterEach(cleanup)
+
+          it('should parse the schema from table name', async () => {
+            await ctx.db.schema.createSchema('pets').ifNotExists().execute()
+            await ctx.db.schema.createTable('pets.name').addColumn('id', 'serial').execute();
```
koskimas, line 2456: The schema and table names are kinda confusing. Why is the table named `name`? And the schema `pets` since there's already a `pets` table? Change them into something else and this is good to go.

koskimas (changes requested):

test/node/src/schema.test.ts
koskimas: Oh jeesh, why have I used such a dumb schema name there 😄 Well I guess this is ok then.

PR #477, opened 2023-05-10
Allow specifying custom type in setDataType

### LINKED ISSUE #474: Updating column data type does not allow custom data types with SQL expression
This is a small issue with the ts typings.

The following works:
```
sql`create domain email as citext`.execute(db);

db.schema
    .createTable("email")
    .addColumn("email", sql`email`, (col) => col.primaryKey())
    .execute();
```

But the following is not allowed:
```
sql`create domain email as citext`.execute(db);

db.schema
    .alterTable("email")
    .alterColumn("email", (col) => col.setDataType(sql`email`))
    .execute();
```

The second argument of `addColumn` is a `DataTypeExpression` which is `ColumnDataType | Expression<any>`, while the argument for `setDataType` is just `ColumnDataType`

hannesj:
Fixes #474


koskimas:
Thank you!

PR #478, opened 2023-05-10
Workaround obscure RDS Data Api Bug

thdxr:
So I know this isn't ideal because it's making changes in kysely core for a 3rd party adapter - however it is also the simplest fix

There's some bug in the RDS api that does not like `in` statements used with char fields - it's interpreting the in values as strings instead of chars and failing

Switching it to an orWhere statement fixes the problem

Wanted to throw this up here in case you were willing to accept this change - otherwise we'll have to copy the whole PostgresIntrospector into kysely-data-api to make this one change

### DIFF
--- a/src/dialect/postgres/postgres-introspector.ts
+++ b/src/dialect/postgres/postgres-introspector.ts
@@ -70,7 +70,10 @@ export class PostgresIntrospector implements DatabaseIntrospector {
           .as('auto_incrementing'),
       ])
       // r == normal table
-      .where(qb => qb.where('c.relkind', '=', 'r').orWhere('c.relkind', '=', 'v'))
+      .where(({ cmpr, or }) => or([
+        cmpr('c.relkind', '=', 'r'),
+        cmpr('c.relkind', '=', 'v'),
+      ])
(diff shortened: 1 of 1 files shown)


src/dialect/postgres/postgres-introspector.ts
```diff
@@ -70,7 +70,7 @@ export class PostgresIntrospector implements DatabaseIntrospector {
           .as('auto_incrementing'),
       ])
       // r == normal table
-      .where('c.relkind', 'in', ['r', 'v'])
+      .where(qb => qb.where('c.relkind', '=', 'r').orWhere('c.relkind', '=', 'v'))
```
igalklebanov, line 73: `.orWhere` is deprecated. Let's use the `ExpressionBuilder` instead:

```suggestion
      .where(({ cmpr, or }) => or([
        cmpr('c.relkind', '=', 'r'),
        cmpr('c.relkind', '=', 'v'),
      ])
```

igalklebanov (changes requested):
Hey 👋 

Thank you! ❤️ 

Please check comment.

PR #483, opened 2023-05-12
feat: increase fault tolerance to deep types error

### LINKED ISSUE #481: Investigate excessively deep types
In the [excessively-deep-types doc](https://github.com/kysely-org/kysely/blob/master/site/docs/recipes/excessively-deep-types.md) you wrote "I know what you're thinking: "can't this be done automatically?" No, unfortunately it can't. There's no way to do this using current TypeScript features".

Actually thats not quite right, there is one technique that can help typescript to reset instantiation depth level. Look at the example below:

```ts
type A<T> = { item: T }

type Test<T> = A<
  A<A<A<A<A<A<A<A<A<A<A<A<A<A<A<A<A<A<A<A<A<A<A<A<T>>>>>>>>>>>>>>>>>>>>>>>>
>

type Error = Test<number>
```

Here Test is a type (25 levels deep) that causes `Type instantiation is excessively deep and possibly infinite.` error. But we can force TS to calculate generic type A "in place":

```ts
type A<T> = { item: T } extends infer X ? X : never

type Test<T> = A<
  A<A<A<A<A<A<A<A<A<A<A<A<A<A<A<A<A<A<A<A<A<A<A<A<T>>>>>>>>>>>>>>>>>>>>>>>>
>

type NoError = Test<number>
```

Now everything is ok. The key idea here is to add `extends infer X ? X : never` to the innermost generic type to force TS to unwrap it.

So maybe it will help you improve inference for complex queries.

schusovskoy:
Closes #481

Background: TS treats differently simple type alias with generic parameters and type alias which uses conditional type inside. In particular using a conditional type instead of a simple generic doesn't create a new level of nesting. This feature allows us to create deep types without encountering "Type instantiation is excessively deep and possibly infinite" error.

In this PR I added `DrainOuterGeneric` type which transforms original type to conditional, and therefore removes outer generic type wrapper and reduces instantiation depth. Next I replaced every Record with ShallowRecord (the same type but without wrapper). But the key change was to wrap in `DrainOuterGeneric` every "simple" generic type alias located in the return type of query builders methods.

With this relatively small changes we're now able to safely remove every $assertType calls in `test/typings/test-d/with.test-d.ts`.

Also as a side effect it improves performance of type checking by 13%.

Running `npm run test:typings` before changes.
Screenshot 2023-05-12 at 00 29 34

Running `npm run test:typings` after changes.
Screenshot 2023-05-12 at 00 32 53

### DIFF
--- a/test/typings/test-d/with.test-d.ts
+++ b/test/typings/test-d/with.test-d.ts
@@ -80,36 +80,11 @@ async function testWith(db: Kysely<Database>) {
 
 async function testManyWith(db: Kysely<Database>) {
   const res = await db
-    .with('w1', (eb) =>
-      eb
-        .selectFrom('person')
-        .select('first_name as fn1')
-        .$assertType<{ fn1: string }>()
-    )
-    .with('w2', (eb) =>
-      eb
-        .selectFrom('person')
-        .select('first_name as fn2')
-        .$assertType<{ fn2: string }>()
-    )
-    .with('w3', (eb) =>
-      eb
-        .selectFrom('person')
-        .select('first_name as fn3')
-        .$assertType<{ fn3: string }>()
-    )
-    .with('w4', (eb) =>
-      eb
-        .selectFrom('person')
-        .select('first_name as fn4')
-        .$assertType<{ fn4: string }>()
-    )
-    .with('w5', (eb) =>
-      eb
-        .selectFrom('person')
-        .select('first_name as fn5')
-        .$assertType<{ fn5: string }>()
-    )
+    .with('w1', (eb) => eb.selectFrom('person').select('first_name as fn1'))
+    .with('w2', (eb) => eb.selectFrom('person').select('first_name as fn2'))
+    .with('w3', (eb) => eb.selectFrom('person').select('first_name as fn3'))
+    .with('w4', (eb) => eb.selectFrom('person').select('first_name as fn4'))
+    .with('w5', (eb) => eb.selectFrom('person').select('first_name as fn5'))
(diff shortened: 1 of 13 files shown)


koskimas:
This seems amazing! I didn't believe something like this was possible. The only thing I'm wondering is are we trusting an obscure typescript implementation detail here that could change in the future?

Also the tests still seem to fail. If you're able to make them work, this will definitely get merged.

test/typings/test-d/with.test-d.ts
```diff
@@ -132,6 +107,10 @@ async function testManyWith(db: Kysely<Database>) {
     .selectAll()
     .executeTakeFirstOrThrow()
 
+  type IsAny<T> = 0 extends 1 & T ? true : false
```
koskimas, line 110: No need for this. `expectType` fails if the type is too wide.

koskimas:
The test error is only about not having a `.js` suffix in some imports.

igalklebanov:
Great stuff @schusovskoy! 💪 

We need to make sure this doesn't break autocompletion - probably doesn't.

schusovskoy:
> The only thing I'm wondering is are we trusting an obscure typescript implementation detail here that could change in the future?

I think that it's like conceptual difference between type alias and conditional type, where the last is some kind of "computation" in type system and therefore doesn't require to create intermediate type bindings.

Anyway they have an [issue](https://github.com/microsoft/TypeScript/issues/34933) which promise to solve the root cause of the problem and some day we can safely remove this "hack".

PR #504, opened 2023-05-19
Update README.md

leleofg:
add example enums in documentation


koskimas:
Thanks, but we're working on the new site that will replace the readme soon.

PR #512, opened 2023-05-22
feat(sql.raw): Allow parameters in raw sql

### LINKED ISSUE #511: Raw query with parameters
I'm in the process of migrating an application to use Kysely, and in a few areas, I already have hand-written SQL statements with parameters like:

```sql
select * from "person" where first_name between $1 and $2;
```

> NOTE: This is a simple example, in reality, the queries are quite complex.

Is there a way to run a raw query like this with Kysely? 

FWIW, the API that I *wish* existed would be `sql.raw(query, ...paramerters)`;

nicksrandall:
Fixes #511

### DIFF
--- a/src/query-compiler/compiled-query.ts
+++ b/src/query-compiler/compiled-query.ts
@@ -9,11 +9,11 @@ export interface CompiledQuery<O = unknown> {
 }
 
 export const CompiledQuery = freeze({
-  raw(sql: string, ...parameters: unknown[]): CompiledQuery {
+  raw(sql: string, parameters: unknown[] = []): CompiledQuery {
     return freeze({
       sql,
       query: RawNode.createWithSql(sql),
-      parameters: freeze(parameters ?? []),
+      parameters: freeze(parameters),
     })
   },
 })
--- a/src/query-compiler/compiled-query.ts
+++ b/src/query-compiler/compiled-query.ts
@@ -9,11 +9,11 @@ export interface CompiledQuery<O = unknown> {
 }
 
 export const CompiledQuery = freeze({
-  raw(sql: string): CompiledQuery {
+  raw(sql: string, ...parameters: unknown[]): CompiledQuery {
     return freeze({
       sql,
       query: RawNode.createWithSql(sql),
-      parameters: freeze([]),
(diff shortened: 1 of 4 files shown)


test/node/src/raw-sql.test.ts
```diff
@@ -268,6 +268,14 @@ for (const dialect of DIALECTS) {
       })
     }
 
+    it('sql.raw should accept parameters', async () => {
+      if (dialect === 'postgres') {
+          const query = sql.raw('select * from "person" where "public"."person"."first_name" between $1 and $2', 'A', 'B').compile(ctx.db);
+          expect(query.sql).to.equal('select * from "person" where "public"."person"."first_name" between $1 and $2');
+          expect(query.parameters).to.deep.equal(['A', 'B']);
+      }
+    });
```
nicksrandall, line 277: I wasn't sure how to test this. I'm open to feedback on a better way to do this.

src/query-compiler/compiled-query.ts
```diff
@@ -9,11 +9,11 @@ export interface CompiledQuery<O = unknown> {
 }
 
 export const CompiledQuery = freeze({
-  raw(sql: string): CompiledQuery {
+  raw(sql: string, ...parameters: unknown[]): CompiledQuery {
```
igalklebanov, line 12: Anyone using this directly with parameters will probably have them in an array. Let's not force consumer to spread.

```suggestion
  raw(sql: string, parameters: unknown[] = []): CompiledQuery {
```

src/query-compiler/compiled-query.ts
```diff
@@ -9,11 +9,11 @@ export interface CompiledQuery<O = unknown> {
...
     return freeze({
       sql,
       query: RawNode.createWithSql(sql),
-      parameters: freeze([]),
+      parameters: freeze(parameters ?? []),
```
igalklebanov, line 16:
```suggestion
      parameters: freeze(parameters),
```

igalklebanov (commented):
I think that's a good idea.

Left some suggestions..

nicksrandall:
@igalklebanov I *think* this is ready. I applied your suggestions.

PR #521, opened 2023-05-27
docs: getting started.

igalklebanov:
This PR goes over our new docs' "getting started" page and makes it friendlier.


site/docs/getting-started/Querying.tsx
```diff
@@ -0,0 +1,115 @@
...
+export type Person = Selectable<PersonTable>
+
+export async function findPersonById(id: Person['id']) {
+  return await db.selectFrom('person')
+    .where('id', '=', id)
+    .selectAll()
+    .executeTakeFirst()
+}
+
+export async function findPeople(criteria: Partial<Person>) {
```
koskimas, line 66: Looking at this example, maybe I need to reconsider my opinion about the

```ts
where({ first_name: 'Sami' })
```

overload 🤔

igalklebanov: I had an `Object.entries` solution at first, but thought it might be too complex for some our readers. (not taking any assumptions on their JavaScript skills, only on them knowing SQL).

igalklebanov:
> This is looking great! Awesome job 👍
> 
> One thing I'd add: We should mention the `Selectable`, `Insertable` etc. types and that you shouldn't use the table interfaces as row types directly. That's a very intuitive thing to do and we've had a bunch of issues along the lines of "Generated<string> is not assignable to string".

Great idea! I'll move them to the `Types` section and add explanations.

site/docs/getting-started/Dialects.tsx
```diff
@@ -0,0 +1,147 @@
...
+interface UnsupportedDriverProps {
+  dialect: string
+  driverNPMPackage: string
+  packageManager: PackageManager
+}
+
+function UnsupportedDriver(props: UnsupportedDriverProps) {
+  const { dialect, packageManager } = props
+
+  const titleCasedPackageManager = titlecase(packageManager)
```
igalklebanov, line 137: not to self, use the pretty package managers object instead of this.

koskimas:
Pure greatness 💯

PR #525, opened 2023-06-02
test deno with real drivers.

igalklebanov:
This PR proves Kysely can run in Deno with both PostgresDialect & MysqlDialect passing tests. Couldn't get `better-sqlite3` to work.

Had to bump workflows' deno version to v1.34 to support npm specifiers (npm specifiers are officially supported since v1.28, so I prefer we recommend using these).

post-merge actions:

- make required PR checks align with new deno version.


koskimas (approved):
The file name `cdn.test.ts` doesn't make much sense anymore (or the mentions of CDN inside it). Maybe it could be `npm.test.ts` or something?

igalklebanov:
> The file name `cdn.test.ts` doesn't make much sense anymore (or the mentions of CDN inside it). Maybe it could be `npm.test.ts` or something?

Good idea!

PR #531, opened 2023-06-11
docs: add databases section @ site home page.

igalklebanov:
image

Need to host the logos instead of using external URLs, other than that, pretty much done.

Logos order:
1. core dialects.
2. by revenue / funds raised - as a way to guess market share / marketing power.
3. by github stars.


fhur:
Very cool. Some thoughts:

1. Would be useful if the the blocks were clickable (e.g. click on planetscale sends you to the planetscale docs, in another window, so we don't lose the session)
2. With regards to copy, what do you think of "official dialect" as opposed to "core dialect"?

igalklebanov:
@fhur 

Good ideas! Also, we should probably optimize loading of these images, I lazily copied and used external urls for this PoC.

I think "official" might confuse people. e.g. thinking that we have an official Supabase dialect.

Sent you an invite.

igalklebanov:
Also, "official" sheds a negative light on "community".

PR #540, opened 2023-06-20
Consistently used sqlite_master for compatibility with older SQLite versions

mphill:
There is a mismatch in the schema name being used in the inspector, in one query sqlite_schema is used, then directly after sqlite_master is used.  They renamed sqlite_master to sqlite_schema, but both exist at this point and will for the foreseeable future.  

In 3.33.0 created an alias for sqlite_master, sqlite_schema.

See:

https://sqlite.org/forum/forumpost/d90adfbb0a

Thanks!

### DIFF
--- a/src/dialect/sqlite/sqlite-introspector.ts
+++ b/src/dialect/sqlite/sqlite-introspector.ts
@@ -28,7 +28,7 @@ export class SqliteIntrospector implements DatabaseIntrospector {
     options: DatabaseMetadataOptions = { withInternalKyselyTables: false }
   ): Promise<TableMetadata[]> {
     let query = this.#db
-      .selectFrom('sqlite_schema')
+      .selectFrom('sqlite_master')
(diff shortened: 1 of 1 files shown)


koskimas:
Thank you!

igalklebanov (commented):
Let's use `sqlite_schema` instead of `sqlite_master`, as "master" might be offensive and its also the legacy name now.

mphill:
https://developer.android.com/reference/android/database/sqlite/package-summary

I’m trying to get kysely working on android via react native/expo. As you see kysely will currently only support the latest android version due to the sqlite_schema reference. Effectively excluding billions of devices. 

Using sqlite_master allows kysely to work on all Android versions, all versions of SQLite everywhere. 


Android aside, kysely will not work on older SQLite versions release priory to 8/2020 as a result of using sqlite_schema.

igalklebanov (approved):

PR #548, opened 2023-06-24
docs(getting started): add clarification to Types section

oscartorres-10:
Clarifying that automatic generation of the Database interface could lead to a different file of the one shown currently in the documentation, by warning readers of this could help saving some time if they find mismatches, this happened to me while using `prisma-kysely` package.

### DIFF
--- a/site/docs/getting-started/_types.mdx
+++ b/site/docs/getting-started/_types.mdx
@@ -58,6 +58,6 @@ export type EditedPet = Updateable<PetTable>
 
 For production apps, it is recommended to automatically generate your <code>Database</code>
 interface by introspecting your production database or Prisma schemas. Generated types
-might differ in naming convention, internal order, etc. Find out more at ["Generating types"(https://kysely.dev/docs/generating-types).
+might differ in naming convention, internal order, etc. Find out more at ["Generating types"](https://kysely.dev/docs/generating-types).
 
 :::
--- a/site/docs/getting-started/_types.mdx
+++ b/site/docs/getting-started/_types.mdx
@@ -57,8 +57,7 @@ export type EditedPet = Updateable<PetTable>
 :::tip Codegen
 
 For production apps, it is recommended to automatically generate your <code>Database</code>
-interface by introspecting your production database or Prisma schemas. Find out more at
-["Generating types"](https://kysely.dev/docs/generating-types).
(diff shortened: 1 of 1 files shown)


site/docs/getting-started/_types.mdx
```diff
@@ -59,5 +59,6 @@ export type EditedPet = Updateable<PetTable>
 For production apps, it is recommended to automatically generate your <code>Database</code>
 interface by introspecting your production database or Prisma schemas. Find out more at
 ["Generating types"](https://kysely.dev/docs/generating-types).
+Consider that automatic code generation could lead to different naming and structure of this file, for example the Database interface could be named DB and placed at the bottom.
```
igalklebanov, line 62: Generated types are hinted at here, but they're not really part of the demonstration.
For that we've got the "Generating types" page, and in all honesty, each individual community library's documentation.

I think this is a good idea! But the `DB` part or "placed at the bottom" are too specific.

```suggestion
interface by introspecting your production database or Prisma schemas. Generated types
might differ in naming convention, internal order, etc. Find out more at ["Generating types"(https://kysely.dev/docs/generating-types).
```

Is this clear enough without coupling ourselves to community libraries code?

oscartorres-10: Great points on decoupling from other libraries. What you wrote is good enough for me! Although I would put it all in the same line, I'm seeing a line break after "Generated types"

oscartorres-10: > More in-depth text should go to "Generating types" page

Would be useful if I try to make an improvement to that page by adding maybe a note with more specifics?

igalklebanov: > Great points on decoupling from other libraries. What you wrote is good enough for me! Although I would put it all in the same line, I'm seeing a line break after "Generated types"

I generally try and make markdown lines not longer than 80 characters for readability in code editors. This doesn't affect how the text is viewed in the site or in markdown preview.

oscartorres-10: Got it, didn't knew about that, thanks

igalklebanov: > > More in-depth text should go to "Generating types" page
> 
> Would be useful if I try to make an improvement to that page by adding maybe a note with more specifics?

Sure, give it a pass and see if it can be improved.

koskimas (approved):

PR #549, opened 2023-06-25
Update repository url in package.json

gioragutt:


igalklebanov (approved):
Good catch! 💪 

Thanks!

PR #551, opened 2023-06-27
docs(getting-started): declare 'query' using 'let' instead of 'const' to allow re-assignment

robertgr991:


igalklebanov (approved):
Good catch! 💪

PR #553, opened 2023-06-27
docs: asset perf tweaks

fhur:
# What changed

1. Minor perf tweaks to improve lighthouse score
2. Self host assets to prevent links randomly going dead (as a result fixed the two broken assets)
3. Tweaked descriptions a bit to prevent images being squished
4. Using lighter GIF (Essentially dropped every a fraction of the frames). Still huge 5MBs, but at least ~50% lighter.

## Old lighthouse score
Screenshot 2023-06-27 at 22 25 37

## New lighthouse score
Screenshot 2023-06-27 at 22 25 01


igalklebanov:
Love the improvements!

Wonder if the big 5MB could be streamed as an autoplaying looping video without controls.

I think the video might be too sped up. But I like the idea, the message comes across much quicker.

fhur:
There's other ways for sure to improve the gif, but this was the lowest effort one 😅

koskimas:
Good changes, except that the GIF is now much much faster and really hard to follow. It needs to go back to the old speed.

koskimas (changes requested):
Good changes, except that the GIF is now much much faster and really hard to follow. It needs to go back to the old speed. With this speed, you don't even see the autocompletions. We might see them since we know what to look for, but people watching that for the first time won't notice the cool things.

I think the GIF was the reason for Kysely's quick initial growth. I spent a lot of time optimizing it. Let's not change its speed. Of course making it smaller is a good idea, but other than that, let's keep it as is.

koskimas:
Based on some quick tests, converting the GIF to webm or mp4 with high enough quality, results in a ~2MB file. 900KB file still look decent, but the text is a bit soft already.

fhur:
Alright, should be ready for another review :)

fhur:
Screenshot 2023-06-28 at 14 56 18

And the new lighthouse score!

koskimas (approved):

PR #554, opened 2023-06-28
docs: add auto generated warning

fhur:
This MR simply adds the classical "this is auto-generated" warning, to prevent auto-generated files to accidentally be modified manually.


koskimas:
This PR seems to be dead. Closing.

PR #560, opened 2023-07-03
Update outdated alterColumn function call example

BSteffaniak:
This function call syntax was changed in the breaking change noted in https://github.com/kysely-org/kysely/releases/tag/0.23.0

Thanks!


igalklebanov (approved):
LGTM! 🚀 

Thanks!

PR #562, opened 2023-07-04
docs: fix introspecting relation metadata example

virtuallyunknown:
This PR fixes/improves the following:

The example for Introspecting Relation Metadata has two minor mistakes.

- The pg library does not provide a named export `Pool`, use default import and destructuring instead
- `db().introspection.getTables()` is not callable, needs to be `db.introspection.getTables()`
- Call the `logDatabaseSchema` function at the end of the example to see the results

### DIFF
--- a/site/docs/recipes/introspecting-relation-metadata.md
+++ b/site/docs/recipes/introspecting-relation-metadata.md
@@ -6,7 +6,8 @@ The example below uses a PostgreSQL connection to print information about all ta
 
 ```ts
 import { Kysely, PostgresDialect } from 'kysely'
-import { Pool } from 'pg'
+import pg from 'pg'
+const { Pool } = pg
 
 async function logDatabaseSchema() {
   const db = new Kysely({
@@ -17,11 +18,13 @@ async function logDatabaseSchema() {
     }),
   })
 
-  const tables = await db().introspection.getTables()
+  const tables = await db.introspection.getTables()
   //        ^?  TableMetadata[]
 
   console.log({ tables })
 }
+
+logDatabaseSchema()
 ```
 
 For more information check the docs for details on the interfaces [DatabaseIntrospector](https://kysely-org.github.io/kysely/interfaces/DatabaseIntrospector.html) and [TableMetadata](https://kysely-org.github.io/kysely/interfaces/TableMetadata.html).


josueayala27 (approved):

site/docs/recipes/introspecting-relation-metadata.md
```diff
@@ -6,7 +6,8 @@ The example below uses a PostgreSQL connection to print information about all ta
 
 ```ts
 import { Kysely, PostgresDialect } from 'kysely'
-import { Pool } from 'pg'
+import pg from 'pg'
+const { Pool } = pg
```
igalklebanov, line 10: I'm not sure about this one. Are you using Deno?

https://github.com/kysely-org/kysely/blob/master/test/node/src/test-setup.ts#L5

virtuallyunknown: > I'm not sure about this one. Are you using Deno?
> 
> Our Node.js tests use `import { Pool } from 'pg'` https://github.com/kysely-org/kysely/blob/master/test/node/src/test-setup.ts#L5

Nope, I am using Node 20. I think your tests are working because the `tsc` compiler is converting to CommonJS, and the test file ends up looking like this:

```js
// kysely/test/node/dist/select.test.js
const pg_1 = require("pg");

// later used as:
new pg_1.Pool(...)
```

However, check out this bare minimum example:

https://stackblitz.com/edit/stackblitz-starters-b7zu7w

```
import { Pool } from 'pg';
         ^^^^
SyntaxError: Named export 'Pool' not found. The requested module 'pg' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'pg';
const { Pool } = pkg;
```

This can be an issue with TypeScript projects as well. For example, I am using the following esbuild build configuration, and the named export won't work:

```js
await build({
    platform: 'node',
    target: 'node16',
    format: 'esm',
    bundle: false,
});
```

In any case, I think using default instead of named export is just the safer option. Oddly enough, the official pg docs example is using named export, but obviously that doesn't always work as demonstrated above.

https://github.com/brianc/node-postgres/issues/2819
https://stackoverflow.com/a/70306750/3258251

PR #574, opened 2023-07-09
add cloudflare workers test.

igalklebanov:
Hey 👋 

This PR proves Kysely can run in Cloudflare Workers, and PostgresDialect can execute queries.

> Though Cloudflare Workers behave similarly to JavaScript in the browser or in Node.js, there are a few differences in how you have to think about your code. Under the hood, the Workers runtime uses the V8 engine — the same engine used by Chromium and Node.js. The Workers runtime also implements many of the standard [APIs](https://developers.cloudflare.com/workers/runtime-apis/) available in most modern browsers.
https://developers.cloudflare.com/workers/learning/how-workers-works/


koskimas (approved):
LGTM!

PR #578, opened 2023-07-11
Support materialized CTEs using CTE builder

### LINKED ISSUE #246: Support [cte] as materialized (...)
Since Postgres 12, there's now the ability to tell Postgres to MATERIALIZE a CTE, this can have significant performance benefits downstream. Something as simple as this

```
db.with("cte", (qb) => qb.selectFrom("user").select("name")).materialized()
```

Would generate
```
WITH cte AS MATERIALIZED (select name from user)
```

Likewise `.materialized(false)` could generate ` AS NOT MATERIALIZED `

koskimas:
The first argument of `with` can now be a callback that's given a CTE builder. The CTE builder can be used to create the common table expression statement.

```ts
db.with(
    (cte) => cte('jennifers').materialized(),
    (db) =>
      db
        .selectFrom('person')
        .where('first_name', '=', 'Jennifer')
        .select(['first_name', 'last_name as ln', 'gender'])
  )
  .selectFrom('jennifers')
  .select(['first_name', 'ln'])
  .execute()
```

The syntax feels natural and this way `as materialized` is "in the correct place" between the name and the query. But we're introducing a new way to do things here and I'm not sure if this a good precedent for future development, or just something completely disjoint from all other builders.

Putting the CTE builder last, after the query, also feels bad since it's in the wrong place. But that would be more in line with the other builders. 🤷

Closes #246

### DIFF
--- a/src/query-compiler/default-query-compiler.ts
+++ b/src/query-compiler/default-query-compiler.ts
@@ -938,6 +938,15 @@ export class DefaultQueryCompiler
   ): void {
     this.visitNode(node.name)
     this.append(' as ')
+
+    if (isBoolean(node.materialized)) {
+      if (node.materialized) {
+        this.append('materialized ')
+      } else {
+        this.append('not materialized ')
+      }
+    }
+
     this.visitNode(node.expression)
   }
 
--- a/src/query-compiler/default-query-compiler.ts
+++ b/src/query-compiler/default-query-compiler.ts
@@ -940,11 +940,11 @@ export class DefaultQueryCompiler
     this.append(' as ')
 
     if (isBoolean(node.materialized)) {
-      if (node.materialized) {
-        this.append('materialized ')
-      } else {
-        this.append('not materialized ')
+      if (!node.materialized) {
+        this.append('not ')
       }
+      
+      this.append('materialized ')
     }
 
     this.visitNode(node.expression)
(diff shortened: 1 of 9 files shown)


koskimas:
@igalklebanov I'll rename `asMaterialized` to `materialized`. `as` is not a part of `materialized` keyword.

src/query-compiler/default-query-compiler.ts
```diff
@@ -938,6 +938,15 @@ export class DefaultQueryCompiler
   ): void {
     this.visitNode(node.name)
     this.append(' as ')
+
+    if (isBoolean(node.materialized)) {
+      if (node.materialized) {
+        this.append('materialized ')
+      } else {
+        this.append('not materialized ')
+      }
```
igalklebanov, line 947: alternatively could...

```suggestion
      if (!node.materialized) {
        this.append('not ')
      }
      
      this.append('materialized ')
```

I try to avoid `else` as much as possible.

src/query-creator.ts
igalklebanov, line 1: Could we split into 2 overloads each, to simplify the first argument (name and type)?

igalklebanov (approved):
LGTM! 🚀 

left some nitpicky minor comments/suggestions..

src/query-creator.ts
koskimas: I tried that. It triples the amount of code. Not a good tradeoff.

PR #588, opened 2023-07-14
feat: kysely/helpers/sqlite

jlarmstrongiv:
I have tested the queries in the comments—those SQLite queries work as intended.

Unfortunately, I could not use the PostgreSQL helpers due to `Parse error: no such column: agg`.  It seems that SQLite still expects column names.

I have also modified the kysely functions to match the SQLite queries.

The schema used in the examples is:
```sql
CREATE TABLE IF NOT EXISTS "person" ("id" integer primary key, "first_name" text not null, "last_name" text, "gender" text not null, "created_at" text default CURRENT_TIMESTAMP not null);
CREATE TABLE IF NOT EXISTS "pet" ("id" integer primary key, "name" text not null unique, "owner_id" integer not null references "person" ("id") on delete cascade, "species" text not null, is_favorite boolean);
CREATE INDEX "pet_owner_id_index" on "pet" ("owner_id");
```

with inserts like:
```
insert into person (id, first_name, last_name, gender, created_at) values (2, 'joe', 'joel', 'j', '2');
insert into person (id, first_name, last_name, gender, created_at) values (3, 'sue', 'suel', 's', '3');
insert into pet (id, name, owner_id, species, is_favorite) values (1, 'jill', 2, 'j', true);
insert into pet (id, name, owner_id, species, is_favorite) values (2, 'mark', 2, 'm', true);
insert into pet (id, name, owner_id, species) values (3, 'cole', 3, 'c');
```

EDIT: will add tests to the suite https://github.com/kysely-org/kysely/blob/master/test/node/src/json.test.ts thank you @koskimas

### DIFF
--- a/site/docs/recipes/relations.md
+++ b/site/docs/recipes/relations.md
@@ -13,7 +13,13 @@ In this recipe we show one way to do that when using the built-in PostgreSQL, My
 
 ## The `json` data type and functions
 
-PostgreSQL, MySQL, and SQLite have rich JSON support through their `json` data types and functions. `pg` and `mysql2`, the node drivers, automatically parse returned `json` columns as json objects. With the combination of these two things, we can write some super efficient queries with nested relations.
+PostgreSQL and MySQL have rich JSON support through their `json` data types and functions. `pg` and `mysql2`, the node drivers, automatically parse returned `json` columns as json objects. With the combination of these two things, we can write some super efficient queries with nested relations.
+
+With the `ParseJSONResultsPlugin`, SQLite can also automatically parse results:
+
+```ts
+db = db.withPlugin(new ParseJSONResultsPlugin())
+```
 
 Let's start with some raw postgres SQL, and then see how we can write the query using Kysely in a nice type-safe way.
 
--- a/site/docs/recipes/relations.md
+++ b/site/docs/recipes/relations.md
@@ -7,13 +7,13 @@ Kysely IS a query builder. Kysely DOES build the SQL you tell it to, nothing mor
 
 Phew, glad we got that out the way..
 
-All that was said above doesn't mean there's no way to nest related rows in your queries. 
-You just have to do it with the tools SQL and the underlying dialect (e.g. PostgreSQL or MySQL) provide.
-In this recipe we show one way to do that when using the built-in PostgreSQL and MySQL dialects.
+All that was said above doesn't mean there's no way to nest related rows in your queries.
+You just have to do it with the tools SQL and the underlying dialect (e.g. PostgreSQL, MySQL, or SQLite) provide.
+In this recipe we show one way to do that when using the built-in PostgreSQL, MySQL, and SQLite dialects.
 
 ## The `json` data type and functions
 
-Both PostgreSQL and MySQL have rich JSON support through their `json` data types and functions. `pg` and `mysql2`, the node drivers, automatically parse returned `json` columns as json objects. With the combination of these two things, we can write some super efficient queries with nested relations.
+PostgreSQL, MySQL, and SQLite have rich JSON support through their `json` data types and functions. `pg` and `mysql2`, the node drivers, automatically parse returned `json` columns as json objects. With the combination of these two things, we can write some super efficient queries with nested relations.
 
 Let's start with some raw postgres SQL, and then see how we can write the query using Kysely in a nice type-safe way.
 
@@ -82,10 +82,14 @@ These helpers are included in Kysely and you can import them from the `helpers` 
 import { jsonArrayFrom, jsonObjectFrom } from 'kysely/helpers/postgres'
 ```
 
-MySQL versions of the helpers are slightly different but you can use them the same way. You can import them like this:
+MySQL and SQLite versions of the helpers are slightly different but you can use them the same way. You can import them like this:
 
 ```ts
 import { jsonArrayFrom, jsonObjectFrom } from 'kysely/helpers/mysql'
+```
+
+```ts
+import { jsonArrayFrom, jsonObjectFrom } from 'kysely/helpers/sqlite'
 ```
 
 With these helpers, our example query already becomes a little more bearable to look at:
--- a/src/helpers/sqlite.ts
+++ b/src/helpers/sqlite.ts
@@ -164,7 +164,8 @@ export function jsonBuildObject<O extends Record<string, Expression<unknown>>>(
   }>
 > {
   return sql`json_object(${sql.join(
-    Object.keys(obj).flatMap((k) => [sql.lit(k), obj[k]])
+    Object.keys(obj).flatMap((k) => [sql.lit(k), obj[k]]),
+    sql` || `
   )})`
 }
 
--- a/src/helpers/sqlite.ts
+++ b/src/helpers/sqlite.ts
@@ -57,7 +57,7 @@ import { Simplify } from '../util/type-utils.js'
 export function jsonArrayFrom<O>(
   expr: SelectQueryBuilder<any, any, O>
 ): RawBuilder<Simplify<O>[]> {
-  return sql`(select coalesce(json_arrayagg(json_object(${sql.join(
+  return sql`(select coalesce(json_group_array(json_object(${sql.join(
     getJsonObjectArgs(expr.toOperationNode(), 'agg')
   )})), '[]') from ${expr} as agg)`
 }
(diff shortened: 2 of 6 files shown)


koskimas:
Nice job, Thank you! Both postgres and mysql helpers are tested [here](https://github.com/kysely-org/kysely/blob/master/test/node/src/json.test.ts). You can add sqlite tests to the same suite.

koskimas:
You also need to add new export from `kysely/helpers/sqlite` to package.json. See the section with 

```ts
    "./helpers/mysql": {
      "import": "./dist/esm/helpers/mysql.js",
      "require": "./dist/cjs/helpers/mysql.js",
      "default": "./dist/cjs/helpers/mysql.js"
    }
```

igalklebanov (commented):
Thank you! 💪 

Let's not forget to mention how these can be imported in the [relations recipe](https://kysely.dev/docs/recipes/relations).

jlarmstrongiv:
@igalklebanov how do I concat in sqlite with kysely?  I saw in another comment in another test that:

```js
return sql`json_object(${sql.join(
    Object.keys(obj).flatMap((k) => [sql.lit(k), obj[k]])
  )})`
```

should return the concatenation operator `||` in postgres (and I assume SQLite), but that doesn’t seem to be the case.  I am receiving the error:

```
SqliteError: no such function: concat
```

site/docs/recipes/relations.md
```diff
@@ -7,13 +7,13 @@ Kysely IS a query builder. Kysely DOES build the SQL you tell it to, nothing mor
...
-You just have to do it with the tools SQL and the underlying dialect (e.g. PostgreSQL or MySQL) provide.
-In this recipe we show one way to do that when using the built-in PostgreSQL and MySQL dialects.
+All that was said above doesn't mean there's no way to nest related rows in your queries.
+You just have to do it with the tools SQL and the underlying dialect (e.g. PostgreSQL, MySQL, or SQLite) provide.
+In this recipe we show one way to do that when using the built-in PostgreSQL, MySQL, and SQLite dialects.
 
 ## The `json` data type and functions
 
-Both PostgreSQL and MySQL have rich JSON support through their `json` data types and functions. `pg` and `mysql2`, the node drivers, automatically parse returned `json` columns as json objects. With the combination of these two things, we can write some super efficient queries with nested relations.
+PostgreSQL, MySQL, and SQLite have rich JSON support through their `json` data types and functions. `pg` and `mysql2`, the node drivers, automatically parse returned `json` columns as json objects. With the combination of these two things, we can write some super efficient queries with nested relations.
```
igalklebanov, line 16: Let's suggest that in SQLite's case, you'd need [ParseJSONResultsPlugin](https://github.com/kysely-org/kysely/blob/master/src/plugin/parse-json-results/parse-json-results-plugin.ts) for automatic parsing since the driver doesn't do it?

jlarmstrongiv: @igalklebanov mentioned the `ParseJSONResultsPlugin` in the docs with example usage 👍

koskimas:
> @igalklebanov how do I concat in sqlite with kysely? I saw in another comment in another test that:
> 
> ```js
> return sql`json_object(${sql.join(
>     Object.keys(obj).flatMap((k) => [sql.lit(k), obj[k]])
>   )})`
> ```
> 
> should return the concatenation operator `||` in postgres (and I assume SQLite), but that doesn’t seem to be the case. I am receiving the error:
> 
> ```
> SqliteError: no such function: concat
> ```

The code you provided has nothing to do with concatenation. That SQL snippet should produce

```ts
json_object('foo', some_expression, 'bar', some_other_expression, ...)
```

There's no concatenation anywhere, unless one of the expressions has concatenation in them.

I'm guessing the comment you referred to jus the `||` doesn't exist on mysql. But it does on postgres and sqlite.

rprwhite:
> I'm guessing the comment you referred to just meant the `||` operator doesn't exist on mysql. But it does on postgres and sqlite. Change the test so that `||` is used on sqlite instead of the `concat` function.

Replacing the `concat` function 
https://github.com/kysely-org/kysely/blob/5d53ac45316497d29b7c3a63b5b30c7c158f93b8/test/node/src/json.test.ts#L175
with 
```typescript
  full: sql<string>`first_name || ' ' || last_name`
```
Throws a type error at build 
```error
TypeError: SQLite3 can only bind numbers, strings, bigints, buffers, and null
```
I assume, because of this https://github.com/WiseLibs/better-sqlite3/issues/907#issuecomment-1324547474

It does work in postgres and mysql (as long as PIPES_AS_CONCAT is set).
What's the best way to approach this for sqlite?

koskimas:
That error makes no sense. That sql isn't binding anything. it's a raw SQL snippet. My guess is that your implementation of `jsonBuildObject` tries to send the whole `sql` instance as a parameter for some reason.

koskimas:
The next problem is that sqlite doesn't parse JSON (since there's no JSON datatype). You need to add this

```ts
      if (dialect === 'sqlite') {
        db = db.withPlugin(new ParseJSONResultsPlugin())
      }
```

in the `before` block. But it seems that `ParseJSONResultsPlugin` only parses the top level objects. In the nested relations, the data is doubly, triply etc. encoded json and the plugin would need to do the parsing recursively. @igalklebanov

igalklebanov:
@koskimas yeah, this was good enough for the JSON traversal tests. Maube add an optional flag to recurse?

koskimas:
I can fix the plugin. I don't think there's any need for a flag.

koskimas:
The plugin is now fixed.

src/helpers/sqlite.ts
```diff
@@ -0,0 +1,191 @@
...
+import { SelectQueryNode } from '../operation-node/select-query-node.js'
+import { SelectQueryBuilder } from '../query-builder/select-query-builder.js'
+import { RawBuilder } from '../raw-builder/raw-builder.js'
+import { sql } from '../raw-builder/sql.js'
+import { Simplify } from '../util/type-utils.js'
+
+/**
+ * A SQLite helper for aggregating a subquery into a JSON array.
+ *
+ * NOTE: This helper is only guaranteed to fully work with the built-in `SQLiteDialect`.
```
koskimas, line 15: SqliteDialect instead of SQLiteDialect

src/helpers/sqlite.ts
```diff
@@ -0,0 +1,191 @@
...
+    getJsonObjectArgs(expr.toOperationNode(), 'agg')
+  )})), '[]') from ${expr} as agg)`
+}
+
+/**
+ * A SQLite helper for turning a subquery into a JSON object.
+ *
+ * The subquery must only return one row.
+ *
+ * NOTE: This helper is only guaranteed to fully work with the built-in `SQLiteDialect`.
```
koskimas, line 70: SqliteDialect instead of SQLiteDialect

src/helpers/sqlite.ts
```diff
@@ -0,0 +1,191 @@
...
+): RawBuilder<Simplify<O> | null> {
+  return sql`(select json_object(${sql.join(
+    getJsonObjectArgs(expr.toOperationNode(), 'obj')
+  )}) from ${expr} as obj)`
+}
+
+/**
+ * The SQLite `json_object` function.
+ *
+ * NOTE: This helper is only guaranteed to fully work with the built-in `SQLiteDialect`.
```
koskimas, line 123: SqliteDialect instead of SQLiteDialect

jlarmstrongiv:
Good catch, fixed that typo @koskimas 👍

src/helpers/sqlite.ts
```diff
@@ -0,0 +1,191 @@
...
+import { RawBuilder } from '../raw-builder/raw-builder.js'
+import { sql } from '../raw-builder/sql.js'
+import { Simplify } from '../util/type-utils.js'
+
+/**
+ * A SQLite helper for aggregating a subquery into a JSON array.
+ *
+ * NOTE: This helper is only guaranteed to fully work with the built-in `SqliteDialect`.
+ * While the produced SQL is compatible with all SQLite databases, some 3rd party dialects
+ * may not parse the nested results into arrays.
```
koskimas, line 17: The built-in `SqliteDialect` doesn't parse the results either. This comment could be replaced with instructions to use the `ParseJSONResultsPlugin`. Same for the other functions in this module.

koskimas:
After the small comment change, I think this is good to go!

src/helpers/sqlite.ts
jlarmstrongiv: Sounds good 👍 added those notes and instructions

ncrmro:
Howdy yall @koskimas the sqlite pacakge doesn't appear to be in `"kysely": "^0.26.1",` ?

jlarmstrongiv:
@ncrmro follow https://github.com/kysely-org/kysely/issues/628 for updates

PR #589, opened 2023-07-14
Convert `SelectQueryBuilder` and `RawBuilder` to interfaces

koskimas:
My guess is that these two classes are most used ones (in addition to Kysely) in customer code. By turning them to interfaces we get rid of some of the type mismatches when there are multiple versions of Kysely.


igalklebanov (approved):
This hopefully will cause less headaches for us, our users, and https://github.com/serverless-stack/sst's users/maintainers.

LGTM! 🚀 

We should probably add a test that uses a query builder from another Kysely instance as an argument, but with the same Database interface, and verify it still works and doesn't throw compilation errors.

PR #590, opened 2023-07-14
Support boolean subqueries in addition to boolean expressions in filters

koskimas:
There were couple of issues in filter expressions this PR fixes:

1. Filters allowed all kinds of expressions, not just boolean expressions.
2. Filters didn't allow subqueries that select a boolean value.

This PR can be considered a breaking change since people will need to add explicit types for raw sql filters if they didn't have them already. On the other hand this PR fixes a typing bug, so it's open to interpretation.


igalklebanov (approved):
I just love type-safety improvements, great stuff! 👏 

LGTM! 🚀

PR #595, opened 2023-07-17
add MS SQL Server dialect.

### LINKED ISSUE #38: SQL Server support
I am really impress by Kysely and your work. Any plan for SQL Server dialect?

igalklebanov:
might close #38.

This PR adds a core dialect for MS SQL Server. It uses the tedious library as the underlying driver, and the tarn library for connection pool management. All external code is provided through the dialect's config upon instantiation.
It also introduces MS SQL Server to our testing suites.
It also makes necessary small refactors in our testing suites, being more explicit about what's being tested, now that we have more than 3 core dialects.

---

Gaps in our querying or schema APIs relevant to mssql that should be addressed in future PRs:

1. implement the `output` clause so insert id can be retrieved.
2. implement `merge` queries to support `on conflict` like queries.
3. implement `identity` method @ column builders.
4. add `bit`, `uniqueidentifier`, `nvarchar` data types.
5. add `persisted` @ column builders.
6. ability to create things if they don't exist - if possible.
8. docs site - getting started, other places that mention dialects.
9. [computed columns](https://learn.microsoft.com/en-us/sql/relational-databases/tables/specify-computed-columns-in-a-table?view=sql-server-ver16#to-add-a-computed-column-when-creating-a-table).
10. [or alter view](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-view-transact-sql?view=sql-server-ver16).
11. [from at create type](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-type-transact-sql?view=sql-server-ver16).
12. support setting data type and more things (e.g. `not null`) in the same `alter column` clause.
13. [json_value(ref, <json_path>)](https://learn.microsoft.com/en-us/sql/relational-databases/json/json-data-sql-server?view=sql-server-ver16#extract-values-from-json-text-and-use-them-in-queries).
14. json helpers, if possible.
15. `top N` support.
16. `limit offset` support.


src/dialect/mssql/mssql-dialect-config.ts
```diff
@@ -0,0 +1,34 @@
...
+import { Request, TYPES } from 'tedious'
+import { DatabaseConnection } from '../../driver/database-connection.js'
+
+export interface MssqlDialectConfig {
+  /**
+   * A mssql Pool instance or a function that returns one.
+   *
+   * If a function is provided, it's called once when the first query is executed.
+   *
+   * https://github.com/sidorares/node-mysql2#using-connection-pools
```
igalklebanov, line 11: note to self, refer to the `mssql` connection pools readme section, not `mysql2`'s.

src/dialect/mssql/mssql-dialect-config.ts
```diff
@@ -0,0 +1,34 @@
...
+   */
+  pool: ConnectionPool | (() => Promise<ConnectionPool>)
+
+  tedious: Tedious
```
igalklebanov, line 15: note to self, jsdocs here explaining what this is.

package.json
```diff
@@ -73,6 +74,7 @@
     "@types/chai-as-promised": "^7.1.5",
     "@types/chai-subset": "^1.3.3",
     "@types/mocha": "^10.0.1",
+    "@types/mssql": "^8.1.2",
```
igalklebanov, line 77: note to self, remove.

src/parser/binary-operation-parser.ts
```diff
@@ -133,7 +133,11 @@ export function parseFilterList(
   const combine = combinator === 'and' ? AndNode.create : OrNode.create
 
   if (list.length === 0) {
-    return ValueNode.createImmediate(combinator === 'and')
+    return BinaryOperationNode.create(
+      ValueNode.createImmediate(1),
+      OperatorNode.create('='),
+      ValueNode.createImmediate(combinator === 'and' ? 1 : 0)
+    )
```
igalklebanov, line 140: `true` and `false` literals are not allowed in mssql, `1 = 1` and `1 = 0` instead work in all the engines.

dylel:
super excited for this!

vicary:
Sorry for adding noise, but looking at the progress every day is invigorating.

MartinEide:
Amazing work. I have kept my eyes on possible lightweight query builders with good support for typescript against ms server. Crossing my fingers that it works out.

src/dialect/dialect-adapter.ts
```diff
@@ -9,6 +9,14 @@ import { Kysely } from '../kysely.js'
  * it. For that there's a `supportsTransactionalDdl` boolean in this interface.
  */
 export interface DialectAdapter {
+  /**
+   * Whether or not this dialect supports `if not exists` in creation of tables/schemas/views/etc.
+   *
+   * If this is false, tables and schemas are created without `if not exists` in
+   * migrations. This is not a problem if the dialect supports transactional DDL.
```
igalklebanov, line 16: note to self, this should say that Kysely's internal migration tables and schemas.

BStoller:
@igalklebanov any ideas on when we should see this be released? I am actively looking into solutions and would love to use Kysely if SQL Server support were added as that is my DB layer.

igalklebanov:
Ready for review. 😮‍💨

NayzacGithub:
Igal doing the lord's work right here, thank you so much.

MartinEide:
Often these sql servers are hosted in azure. Put any thought into supporting authentication using azure active directory?
https://learn.microsoft.com/en-us/azure/azure-sql/database/azure-sql-javascript-mssql-quickstart

```js

import * as dotenv from 'dotenv';
import sql from 'mssql';
dotenv.config({ path: `.env.${process.env.NODE_ENV}`, debug: true });

const server = process.env.AZURE_SQL_SERVER;
const database = process.env.AZURE_SQL_DATABASE;
const port = parseInt(process.env.AZURE_SQL_PORT);
const type = process.env.AZURE_SQL_AUTHENTICATIONTYPE;


export const config = {
    server,
    port,
    database,
    authentication: {
        type
    },
    options: {
        encrypt: true
    }
};

sql.connect(config);
```

igalklebanov:
@MartinEide we're unopinionated by design. always refer to the underlying driver's (tedious in this case) docs for connection options. 

I have no idea, never used ms sql server in production.

dylel:
It appears that limit(1) isn't converted correctly, currently it adds it to the end of the select statement, however in sql server its called TOP instead of limit.

```
SQL SERVER: SELECT TOP(1) * FROM dbo.MyTable
MySQL: SELECT * FROM dbo.MyTable LIMIT 1
```

I don't think it can be used anywhere else apart from directly after the select keyword and might need to be hoisted there somehow. Maybe it would be better to add a completely different method like .top(1) but I realise that would stray from the current api. It's likely that offset will cause a similar issue as I think mysql can specify offset, limit in one line where as sql server separates them completely but I'm not that familiar with mysql.

Is there another way to call limit in a query, raw sql? i'm currently using this in a nested sub query and would be happy to not have to add a method or touch the current api to get this to work if there was some other way.

 I've been putting this through its paces and seems to be pretty solid overall as this was the only thing I found so far, its my first time really diving into kysely but I'll keep reporting in as I explore the feasibility of bringing my c# codebase to typescript.

TLDR: looks great so far

igalklebanov:
@dylel This PR only handles the mssql dialect. Anything related to querying APIs will or will not be dealt with in other PRs if and when this PR is merged.

Check the tests to see how I managed to pull things off, usually using `modifyX` methods that allow raw SQL before/after certain keywords. These are temporary workarounds.

vicary:
@dylel Do you think it's a good idea. to normalize it into `OFFSET X ROWS FETCH NEXT Y ROWS ONLY` instead of `TOP X`?

dylel:
> @dylel Do you think it's a good idea to normalize it into `OFFSET X ROWS FETCH NEXT Y ROWS ONLY` instead of `TOP X`?

I think the verbose solution would be better, and have them be completely separate methods. I don't like the idea of introducing 'TOP X' and would be in favor of just polyfiling .limit() to have a case for mssql so that users can just use the same api and not have to carry the mental model of "i'm using sql server therefore I have to call a different method" but i'm sure @igalklebanov has some thoughts for future PR's. I've been successfully using the modifyStart/End functions for the time being and they work well for now

vicary:
I don't think `TOP X` is a good idea, because `.offset()` never works.

For the `offset ... fetch` route, we may start with the following overrides in `mssql-query-compiler.ts`,

```ts
  protected override visitLimit(node: LimitNode): void {
    this.append('offset ')
    this.visitNode(node.limit)
    this.append(' rows')
  }

  protected override visitOffset(node: OffsetNode): void {
    this.append('fetch next ')
    this.visitNode(node.offset)
    this.append(' rows only')
  }
```

But it leaves one question, `offset` is always required even if `.offset()` is never called. I am not sure if it's a good idea to copy whole contents of `visitSelectQuery` in an override just to do the following:

```diff
    if (node.limit) {
      this.append(' ')
      this.visitNode(node.limit)
+   } else {
      this.append(' offset 0 rows')
    }
```

src/util/object-utils.ts
```diff
@@ -34,7 +34,7 @@ export function isDate(obj: unknown): obj is Date {
   return obj instanceof Date
 }
 
-export function isBigInt(obj: unknown): obj is BigInt {
+export function isBigInt(obj: unknown): obj is bigint {
```
koskimas, line 37: oof 😬 Nice fix

koskimas (approved):
Finally got to reviewing this giant! Amazing work! I have nothing to nitpick here. Let's merge this and improve as we go 💯

movahhedi:
When will this be released? Need it soo much

movahhedi:
Guys? Any updates?

koskimas:
This is now released with 0.27.0

movahhedi:
@koskimas Thanks for making our lives easier! 🎉🥳

PR #598, opened 2023-07-18
parse json recursively in ParseJsonResultsPlugin

koskimas:
I tested the speed of this thing and it seems to be plenty fast enough. Runtime is a fraction of a millisecond for a result with three levels of nesting and hundreds of objects.


igalklebanov (approved):
LGTM! 🚀

PR #600, opened 2023-07-19
Add `lit` method to expression builder

### LINKED ISSUE #597: Add `.lit(number|boolean|null)` @ `ExpressionBuilder`.
Hey 👋 

`sql` template tag's `.lit` method accepts anything and injects literal values to the query (not parametrized).

`ExpressionBuilder` is everywhere, but doesn't have this functionality, forcing consumers to import `sql` template tag. Let's add it but only accept types that are not dangerous (SQL injection) - numbers, booleans, nulls.

koskimas:
Closes #597


igalklebanov (approved):
LGTM! 🚀

PR #602, opened 2023-07-19
Add between method to expression builder

koskimas:
Don't know if this is a good DX. Feel free to shoot this down.

### DIFF
--- a/src/expression/expression-builder.ts
+++ b/src/expression/expression-builder.ts
@@ -46,6 +46,7 @@ import { ExpressionWrapper } from './expression-wrapper.js'
 import {
   ComparisonOperator,
   JSONOperatorWith$,
+  OperatorNode,
   UnaryOperator,
 } from '../operation-node/operator-node.js'
 import { SqlBool } from '../util/type-utils.js'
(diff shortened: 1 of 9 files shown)


igalklebanov (approved):
I was praying and hoping for this day to come. 😄 

My beloved `between`!!!! ✨ 

Looks good to me! 🚀 

One thing to consider, PostgreSQL supports a [`between symmetric`](https://andreigridnev.com/blog/2016-03-20-between-symmetric-operator-in-postgresql/) that should be a great QoL solution (no need to compare inputs before querying).

How would we tackle this? maybe..

```ts
eb.betweenSymmetric(expr, start, end)
// or
eb.between(expr, start, end).symmetric()
```

koskimas:
Maybe `eb.betweenSymmetric(expr, start, end)` would be the lesser evil? Should I add it?

igalklebanov:
> Maybe `eb.betweenSymmetric(expr, start, end)` would be the lesser evil? Should I add it?

Yeah, seems like an easy win.

koskimas:
> I was praying and hoping for this day to come.

The only reason I was agains `between` was that I didn't want to add `whereBetween`, `orWhereBetween`, `onBetween` etc. everywhere. In the expression builder, it's an easy and natural addition.

igalklebanov:
@frencojobs release pending.

csulit:
I hope we get this. 😁

PR #605, opened 2023-07-20
Support select queries without from clause

koskimas:
This has been requested a lot. Let's hope knex peeps don't get too confused.

`select` returns `SelectQueryBuilder` that has join methods that are not supported without a `from` clause. The alternative was copy-pasting the whole `SelectQueryBuilder` without the join methods. I think this is the lesser evil.


koskimas:
@igalklebanov I changed the method name to `selectNoFrom` as discussed in Discord.

igalklebanov (approved):
LGTM! 🚀

PR #606, opened 2023-07-21
Support arrays in set operations

### LINKED ISSUE #287: allow passing multiple queries to set operations (e.g. `.union(...)`) in a single invocation.
as brought up by @BlakeB415 in #285:

> It would also be great if there could also be a way to provide an array of expressions into the function so we can avoid using a for loop.

koskimas:
Closes #287


igalklebanov (approved):
LGTM! 🚀

PR #611, opened 2023-07-23
Support tuples

### LINKED ISSUE #367: supporting "where in" with multiple columns
Looking for a way to support a multi-column `where in` clause like this:

```ts
const cols = ['colA', 'colB']
const values = [[1, 2], [3, 4]]

qb.where(cols, 'in', values)
```

Which gives a `TypeError: exp.toOperationNode is not a function`

This seems to work instead:

```ts
const whereCol = Array.isArray(whereInCol)
  ? sql`(${sql.join(whereInCol.map((c) => sql.ref(c)))})`
  : whereInCol;
const whereVals = Array.isArray(whereInCol)
  ? sql`(${sql.join(whereInIds.map((ids) => sql`(${sql.join(ids)})`))})`
  : whereInIds;

qb = qb.where(whereCol, "in", whereVals);
```

But wanted to see if there was either a simpler approach I was missing that would be better here, or if this is something that could added to be be handled by kysely if not.

koskimas:
This PR adds tuple support. After trying out a bunch of options, I ended up with a function overload approach up to tuples of length five. There's also a partially unsafe `$asTuple` function in `SelectQueryBuilder` that I don't love.

The reason for my choices basically boils down to the fact that typescript doesn't allow the index of a type in an union to be observed. They've stated in some issues that it will never be observable.

Because of this, there's no (sane) conversion from an object to a tuple, which is why the `$asTuple` method is necessary. If you google this, there is a solution, but it takes around 3 gigabytes of memory to convert a single object to a tuple, so that's a hard no-no.

Using function overloads instead of a recursive type (like the one in coalesce) is also due to performance AND autocompletion issues. The recursive type immediately caused ~one second delays and I wasn't able to get autocompletion to work. Just like we don't currently have autocompletion in `coalesce` after the first argument.

Having a maximum length of five for tuples shouldn't be an issue. My guess is that the most common use case for this is composite primary keys and the length for those is usually 2 or 3. We can also add more overloads if someone requests them.

Closes #367

### DIFF
--- a/src/parser/select-parser.ts
+++ b/src/parser/select-parser.ts
@@ -42,11 +42,13 @@ export type SelectArg<
   | ReadonlyArray<SE>
   | ((eb: ExpressionBuilder<DB, TB>) => ReadonlyArray<SE>)
 
-export type Selection<DB, TB extends keyof DB, SE> = {
-  [A in ExtractAliasFromSelectExpression<SE>]: SelectType<
-    ExtractTypeFromSelectExpression<DB, TB, SE, A>
-  >
-}
+export type Selection<DB, TB extends keyof DB, SE> = [SE] extends [unknown]
+  ? {
+      [A in ExtractAliasFromSelectExpression<SE>]: SelectType<
+        ExtractTypeFromSelectExpression<DB, TB, SE, A>
+      >
+    }
+  : never
 
 type ExtractAliasFromSelectExpression<SE> = SE extends string
   ? ExtractAliasFromStringSelectExpression<SE>
@@ -151,11 +153,13 @@ type ExtractTypeFromStringSelectExpression<
     : never
   : never
 
-export type AllSelection<DB, TB extends keyof DB> = Selectable<{
-  [C in AnyColumn<DB, TB>]: {
-    [T in TB]: C extends keyof DB[T] ? DB[T][C] : never
-  }[TB]
-}>
+export type AllSelection<DB, TB extends keyof DB> = [DB] extends [unknown]
+  ? Selectable<{
+      [C in AnyColumn<DB, TB>]: {
+        [T in TB]: C extends keyof DB[T] ? DB[T][C] : never
+      }[TB]
+    }>
+  : never
 
 export function parseSelectArg(
   selection: SelectArg<any, any, SelectExpression<any, any>>
--- a/src/query-builder/select-query-builder.ts
+++ b/src/query-builder/select-query-builder.ts
@@ -71,6 +71,7 @@ import { KyselyTypeError } from '../util/type-error.js'
 import { Selectable } from '../util/column-type.js'
 import { Streamable } from '../util/streamable.js'
 import { ExpressionOrFactory } from '../parser/expression-parser.js'
+import { ExpressionWrapper } from '../expression/expression-wrapper.js'
 
 export interface SelectQueryBuilder<DB, TB extends keyof DB, O>
   extends WhereInterface<DB, TB>,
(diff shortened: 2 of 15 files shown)


thelinuxlich:
Wouldn't `refTuple` be a better option, following the pattern of `whereRef`?

koskimas:
> Wouldn't `refTuple` be a better option, following the pattern of `whereRef`?

Yeah, but on the other hand, most functions take references by default. I don't know which is the better default.

koskimas:
@thelinuxlich Or did you mean we'd have `refTuple` and `valTuple` or `refTuple` and `tuple`?

thelinuxlich:
refTuple and tuple

thelinuxlich:
because val at least there is no precedent in the API, right?

igalklebanov:
@koskimas @thelinuxlich 
I didn't go over the changes yet, but regarding the naming..
feels like it should be `tuple(...literalsOrExpressions)` and `tupleRef(...refs)`.
it'll be mostly used with literals in where clauses.

igalklebanov:
> @koskimas @thelinuxlich Sorry, I didn't go over the changes yet, but regarding the naming.. feels like it should be `tuple(...literalsOrExpressions)` and `tupleRef(...refs)`. it'll be mostly used with literals in where clauses, and this way it's aligned with other places in the API - consistency.

Actually, when used in where, there's still going to be ref usage in the lhs. this is a tough one.

koskimas:
I think it should be `refTuple` "reference tuple" instead of `tupleRef` "tuple reference". Tuple reference sounds like a reference to a tuple. Instead we have a tuple of references.

koskimas:
> Actually, when used in where, there's still going to be ref usage in the lhs. This is a tough one, but seems like consistency should rule here.

Yeah, for now pretty much the only use case for this is

```ts
eb(refTuple('first_name', 'last_name'), '=', tuple('Jennifer', 'Aniston'))
```

so I'd say `refTuple` will be used just as often as `tuple`.

koskimas:
@igalklebanov @thelinuxlich I renamed the functions.

igalklebanov (commented):
Have you tried an array approach without recursion?

src/parser/select-parser.ts
```diff
@@ -42,11 +42,13 @@ export type SelectArg<
   | ReadonlyArray<SE>
   | ((eb: ExpressionBuilder<DB, TB>) => ReadonlyArray<SE>)
 
-export type Selection<DB, TB extends keyof DB, SE> = {
-  [A in ExtractAliasFromSelectExpression<SE>]: SelectType<
-    ExtractTypeFromSelectExpression<DB, TB, SE, A>
-  >
-}
+export type Selection<DB, TB extends keyof DB, SE> = [SE] extends [unknown]
```
igalklebanov, line 45: is this one of those stack resetting tricks? or just for type errors?

src/query-builder/select-query-builder.ts
```diff
@@ -1414,6 +1415,87 @@ export interface SelectQueryBuilder<DB, TB extends keyof DB, O>
...
+    K3 extends keyof O,
+    K4 extends keyof O,
+    K5 extends keyof O
+  >(
+    key1: K1,
+    key2: K2,
+    key3: K3,
+    key4: K4,
+    key5: K5
+  ): ExpressionWrapper<DB, TB, [O[K1], O[K2], O[K3], O[K4], O[K5]]>
```
igalklebanov, line 1497: Is it possible to enforce there are no duplicates here?

Something like:

```suggestion
  $asTuple<K1 extends keyof O, K2 extends Exclude<keyof O, K1>>(
    key1: K1,
    key2: K2
  ): ExpressionWrapper<DB, TB, [O[K1], O[K2]]>

  $asTuple<
    K1 extends keyof O,
    K2 extends Exclude<keyof O, K1>,
    K3 extends Exclude<keyof O, K1 | K2>
  >(
    key1: K1,
    key2: K2,
    key3: K3
  ): ExpressionWrapper<DB, TB, [O[K1], O[K2], O[K3]]>

  $asTuple<
    K1 extends keyof O,
    K2 extends Exclude<keyof O, K1>,
    K3 extends Exclude<keyof O, K1 | K2>,
    K4 extends Exclude<keyof O, K1 | K2 | K3>
  >(
    key1: K1,
    key2: K2,
    key3: K3,
    key4: K4
  ): ExpressionWrapper<DB, TB, [O[K1], O[K2], O[K3], O[K4]]>

  $asTuple<
    K1 extends keyof O,
    K2 extends Exclude<keyof O, K1>,
    K3 extends Exclude<keyof O, K1 | K2>,
    K4 extends Exclude<keyof O, K1 | K2 | K3>,
    K5 extends Exclude<keyof O, K1 | K2 | K3 | K4>
  >(
    key1: K1,
    key2: K2,
    key3: K3,
    key4: K4,
    key5: K5
  ): ExpressionWrapper<DB, TB, [O[K1], O[K2], O[K3], O[K4], O[K5]]>
```

igalklebanov (approved):
Amazing stuff! 👨‍🎨 

Left some comments..

src/query-builder/select-query-builder.ts
igalklebanov: Also, there's probably a way to enforce all keys from `O` were used.

koskimas: That just enforces one of the columns.

src/parser/select-parser.ts
koskimas: Forces the type to be evaluated in-place. You see the resulting type in the IDE instead of `Selection<...>`. Also doesn't increase the stack depth.

src/query-builder/select-query-builder.ts
koskimas: No wait... You're right. That should work

PR #622, opened 2023-07-29
fix create index `using` support @ MySQL.

igalklebanov:
Hey :wave:

Currently the way we compile create index queries with the `using` keyword is not aligned with MySQL's spec.
MySQL's spec accepts `using` only before the `on` keyword.

This PR flips the order for MySQL.


koskimas (approved):
Would be nice to not have to duplicate the whole function, but I don't know how to do it. Adding a separate flag for this is equally bad (we've done that too in some places).

I think this is good as-is 👍

PR #634, opened 2023-08-04
Allow dialects and users to modify internal Kysely types

koskimas:
This is a big one 😓

The idea is to carry around a bag of configuration types in addition to the database tables.

Now we have the `DB` type that's the first generic argument of pretty much every class and interface in Kysely. It simply contains the tables and their columns in the DB along with any joined subqueries and aliased tables/subqueries, added CTEs etc. Basically it contains the tables that are possible to query from in the current context.

The idea is to turn that type from a set of tables to `{ tables: Tables, config: TypeConfig }`. `Tables` is the old `DB` type and `TypeConfig` can contain all kinds of configuration that'll affect the types throughout the API.

The first and biggest reason to consider adding this is allowing the dialects and the user override some DB types like the default boolean type.

Now the boolean type is always `type SqlBool = boolean | 0 | 1` since currently the types are unaware of the underlying dialect and any custom parsing installed by the user. That type is the superset of all known dialects' default boolean types. This PR would allow `PostgresDialect` to specify that it's boolean type is `boolean` and `MysqlDialect` to specify it's `0 | 1`. This PR would also allow the user to change the type if they've configured the underlying database driver to return something else.

There are a couple of breaking changes in the current implementation:

### The DB type changes

As mentioned before, the DB type will change from `Tables` to `{ tables: Tables, config: TypeConfig }`. All existing type generators will still work. The type they create will be passed to the `tables` property.

The only change needed in the 3rd party type generators is to change the type name from `DB` to `Tables` to avoid confusion. Even that's not necessary, but could be a good change.

### Kysely constructor can't be used anymore

Because typescript doesn't allow partial inference of function arguments, we need to change the way `Kysely` instances are created. We need to be able to infer any type config specified by the dialect. The proposed way is the following:

```ts
import { kysely, InferDB } from 'kysely'

// The type of `db` is still `Kysely<DB>`.
export const db = kysely<Tables>()
  // We need a separate function call for setting the dialect
  // to be able to pick any type configuration it may or may not
  // contain.
  .dialect(new PostgresDialect(...))
  // The user can override any type configuration like this.
  .typeConfig<{ bigIntType: bigint }>()
  .log(logSettings)
  .plugins(plugins)
  .build()

// You can extract the new `DB` type like this. The `DB` consists of
// the `Tables` type and any configuration picked and merged from the
// `dialect` and `typeConfig` method calls.
export type DB = InferDB<typeof db>
```

## Is it worth it

I don't know. This a lot of changes and can potentially destroy the type performance and maintainability in the future. Currently all we gain is the correct `boolean` type when you select a boolean expression and the correct nubmer type in some built-in functions like `count` and `countAll`.

The added complexity is significant. The types got about 30% slower (according to our typing test set) but this PR also contains some unrelated optimizations that erased the 30% slowdown. If we don't merge this, we can add the same optimizations to the current master branch.


igalklebanov:
First off, great stuff! tough changes and to make it all work. 👏 

I like this solution, it's the cleanest of all we discussed. But as you pointed out, it comes with a compile-time performance hit, and it's a solution looking for a problem that's worth it all.
Also, like you, I'm not quite sure how this affects Kysely's extensibility, how it affects helpers, and other things our users can and do today, or will be able to do tomorrow.

koskimas:
This turned out to be a really bad idea. Too complicated for a relatively small impact.

PR #635, opened 2023-08-05
Add SQLocal to 3rd party dialects list

DallasHoff:
I've been building a library designed to make working with the WebAssembly build of SQLite easier, and it includes a dialect for Kysely, so I would like to add it to the 3rd party dialects list on the Kysely site.

Repo: https://github.com/DallasHoff/sqlocal
Guide: https://sqlocal.dallashoffman.com/kysely/setup


igalklebanov (approved):
LGTM! 🚀

PR #637, opened 2023-08-07
Update Deno SQLite page

alexgleason:
It's now possible to use SQLite with Deno and Kysely: https://gitlab.com/soapbox-pub/kysely-deno-sqlite

### DIFF
--- a/site/docs/runtimes/deno.mdx
+++ b/site/docs/runtimes/deno.mdx
@@ -5,50 +5,17 @@ title: 'Running on Deno'
 
 # Running on Deno
 
-Kysely doesn't include drivers for Deno, but you can still use Kysely as a query builder or implement your own driver:
+It's possible to use SQLite with Deno by using [`kysely-deno-sqlite`](https://gitlab.com/soapbox-pub/kysely-deno-sqlite).
+This method uses [deno-sqlite](https://github.com/dyedgreen/deno-sqlite) under the hood.
 
 ```ts
-// We use jsdeliver to get Kysely from npm.
-import {
-  DummyDriver,
-  Generated,
-  Kysely,
-  PostgresAdapter,
-  PostgresIntrospector,
-  PostgresQueryCompiler,
-} from 'https://cdn.jsdelivr.net/npm/kysely/dist/esm/index.js'
+import { Kysely } from 'npm:kysely';
+import { DB as Sqlite } from 'https://deno.land/x/sqlite/mod.ts';
+import { DenoSqliteDialect } from 'https://gitlab.com/soapbox-pub/kysely-deno-sqlite/-/raw/main/mod.ts';
 
-interface Person {
-  id: Generated<number>
-  first_name: string
-  last_name: string | null
-}
-
-interface Database {
-  person: Person
-}
-
-const db = new Kysely<Database>({
-  dialect: {
-    createAdapter() {
-      return new PostgresAdapter()
-    },
-    createDriver() {
-      // You need a driver to be able to execute queries. In this example
-      // we use the dummy driver that never does anything.
-      return new DummyDriver()
-    },
-    createIntrospector(db: Kysely<unknown>) {
-      return new PostgresIntrospector(db)
-    },
-    createQueryCompiler() {
-      return new PostgresQueryCompiler()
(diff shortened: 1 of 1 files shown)


site/docs/runtimes/deno.mdx
```diff
@@ -5,50 +5,17 @@ title: 'Running on Deno'
...
-console.log(sql.sql)
+import { Kysely } from 'npm:kysely';
+import { DB as Sqlite } from 'https://deno.land/x/sqlite/mod.ts';
+import { DenoSqliteDialect } from 'https://gitlab.com/soapbox-pub/kysely-deno-sqlite/-/raw/main/mod.ts';
+
+const db = new Kysely({
+  dialect: new DenoSqliteDialect({
+    database: new Sqlite('db.sqlite3'),
+  }),
+});
```
igalklebanov, line 20: I don't think it's a good idea to provide examples using community dialects in the core library's documentation.
We do promote community dialects [here](https://github.com/kysely-org/kysely/blob/master/site/docs/dialects.md) and [here](https://github.com/kysely-org/awesome-kysely).

site/docs/runtimes/deno.mdx
```diff
@@ -5,50 +5,17 @@ title: 'Running on Deno'
 
 # Running on Deno
 
-Kysely doesn't include drivers for Deno, but you can still use Kysely as a query builder or implement your own driver:
+It's possible to use SQLite with Deno by using [`kysely-deno-sqlite`](https://gitlab.com/soapbox-pub/kysely-deno-sqlite).
+This method uses [deno-sqlite](https://github.com/dyedgreen/deno-sqlite) under the hood.
```
igalklebanov, line 9: I don't think this is the right place to promote specific community dialects.
We do promote community dialects [here](https://github.com/kysely-org/kysely/blob/master/site/docs/dialects.md) and [here](https://github.com/kysely-org/awesome-kysely).

The original sentence is partially true, now that Deno supports `npm:<library>` installations and we've got a test proving `PostgresDialect` and `MysqlDialect` work in Deno (not SqliteDialect sadly). Should be re-worded to say that Kysely's PostgreSQL and MySQL dialects work in Deno, but for SQLite there are community dialects - and linking to the dialects page (https://github.com/kysely-org/kysely/blob/master/site/docs/dialects.md).

site/docs/runtimes/deno.mdx
```diff
@@ -5,50 +5,17 @@ title: 'Running on Deno'
...
-    createQueryCompiler() {
-      return new PostgresQueryCompiler()
-    },
-  },
-})
-
-const query = db.selectFrom('person').select('id')
-const sql = query.compile()
-
-console.log(sql.sql)
```
igalklebanov, line 53: We already explain how to use Kysely in Deno with Postgres and MySQL in "getting started", could just link to that instead of an example (https://kysely.dev/docs/getting-started?package-manager=deno).

igalklebanov (changes requested):
Thank you! 🙏 

I don't think the proposed changes are a match. See comments..

alexgleason:
The unique part of my project is the 12 lines of code here: https://gitlab.com/soapbox-pub/kysely-deno-sqlite/-/blob/1705a17d02cd3cdc5ac94b6aa7f549c8c9696c99/src/deno-sqlite-driver.ts#L63-74

I wonder if the existing driver can be refactored to accept either type, or if some other solution could get this mainlined. Then it's not about editing docs but about making it work with Deno by default.

I had originally written a wrapper around deno-sqlite to make it expose better-sqlite's API, but there was a hack in the original driver code that caused me to duplicate the whole driver and edit that part.

See:

https://github.com/kysely-org/kysely/blob/a657787ef4b869f79f45b3dd11c7f950e19a6fd0/src/dialect/sqlite/sqlite-driver.ts#L72

If we can make the driver accept a better interface, we could just adapt each library instead of building a new driver every time.

(I'm also hoping the cool SQLocal driver added recently could mean Deno support in the future, which is being discussed for this very reason: https://github.com/denoland/deno/issues/17833#issuecomment-1616174068)

koskimas:
No activity in a while here. Closing.

PR #645, opened 2023-08-12
fix node-postgres `pool` link

nick-w-nick:
I was browsing the [PostgresPool](https://kysely-org.github.io/kysely-apidoc/interfaces/PostgresPool.html) section of the documentation and noticed that the link to the node-postgres pool API doc was 404ing as it had `/api/` instead of `/apis/` in the link.

Thought I'd submit a quick PR to fix it.

Thanks!


igalklebanov (approved):
Thank you! 💪

PR #656, opened 2023-08-22
Simplify .withTables() return type

codetheweb:
We have a complex schema definition with over 130 tables, and type resolution when passing around a `Transaction` object was very slow because of this:

`type SlowTypeResolution = Transaction<KyselySchema> extends Kysely<KyselySchema> ? true : false`

E.x.

```ts
const runDirectlyOrInTransaction = async (db: Kysely<KyselySchema>) => {
  await db.selectFrom("foo").selectAll().execute()
}

await kysely.transaction().execute(async trx => {
  await runDirectlyOrInTransaction(trx)
})

```

According to a TS trace, the hotspot was the `.withTables()` generic.

With this fix, type resolution time on our project went from > 30s to 4s.

### DIFF
--- a/src/kysely.ts
+++ b/src/kysely.ts
@@ -32,6 +32,7 @@ import { CaseNode } from './operation-node/case-node.js'
 import { parseExpression } from './parser/expression-parser.js'
 import { Expression } from './expression/expression.js'
 import { WithSchemaPlugin } from './plugin/with-schema/with-schema-plugin.js'
+import { DrainOuterGeneric } from './util/type-utils.js'
 
 /**
  * The main Kysely class.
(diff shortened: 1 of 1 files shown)


koskimas:
Thank you!

`Simplify` also loses the underlying types and will convert `DB` into an inline object with 130 properties. You should get the same speedup just by making the type expressions conditional. Something like this

```ts
  withTables<T extends Record<string, Record<string, any>>>(): Kysely<
    [T] extends [unknown] ? DB & T : never
  >
```

We have a `DrainOuterGeneric` helper that does just that:

```ts
  withTables<T extends Record<string, Record<string, any>>>(): Kysely<
    DrainOuterGeneric<DB & T>
  >
```

let me know if you get the same speedup like this

koskimas:
I haven't tested Kysely with this many tables. Please let us know of any additional problems and slowdowns you might experience!

codetheweb:
Yes, that seems to work too. Updated this PR.

Thanks for the quick review!

koskimas (approved):

PR #657, opened 2023-08-22
Fixing a typo in the relations docs

gittgott:
Relations docs had a duplicated word "rows". This fixes that typo


koskimas:
Thanks!

PR #658, opened 2023-08-23
feat/add cjs extension

mtt-artis:
hello there 

this PR allow to migrate cjs files

migrate in esm on windows seem complicated
so i make a transpilation in esm for my project except for migration part which is in cjs 
tsup add the cjs extension to the file

### DIFF
--- a/src/migration/file-migration-provider.ts
+++ b/src/migration/file-migration-provider.ts
@@ -31,6 +31,7 @@ export class FileMigrationProvider implements MigrationProvider {
     for (const fileName of files) {
       if (
         fileName.endsWith('.js') ||
+        fileName.endsWith('.cjs') ||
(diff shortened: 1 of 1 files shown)


mtt-artis:
but would be awesome if the migration works on window in esm :)

koskimas:
It does work. All you need to do is implement your own `MigrationProvider`. It's one function.

mtt-artis:
yes i agree it's not difficult to write my own migration file:

but 
- i think it's better if it works seamless by following the doc (no matter if esm/cjs and win/linux)
- it could remove the question on the discord

i close this PR and link another that might resolve this 
feel free to close it as well if you don't want/like it ;)
https://github.com/kysely-org/kysely/pull/659

thanks again for the lib, and the time you give to the community

PR #430, opened 2023-04-18
fix: fixed ES Module loading with abolute path fails on windows

MrMenezesDev:
Fixed error on windows

Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file and data are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
    at new NodeError (node:internal/errors:399:5)
    at throwIfUnsupportedURLScheme (node:internal/modules/esm/resolve:1059:11)
    at defaultResolve (node:internal/modules/esm/resolve:1135:3)
    at nextResolve (node:internal/modules/esm/loader:163:28)
    at defer (C:\Projetos\cubos\api-typescript-sdkgen-kysely\node_modules\.pnpm\ts-node@10.9.1_@swc+core@1.3.49_@types+node@18.15.11_typescript@4.9.5\node_modules\ts-node\src\esm.ts:159:23)
    at entrypointFallback (C:\Projetos\cubos\api-typescript-sdkgen-kysely\node_modules\.pnpm\ts-node@10.9.1_@swc+core@1.3.49_@types+node@18.15.11_typescript@4.9.5\node_modules\ts-node\src\esm.ts:168:34)
    at C:\Projetos\cubos\api-typescript-sdkgen-kysely\node_modules\.pnpm\ts-node@10.9.1_@swc+core@1.3.49_@types+node@18.15.11_typescript@4.9.5\node_modules\ts-node\src\esm.ts:202:16
    at addShortCircuitFlag (C:\Projetos\cubos\api-typescript-sdkgen-kysely\node_modules\.pnpm\ts-node@10.9.1_@swc+core@1.3.49_@types+node@18.15.11_typescript@4.9.5\node_modules\ts-node\src\esm.ts:409:21)
    at resolve (C:\Projetos\cubos\api-typescript-sdkgen-kysely\node_modules\.pnpm\ts-node@10.9.1_@swc+core@1.3.49_@types+node@18.15.11_typescript@4.9.5\node_modules\ts-node\src\esm.ts:197:12)
    at nextResolve (node:internal/modules/esm/loader:163:28)

PR #659, opened 2023-08-23
Feat/migrator

mtt-artis:
Hey

This PR introduce de `protocol` key on the `FileMigrationProvider` 
This allows window users to add a protocol if they need to 

https://github.com/kysely-org/kysely/pull/430

Thanks again for this amazing library


mtt-artis:
https://github.com/kysely-org/kysely/pull/430#issuecomment-1691251528

src/migration/file-migration-provider.ts
```diff
@@ -31,11 +31,13 @@ export class FileMigrationProvider implements MigrationProvider {
     for (const fileName of files) {
       if (
         fileName.endsWith('.js') ||
+        fileName.endsWith('.cjs') ||
         (fileName.endsWith('.ts') && !fileName.endsWith('.d.ts')) ||
         fileName.endsWith('.mjs') ||
         (fileName.endsWith('.mts') && !fileName.endsWith('.d.mts'))
       ) {
         const migration = await import(
+            this.#props.protocol +
```
koskimas, line 40: This is now `undefined` by default and will break everything.

mtt-artis: sorry 🙏 
did this qwickly from work on the github vscode web interface 
this should fix it

i would prefer rather this solution than this ?

```ts
const isEsm = Boolean(typeof module !== 'undefined' && typeof module.imports !== 'undefined') 
this.#props.path.join(
  isEsm ? "file://" : "",
  this.#props.migrationFolder,
  fileName
)
```

mtt-artis:
Thanks !

PR #660, opened 2023-08-24
Add Xata as a dialect

SferaDev:
Xata is a serverless data platform built on top of PostgreSQL, as a new feature we are now exposing SQL over HTTP and we have built a driver for Kysely.

Implementation: https://github.com/xataio/client-ts/blob/main/packages/plugin-client-kysely/src/driver.ts


igalklebanov (approved):

igalklebanov:
Hey 👋 

Thanks! 🚀

PR #665, opened 2023-08-26
docs: improve "splitting build, compile and execute code" recipe

virtuallyunknown:
Follow up from #662. As per your advice that it would be better if all the examples shown in this recipe worked without unintentional errors, I've added a minimal `Database` interface as seen in other places throughout the documentation.

I was very tempted to leave a note about instantiating the `Kysely` instance with an `any` as a generic type parameter, but ultimately opted against, since it might be conflicting with later examples on this page which make use of `InferResult`.

Cheers!


koskimas:
Thank you!

PR #669, opened 2023-08-28
Fix Node.js stack traces when code in a transaction throws

codetheweb:
The error stack was losing context because the Promise chain [must be unbroken](https://cloudreports.net/v8-zero-cost-async-stack-traces/) for the error context to correctly work.

E.x. on an internal codebase:

```ts
await kysely.transaction().execute(async trx => {
  // generates invalid query
  await trx.selectFrom("foo").where("foo_id", 'in', []).execute()
})
```

before:

Screenshot 2023-08-28 at 4 18 35 PM

after:

Screenshot 2023-08-28 at 4 19 14 PM

(`tests/db/tests/db/postgres-error-stacks.test.ts` is the path to the file that throws)

Added test was previously failing.

### DIFF
--- a/src/driver/single-connection-provider.ts
+++ b/src/driver/single-connection-provider.ts
@@ -1,9 +1,11 @@
 import { DatabaseConnection } from './database-connection.js'
 import { ConnectionProvider } from './connection-provider.js'
 
+const ignoreError = () => {}
+
 export class SingleConnectionProvider implements ConnectionProvider {
   readonly #connection: DatabaseConnection
-  #runningPromise?: Promise<void>
+  #runningPromise?: Promise<any>
 
   constructor(connection: DatabaseConnection) {
     this.#connection = connection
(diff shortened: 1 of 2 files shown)


src/driver/single-connection-provider.ts
```diff
@@ -17,16 +17,11 @@ export class SingleConnectionProvider implements ConnectionProvider {
...
-        this.#runningPromise = undefined
-      })
-      .catch(() => {
-        this.#runningPromise = undefined
-      })
 
-    return promise
+    return promise.finally(() => {
+      this.#runningPromise = undefined
+    })
```
koskimas, line 24: Doesn't this break the whole logic? `this.#runningPromise` now doesn't have the finally block. When you await `this.#runningPromise`, it won't be set to undefined once the promise resolves. The while loop above will go to a busy loop for a while.

koskimas: It probably works, but this may have destroyed the performance.

koskimas: Another issue. The promise stored to `this.#runningPromise` will now throw and the second query waiting for a connection will also fail.

koskimas:
Thanks, but this caused at least two big new issues.

koskimas:
This should work correctly

```ts
  async provideConnection<T>(
    consumer: (connection: DatabaseConnection) => Promise<T>
  ): Promise<T> {
    while (this.#runningPromise) {
      await this.#runningPromise.catch(ignoreError)
    }

    // `#runningPromise` must be set to undefined before it's
    // resolved or rejected. Otherwise the while loop above
    // will misbehave.
    this.#runningPromise = this.#run(consumer).finally(() => {
      this.#runningPromise = undefined
    })

    return this.#runningPromise
  }
```

```ts
function ignoreError(_err: unknown) {
  // Do nothing.
}
```

test/node/src/transaction.test.ts
```diff
@@ -154,6 +156,24 @@ for (const dialect of DIALECTS) {
...
+          await trx.selectFrom('person').where('id', 'in', -1).execute()
+        })
+
+        expect.fail('Expected transaction to fail')
+      } catch (error) {
+        if (dialect === 'sqlite') {
+          expect(error).to.be.instanceOf(SqliteError)
+        } else if (dialect === 'postgres') {
+          expect(error).to.be.instanceOf(PostgresError)
+        }
```
codetheweb, line 171: not very familiar with chai; maybe there's a cleaner way to assert this

koskimas:
Thank you @codetheweb for bringing this up and fixing it 💯

test/node/src/transaction.test.ts
codetheweb: More just to make sure the error being thrown is a database error and not an error in our test code (in which case the assertion below would still pass).

Issue #679, opened 2023-09-06
Add TiDB Serverless Driver


shiyuhang0:
Hi, I'd like to add a 3rd party dialect called [TiDB Serverless Driver](https://github.com/tidbcloud/serverless-js) and add it to the official 3rd party dialects list https://kysely.dev/docs/dialects#3rd-party-dialects

Is there some necessary process to get this done?

koskimas:
Just create a PR and we'll merge it.

shiyuhang0:
Thanks! I will develop the plugin then open a pr

PR #695, opened 2023-09-12
Add tidb serverless dialect

shiyuhang0:
close https://github.com/kysely-org/kysely/issues/679

1. Hi, I developed a TiDB serverless dialect for kysely here https://github.com/tidbcloud/kysely. Any review is welcomed! Write your comment in the issue
2. It's better to merge this PR in the next week. Because our TiDB serverless driver will be available in the next week
3. I have tested it in our dev environment


koskimas:
Thanks!

PR #704, opened 2023-09-18
Passing json tests for mssql dialect

gittgott:
This adds the `jsonArrayFrom`, `jsonObjectFrom`, and `jsonBuildObject` helpers for mssql.
Additionally, the `visitOffset` and `visitLimit` methods were added to the `mssql-query-compiler` as well as some updates to the visitSelectQuery method to allow for the tests to pass where `limit` is used.

### DIFF
--- a/test/node/src/json.test.ts
+++ b/test/node/src/json.test.ts
@@ -16,6 +16,11 @@ import {
   jsonBuildObject as mysql_jsonBuildObject,
 } from '../../../helpers/mysql'
 import {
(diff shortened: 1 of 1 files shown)


igalklebanov:
Hey 👋

Thank you! 🫶

IMHO, if this requires proper limit offset support, its a premature PR. These things should not be rushed, and your compiler-level hack goes against WYSIWYG - a core design principle, so I cannot approve these changes.

Issue #730, opened 2023-10-11
Support `JOIN ... USING ( col1, col2, ... )`


dwickern:
Many databases support this shorthand for joins when the column names are the same in both tables.

From Postgres docs:

> The `USING` clause is a shorthand that allows you to take advantage of the specific situation where both sides of the join use the same name for the joining column(s). It takes a comma-separated list of the shared column names and forms a join condition that includes an equality comparison for each one. For example, joining `T1` and `T2` with `USING (a, b)` produces the join condition `ON T1.a = T2.a AND T1.b = T2.b`.
> 
> ```
> => SELECT * FROM t1 INNER JOIN t2 ON t1.num = t2.num;
>  num | name | num | value
> -----+------+-----+-------
>    1 | a    |   1 | xxx
>    3 | c    |   3 | yyy
> (2 rows)
> 
> => SELECT * FROM t1 INNER JOIN t2 USING (num);
>  num | name | value
> -----+------+-------
>    1 | a    | xxx
>    3 | c    | yyy
> (2 rows)
> ```

PostgreSQL: https://www.postgresql.org/docs/current/queries-table-expressions.html
MySQL https://dev.mysql.com/doc/refman/8.0/en/join.html
Sqlite: https://www.sqlite.org/syntax/join-constraint.html
Oracle: https://docs.oracle.com/javadb/10.6.2.1/ref/rrefsqljusing.html
SQL Server: not supported

Usage could be something like
```ts
db.selectFrom('person as a')
  .innerJoinUsing('person as b', ['first_name', 'last_name'])
  .select(['a.id', 'b.id'])
```

koskimas:
I don't know if this is worth the complication in the code. Expecially if we did this in the most type-safe way possible, we should only accept columns both tables have. That'd be tricky.

But if we do add this, we could add it to the join builder

```ts
db.selectFrom('person as a')
  .innerJoin('person as b', join.using(['first_name', 'last_name']))
  .select(['a.id', 'b.id'])
```

igalklebanov:
@koskimas Apparently, they behave differently:

> A `USING` clause can be rewritten as an `ON` clause that compares corresponding columns. However, although `USING` and `ON` are similar, they are not quite the same. Consider the following two queries:
> 
> ```sql
> a LEFT JOIN b USING (c1, c2, c3)
> a LEFT JOIN b ON a.c1 = b.c1 AND a.c2 = b.c2 AND a.c3 = b.c3
> ```
[...]

PR #733, opened 2023-10-12
add proof-of-concept for join ... using

dwickern:
This is a proof-of-concept for #730. It mainly features `JoinUsingExpression` which calculates the columns present in both tables.

I'm using my originally proposed syntax `innerJoinUsing` for a couple of reasons:
1. Since `JoinBuilder` has type parameters `TB` and `TE` union'd together, it would be difficult to implement a `JoinUsingExpression` without refactoring `JoinBuilder`'s interface
2. The `join ... using` clause cannot be combined with `join ... on`, so there's no use being able to chain together `join.using(...).on(...).onRef(...)`

But if you still prefer the callback syntax, I'm sure we could make it work

### DIFF
--- a/src/parser/join-parser.ts
+++ b/src/parser/join-parser.ts
@@ -36,6 +36,14 @@ type AnyJoinColumnWithTable<DB, TB extends keyof DB, TE> = AnyColumnWithTable<
   FromTables<DB, TB, TE>
 >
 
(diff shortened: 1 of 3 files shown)


koskimas:
The syntax should be

```ts
.innerJoin('pet', join => join.using('id'))
```

PR #742, opened 2023-10-20
Update generating-types.md

deanc:
I've hacked together a new library using node-sql-parser to extract types. I took a look into some of the existing solutions with 1000s of lines of code and thought I could probably do it a lot easier and at least one of the listed libraries is not using the types provided by Kysely (e.g. Insertable, Updateable etc.)

This PR adds a link in the docs :)


igalklebanov (approved):
Hey 👋 

Thank you! 💪

PR #746, opened 2023-10-25
site: 'timestamp' instead of 'text' on migration example

endel:
Small update on the SQLite migration example from the website: https://kysely.dev/docs/migrations#sqlite-migration-example

Replace this:

```typescript
    .addColumn('created_at', 'text', (col) =>
      col.defaultTo(sql`CURRENT_TIMESTAMP`).notNull()
    )
```

With this:

```typescript
    .addColumn('created_at', 'timestamp', (col) =>
      col.defaultTo(sql`CURRENT_TIMESTAMP`).notNull()
    )
```


igalklebanov:
Hey 👋 

Thank you! 💪 

I'm not quite sure this should be encouraged:

a. there is no `timestamp` data type in sqlite.
b. `timestamp` has affinity type `numeric`, and `numeric` can store any data type ([source](https://dev.to/chshersh/sqlite-type-system-is-the-wooorst-mi7)).

endel:
Wow, thank you for the explanation and for sharing the article 🤯 Closing...

PR #747, opened 2023-10-25
fix errors for contributors on Windows

wirekang:
### scripts/module-fixup.js
Since node 'path' module uses platform specific path separator, git-ignored `helpers/` files could be generated with escape character `\`:
`export * from '..\dist\cjs\helpers\postgres.js'`

Updated `path` -> `path.posix` to use `/` as path separator always. 

### New dev-dependency
**cross-env**: for setting environment variables regardless of platform while testing.

---

I found this errors recently by moving primary OS from Ubuntu to Windows.
I can't completly sure this is Windows-related because it's hard to believe that out of 71 contributors, none of them use Windows.


igalklebanov:
> I can't completly sure this is Windows-related because it's hard to believe that out of 71 contributors, none of them use Windows.

I use WSL2, so I never had this issue.

PR #754, opened 2023-11-01
feat: add stream() support for sqlite dialect

tgriesser:
- Adds support for `stream` via the [`.iterate(parameters)`](https://github.com/WiseLibs/better-sqlite3/blob/master/docs/api.md#iteratebindparameters---iterator) API in `better-sqlite3`


test/node/src/select.test.ts
```diff
@@ -879,16 +879,45 @@ for (const dialect of DIALECTS) {
       })
     }
 
-    if (dialect === 'postgres' || dialect === 'mysql' || dialect === 'mssql') {
```
tgriesser, line 882: Could also change this to `|| dialect === 'sqlite'` which would make the diff less noisy, but also seemed strange it then includes all the dialects.

src/dialect/sqlite/sqlite-driver.ts
```diff
@@ -92,8 +92,24 @@ class SqliteConnection implements DatabaseConnection {
...
 
-  async *streamQuery<R>(): AsyncIterableIterator<QueryResult<R>> {
-    throw new Error("Sqlite driver doesn't support streaming")
+  async *streamQuery<R>(
+    compiledQuery: CompiledQuery,
+    _chunkSize: number
+  ): AsyncIterableIterator<QueryResult<R>> {
+    const { sql, parameters, query } = compiledQuery
+    const stmt = this.#db.prepare(sql)
+    if (query.kind === 'SelectQueryNode') {
```
igalklebanov, line 101:
```suggestion
    if (SelectQueryNode.is(query)) {
```

src/dialect/sqlite/sqlite-driver.ts
```diff
@@ -92,8 +92,24 @@ class SqliteConnection implements DatabaseConnection {
     }
   }
 
-  async *streamQuery<R>(): AsyncIterableIterator<QueryResult<R>> {
-    throw new Error("Sqlite driver doesn't support streaming")
+  async *streamQuery<R>(
+    compiledQuery: CompiledQuery,
+    _chunkSize: number
```
igalklebanov, line 97: can we implement it with usage of `chunkSize`?

src/dialect/sqlite/sqlite-driver.ts
```diff
@@ -92,8 +92,24 @@ class SqliteConnection implements DatabaseConnection {
...
+    if (query.kind === 'SelectQueryNode') {
+      const iter = stmt.iterate(parameters) as IterableIterator<R>
+      for (const row of iter) {
+        yield {
+          rows: [row],
+        }
+      }
+    } else {
+      throw new Error(
+        'Sqlite driver only supports queries on select operations'
```
igalklebanov, line 110:
```suggestion
        'Sqlite driver only supports streaming of select queries'
```

igalklebanov (commented):
Hey 👋

Thank you! 💪

Left some comments..

src/dialect/sqlite/sqlite-driver.ts
tgriesser: It was a little unclear what `chunkSize` is meant for - is this for the "#" of rows returned at once? Or is it meant in association with the `chunkSize` option for streams.

In the docstring here, it says it's only supported in PG: https://github.com/kysely-org/kysely/blob/0982d61a59d6dccaaf2a026cad82a77b843be7ac/src/query-executor/query-executor.ts#L59-L63

Though it looks like it's also used in mssql: https://github.com/kysely-org/kysely/blob/0982d61a59d6dccaaf2a026cad82a77b843be7ac/src/dialect/mssql/mssql-driver.ts#L181-L191

Both of those seem to be using it in association with the underlying stream, and it looks to be unused in `mysql`:

https://github.com/kysely-org/kysely/blob/0982d61a59d6dccaaf2a026cad82a77b843be7ac/src/dialect/mysql/mysql-driver.ts#L181-L192

So since the iterator returned from better-sqlite isn't actually a stream, I decided to ignore it.

koskimas (approved):
LGTM! Thank you!

PR #763, opened 2023-11-08
export SelectQueryBuilderImpl

Andersgee:
We have been adding an alternative `.execute()`. I noticed that the actual class (implementation) for SelectQueryBuilder is not exported in recent versions.

The classes for update, delete and insert query builders are already exported.


koskimas:
This class is private and not meant to be exported.

Issue #690, opened 2023-09-10
Support for `CREATE TABLE table AS select_statement`?


AlexErrant:
I think this is a feature request, but does Kysely support creating a new (potentially temporary) table from a select statement?

This feature seems to be supported by the major databases; [MySQL](https://dev.mysql.com/doc/refman/8.0/en/create-table-select.html), [PostGres](https://www.postgresql.org/docs/current/sql-createtableas.html), [Sqlite](https://www.sqlite.org/lang_createtable.html).

koskimas:
Sounds like a good addition 👍. Let's add it.

The API would be:

```ts
db.schema
  .createTable('new_table')
  .as(db.selectFrom('person').select(['id', 'first_name']))
```

LeoOmori:
hi, can i try to work on this?

igalklebanov:
@LeoOmori Go ahead, good luck! If you have any question, feel free to post in the contribution channel on discord.

viraxslot:
Hi all, there's no activity since september so I'd like to work on this issue.
@LeoOmori please ping me if you started to implement something and I have to stop :)

PR #771, opened 2023-11-20
AS statement support for createTable

viraxslot:
Closes: https://github.com/kysely-org/kysely/issues/690

Done:
- added support of `as` statement for `createTable`
- added unit tests
- added tests for typings

### DIFF
--- a/src/operation-node/operation-node-transformer.ts
+++ b/src/operation-node/operation-node-transformer.ts
@@ -411,7 +411,7 @@ export class OperationNodeTransformer {
       onCommit: node.onCommit,
       frontModifiers: this.transformNodeList(node.frontModifiers),
       endModifiers: this.transformNodeList(node.endModifiers),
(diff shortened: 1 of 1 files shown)


src/operation-node/operation-node-transformer.ts
```diff
@@ -411,6 +411,7 @@ export class OperationNodeTransformer {
       onCommit: node.onCommit,
       frontModifiers: this.transformNodeList(node.frontModifiers),
       endModifiers: this.transformNodeList(node.endModifiers),
+      selectQuery: node.selectQuery,
```
koskimas, line 414: You need transform the the `selectQuery`

viraxslot: Thanks, fixed!

koskimas (approved):
Seems perfect! Thanks a lot!

PR #793, opened 2023-12-04
fix: `$if` inside `with` is `{ [x: string]: any }[]`.

### LINKED ISSUE #785: WITH queries lose typesafety when using $if after select
Don't know if this a bug or expected behavior: https://kyse.link/?p=s&i=YkGTBDBQLn5RdD2wss6P
If I swap the order of the $if and select statements, it works.

igalklebanov:
closes #785.

This is voodoo, but works. Nothing else I tried did - the result of the inner `$if` was always `{ [x: string]: any }[]`.


koskimas (approved):
Awesome! Checked this and it didn't slow things down.

PR #796, opened 2023-12-04
fix: query and error logging both occur on error.

### LINKED ISSUE #743: Different query duration in error and query events
Hi 👋 

I'm logging queries using `log` property and when there's an error I can see two logs, first for `error` event, second for `query` event.

They have different `queryDurationMillis`. Does it mean the query is executed twice or what's the reasoning for that?

Example:

```ts
import { CamelCasePlugin, Kysely, PostgresDialect } from 'kysely'
import pg from 'pg'
import environment from './modules/environment.js'
import type { Database } from './types/database.d.js'

const database = new Kysely<Database>({
    pool: new pg.Pool({
      database: environment.DATABASE_NAME,
      host: environment.DATABASE_HOST,
      password: environment.DATABASE_PASSWORD,
      port: environment.DATABASE_PORT,
      user: environment.DATABASE_USER
    })
  }),
  log: (event) => {
    console.log(event.level)
    console.log(event.query.sql)
    console.log(event.queryDurationMillis)
  },
  plugins: [
    new CamelCasePlugin()
  ]
})

try {
  await database.selectFrom('brands').where('nonExistentColumn', '=', 1).execute()
} catch (error) {
  // do nothing
} finally {
  await database.destroy()
}
```

Output:

image

kysely - 0.26.3
pg - 8.11.3
node - 20.8.0

igalklebanov:
closes #743.

This is better than the suggested solution in the comment section. If we moved query logging to the `try` body, we'd catch errors in user's logging logic. User has all the context necessary to handle his logic's exceptions and print necessary errors.

Regarding timing mismatch, not much we can do about it.

Went ahead and fixed `defaultLogger`'s error handling when not `Error` - it was printing `[object Object]`.

### DIFF
--- a/src/util/log.ts
+++ b/src/util/log.ts
@@ -72,7 +72,13 @@ function defaultLogger(event: LogEvent): void {
     if (event.error instanceof Error) {
       console.error(`kysely:error: ${event.error.stack ?? event.error.message}`)
     } else {
-      console.error(`kysely:error: ${event}`)
+      console.error(
+        `kysely:error: ${JSON.stringify({
+          error: event.error,
+          query: event.query.sql,
+          queryDurationMillis: event.queryDurationMillis,
+        })}`
+      )
     }
   }
 }
--- a/src/driver/runtime-driver.ts
+++ b/src/driver/runtime-driver.ts
@@ -117,8 +117,8 @@ export class RuntimeDriver implements Driver {
     connection.executeQuery = async (
       compiledQuery
     ): Promise<QueryResult<any>> => {
-      const startTime = performanceNow()
       let caughtError: unknown
+      const startTime = performanceNow()
 
       try {
         return await executeQuery.call(connection, compiledQuery)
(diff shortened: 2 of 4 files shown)


src/util/log.ts
```diff
@@ -72,7 +72,13 @@ function defaultLogger(event: LogEvent): void {
     if (event.error instanceof Error) {
       console.error(`kysely:error: ${event.error.stack ?? event.error.message}`)
     } else {
-      console.error(`kysely:error: ${event}`)
+      console.error(
+        `kysely:error: ${JSON.stringify({
+          error: event.error,
+          query: event.query.sql,
```
koskimas, line 78: Should also log the parameters

koskimas (approved):
LGTM

Awesome tests as usual. In fact maybe too awesome 😄 We should try to write the minimum amount of tests that still do the job "well enough". We also need to maintain the tests, and having a gazillion tests just slows down development.

PR #797, opened 2023-12-05
Update isPlainObject to exclude objects with constructors

sds:
Hey all, I'm not sure if this is the right solution to the issue we're observing, so would welcome feedback on a better way to solve.

If you are using a custom type parser (e.g. via `pg.types.setTypeParser`) that returns a custom class, the camel case plugin will incorrectly serialize that value to a literal object, stripping it of its prototype.

This is problematic if you rely on the ability to call certain methods specified by the custom class' prototype.

To date, `isPlainObject` is used only in the camel case and JSON parsing plugins. I'm not sure what the original intent of the `isPlainObject` implementation was, so let me know if we think this fix belongs in the camel case plugin.


koskimas:
I don't think this works in every case. You need to at least add a test that checks all the old cases we were detecting specifically. And how about cases where there's no constructor? Like `isPlainObject(Object.create(null))`?

sds:
Thanks for the feedback! Would you want a separate test file for `isPlainObject`, or tests added to the existing [camel case plugin tests](https://github.com/kysely-org/kysely/blob/3d5e0e59d5536e2dbfee8e39ebecd0b5d981536e/test/node/src/camel-case.test.ts)?

In the meantime, I pushed an updated implementation that handles the `Object.create(null)` case.

igalklebanov:
@sds let's add a separate test file, and verify there is no regression.

PR #806, opened 2023-12-16
add dialect specific truncate snippet for sqlite

collinstevens:
sqlite doesn't support a truncate statement, but instead supports truncate optimization on a where-less delete statement

https://www.sqlite.org/lang_delete.html
https://www.techonthenet.com/sqlite/truncate.php


igalklebanov (approved):
nice catch again! 🦅

PR #810, opened 2023-12-19
Correct import of pg in examples

EspadaV8:
The pg package doesn't support ESM so the current imports are incorrect - see <https://github.com/brianc/node-postgres/issues/3060>.

This MR updates the docs so that the examples work after copy/pasting them.

### DIFF
--- a/example/src/app.ts
+++ b/example/src/app.ts
@@ -12,7 +12,7 @@ import { Router } from './router'
 import { userController } from './user/user.controller'
 import { ControllerError } from './util/errors'
 import { isObject } from './util/object'
-import { Pool } from 'pg'
+import pg from 'pg'
 
 export class App {
   #config: Config
@@ -27,7 +27,7 @@ export class App {
     this.#router = new Router()
     this.#db = new Kysely<Database>({
       dialect: new PostgresDialect({
-        pool: async () => new Pool(this.#config.database),
+        pool: async () => new pg.Pool(this.#config.database),
       }),
     })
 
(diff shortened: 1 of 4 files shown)


koskimas:
Seems to work in our tests https://github.com/kysely-org/kysely/blob/680eb46d257f58d437ad8f6334255b48d1e9722d/test/node/src/test-setup.ts#L5C1-L5C38

EspadaV8:
> Seems to work in our tests https://github.com/kysely-org/kysely/blob/680eb46d257f58d437ad8f6334255b48d1e9722d/test/node/src/test-setup.ts#L5C1-L5C38
> 
> Just needs `"moduleResolution": "Node"` in tsconfig I think.

Hmm, I had `bundler` in there (set by create-svelte). I've just changed it to `Node` but get the same error

```
import { Pool } from 'pg';
^

SyntaxError: The requested module 'pg' does not provide an export named 'Pool'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:132:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:214:5)
    at async ModuleLoader.import (node:internal/modules/esm/loader:329:24)
    at async loadESM (node:internal/process/esm_loader:34:7)
    at async handleMainPromise (node:internal/modules/run_main:113:12)
```

madflow:
> Just needs "moduleResolution": "Node" in tsconfig I think.

So it  works in Typescript. Maybe the docs should be amended, that these are Typescript examples?

PR #821, opened 2023-12-29
add `TOP` clause support.

igalklebanov:
Hey :wave:

`top` clause is supported in dialects such a MS SQL Server (MSSQL) and Snowflake.

It works just like `limit` other dialects, but has some interesting modifiers like `percent`.

Its the easiest way to limit rows in a `select` query, given that MSSQL doesn't have a `limit` clause, and requires using `order by ... fetch <n> rows`.
Its also supported in `insert ... expression`/`update`/`delete`/`merge` queries.


src/query-compiler/default-query-compiler.ts
```diff
@@ -1509,6 +1531,14 @@ export class DefaultQueryCompiler
     }
   }
 
+  protected override visitTop(node: TopNode): void {
+    this.append(`top(${node.expression})`)
```
koskimas, line 1535: There's an injection vulnerability for both `expression` and `modifiers`. I could imagine client code that takes the expression and/or modifiers straight from user input.

Add a "parser" that checks that `expression` and `modifier` are in fact a number and the literal `percent` before creating the node.

koskimas (approved):
Check `expression` and `modifier` to prevent SQL injection. Then this is good to go. Great job!

src/query-compiler/default-query-compiler.ts
igalklebanov: Great catch!

done.

igalklebanov:
Changes since last review:

- added the parser to protect from SQL injection.
- added support for top clause in merge queries.

.prettierrc.json
```diff
@@ -1,4 +1,5 @@
 {
   "semi": false,
-  "singleQuote": true
-}
\ No newline at end of file
+  "singleQuote": true,
+  "trailingComma": "all"
```
igalklebanov, line 4: prettier started making 982136491826341298632189 changes removing these everywhere.

PR #822, opened 2023-12-29
add `FETCH` clause support.

igalklebanov:
Hey 👋 

`fetch` clause is supported in dialects such as [PostgreSQL](https://www.postgresql.org/docs/current/sql-select.html), [MS SQL Server (MSSQL)](https://learn.microsoft.com/en-us/sql/t-sql/queries/select-order-by-clause-transact-sql?view=sql-server-ver16) and [OracleDB](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/SELECT.html).

It works just like `limit`, but sometimes `limit` is not supported, so here we are.

It is only supported in `select` queries, or when they're used as subqueries in top-level `insert`/`delete`/`update` queries.

This PR also:
- expands `limit` and `offset` by allowing `bigint`s.
- fixes mssql `offset` compilation.


src/query-compiler/default-query-compiler.ts
```diff
@@ -1509,6 +1515,12 @@ export class DefaultQueryCompiler
     }
   }
 
+  protected override visitFetch(node: FetchNode): void {
+    this.append('fetch next ')
+    this.visitNode(node.rowCount)
+    this.append(` rows ${node.modifier}`)
```
koskimas, line 1521: SQL injection with modifier. Not very serious one. We might have these for modifiers in the existing code too. But we should start checking the values for everything that gets interpolated to the SQL.

koskimas (approved):

src/query-compiler/default-query-compiler.ts
igalklebanov: Great catch!

PR #824, opened 2023-12-31
Fix regression Node 14

korichdaniel:
After bumping to version 0.27.0, support for node 14 is missed.


igalklebanov (approved):
Hey 👋 

Thanks! 🙏

PR #826, opened 2023-12-31
feat: add schema to `TableNode`s within `RawNode`s @ `WithSchemaPlugin`.

### LINKED ISSUE #761: Raw sql execute does not use schema during migration
Hi,

We are currently trying to have a custom migration that runs everytime `reagardless` the current migration step. This migration will hold the last version of some `stored procedure`. These `stored procedure` must run using the `latest` `constant` from our codebase.

We work with schema for testing purpose and we are doing the following 

```typescript
export async function up(db: Kysely<MyDatabase>): Promise<void> {
await sql`
      CREATE OR REPLACE procedure recompute_ranges()
      LANGUAGE plpgsql
      as $$
      declare
        _current record;
      begin
          select * into _current from ${sql.table('something_entry')} where "id" = b_current_entry_id;
      end;
      $$;`.execute(db)          
}
```

The compiled query does not contains any information about the schema provided by using: 
```typescript
const db = new Kysely<MyDatabase>({
      pool: new Pool(...)
    }),
    plugins: [new WithSchemaPlugin(schemaName)],
  });
```

Is there any way to use the right schema ?

Thanks

igalklebanov:
closes #761.

When using `sql.table`, `eb.table`, `sql.ref` and `eb.ref`, we create `TableNode`s. 

When executing `sql` template tag queries or parts of queries, any plugins on the db instance are run - including `WithSchemaPlugin` when available.

Currently, the plugin doesn't handle `RawNode`s, so we only get partial schema addition.

This PR addresses that. Now using `.ref` and `.table` is even more powerful. 👓

### DIFF
--- a/test/node/src/with-schema.test.ts
+++ b/test/node/src/with-schema.test.ts
@@ -1,3 +1,5 @@
+import { Kysely, Sql, sql } from '../../..'
+import { createExpressionBuilder } from '../../../dist/cjs/expression/expression-builder'
 import {
   clearDatabase,
   destroyTest,
@@ -9,6 +11,7 @@ import {
   DIALECTS,
   insert,
   limit,
+  expect,
 } from './test-setup.js'
 
 for (const dialect of DIALECTS.filter(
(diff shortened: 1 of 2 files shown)


test/node/src/with-schema.test.ts
```diff
@@ -517,6 +520,72 @@ for (const dialect of DIALECTS.filter(
       })
     })
 
+    describe('raw sql', () => {
```
koskimas, line 523: I think a little copy-pasta is better than these DRY tests that are really really hard to read. This is fine, but in the future, if there's only a couple of cases, just copy paste them.

test/node/src/with-schema.test.ts
```diff
@@ -517,6 +520,72 @@ for (const dialect of DIALECTS.filter(
...
+          )
+
+        expect(query.sql).to.equal(
+          '"pet" "mammals"."pet" "id" "pet"."id" "mammals"."pet"."id" pet mammals.pet id pet.id mammals.pet.id'
+        )
+      })
+
+      subsuites.forEach(([name, builder]) => {
+        describe(`${name}.table`, () => {
+          it('should add schema', () => {
```
koskimas, line 550: Add a test case like this:

```ts
sql`
  select * from person as p
  inner join pets on pets.owner_id = ${ref.ref('p.id')}
`
```

I think the current code would incorrectly add schema to `mammals.p.id` 🤔

It might be impossible to handle this correctly since we don't know how `ref` and `table` are used in raw SQL.

koskimas (changes requested):

test/node/src/with-schema.test.ts
igalklebanov: Absolutely agree, it's a mess.

test/node/src/with-schema.test.ts
igalklebanov: Damn, got some rethinking to do. Great catch! 💪

PR #828, opened 2024-01-01
add `OUTPUT` clause support.

igalklebanov:
Hey 👋 

`output` clause is supported in dialects such as [MS SQL Server (MSSQL)](https://learn.microsoft.com/en-us/sql/t-sql/queries/output-clause-transact-sql?view=sql-server-ver16).

It works just like `returning`, but:

- forces columns to be prefixed with `inserted` for new values, `deleted` for old values, and table names/aliases that are not the main table. Since we don't support mssql's way of adding more tables to context in all the queries output is supported in, I left it out for a future PR.
- for `merge` queries, `$action` returns `delete`, `insert` or `update`. left it out for a future PR.
- can be used to insert into a variable or another table with the `into` sub-clause. left it out for a future PR.


src/dialect/dialect-adapter.ts
```diff
@@ -31,6 +31,12 @@ export interface DialectAdapter {
    */
   readonly supportsReturning: boolean
 
+  /**
+   * Whether or not this dialect supports the `output` clause in inserts
+   * updates and deletes.
+   */
+  readonly supportsOutput: boolean
```
koskimas, line 38: Hmm.. I wonder if every 3rd party dialect correctly uses the `DialectAdapterBase`. If not, this is kind of a breaking change. Maybe this should be optional?

koskimas (approved):
Greatness! One thing to consider about the `supportsOutput` property. We should probably play it safe and make it optiona.

src/dialect/dialect-adapter.ts
igalklebanov: Good idea! it is optional now.

khangal:
👏👏👏👏👏👏

tmrclark:
When will this be getting released?

magnusrodseth:
@igalklebanov Is this on its way to be published? I cannot find docs on how to use it.

PR #842, opened 2024-01-11
introspect column comments 

### LINKED ISSUE #841: Introspect should also fetch column comments so we can generate typescript comments
Related to https://github.com/RobinBlomberg/kysely-codegen/issues/127

We need comments in the introspection, so we can generate typescript comments for the column. Would you accept a PR for this?

maktouch:
closes #841

### DIFF
--- a/src/dialect/database-introspector.ts
+++ b/src/dialect/database-introspector.ts
@@ -68,4 +68,5 @@ export interface ColumnMetadata {
   readonly isAutoIncrementing: boolean
   readonly isNullable: boolean
   readonly hasDefaultValue: boolean
+  readonly comment?: string
 }
--- a/src/dialect/postgres/postgres-introspector.ts
+++ b/src/dialect/postgres/postgres-introspector.ts
@@ -57,7 +57,7 @@ export class PostgresIntrospector implements DatabaseIntrospector {
         'ns.nspname as schema',
         'typ.typname as type',
         'dtns.nspname as type_schema',
-
+        sql<string | null>`col_description(a.attrelid, a.attnum)`.as('column_description'),
         // Detect if the column is auto incrementing by finding the sequence
         // that is created for `serial` and `bigserial` columns.
         this.#db
@@ -126,6 +126,7 @@ export class PostgresIntrospector implements DatabaseIntrospector {
           isNullable: !it.not_null,
           isAutoIncrementing: !!it.auto_incrementing,
           hasDefaultValue: it.has_default,
+          comment: it.column_description ?? undefined
         })
       )
 
@@ -148,4 +149,5 @@ interface RawColumnMetadata {
   type: string
   type_schema: string
   auto_incrementing: boolean | null
+  column_description: string | null
 }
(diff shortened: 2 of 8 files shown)


koskimas:
How about SQLite and mssql? If they support comments, this feature needs to be implemented for those also.

Tests are missing.

test/node/src/introspect.test.ts
```diff
@@ -94,6 +94,7 @@ for (const dialect of DIALECTS) {
                   isNullable: false,
                   isAutoIncrementing: true,
                   hasDefaultValue: true,
+                  column: null
```
koskimas, line 97: Column?

src/dialect/database-introspector.ts
```diff
@@ -68,4 +68,5 @@ export interface ColumnMetadata {
   readonly isAutoIncrementing: boolean
   readonly isNullable: boolean
   readonly hasDefaultValue: boolean
+  readonly comment?: string | null
```
koskimas, line 71: Optional is enough. No need to make this also nullable. What would the difference between `undefined` and `null` value be?

src/dialect/postgres/postgres-introspector.ts
```diff
@@ -57,6 +57,7 @@ export class PostgresIntrospector implements DatabaseIntrospector {
         'ns.nspname as schema',
         'typ.typname as type',
         'dtns.nspname as type_schema',
+        sql`pg_catalog.col_description(format('%s.%s',isc.table_schema,isc.table_name)::regclass::oid,isc.ordinal_position)`.as('column_description'),
```
koskimas, line 60: I don't think this works. At all. You're referencing tables that don't exist in this query.

maktouch:
> Thank you!
> 
> How about SQLite and mssql? If they support comments, this feature needs to be implemented for those also.
> 
> Tests are also needed for every feature. Add at least one test that adds a comment and checks that it gets returned by the introspector. Currently the tests and the whole feature are completely broken.

Haha it's why I made this field optional. I wasn't familiar enough with them to implement it 😂

Ok I'll get familiar, implement the feature, make it non optional but billable, and do the tests.

koskimas:
> Haha it's why I made this field optional. I wasn't familiar enough with them to implement it 

It should still be optional even if all built-in dialects support it. Not all databases do, and we want 3rd party dialects to be able to implement the API.

maktouch:
> > Haha it's why I made this field optional. I wasn't familiar enough with them to implement it
> 
> It should still be optional even if all built-in dialects support it. Not all databases do, and we want 3rd party dialects to be able to implement the API.

ok that makes sense, so I'll make the value be undefined if it's empty

maktouch:
I've reverted my changes and started from scratch in TDD mode
- The first thing I did was modify `test/node/src/test-setup.ts` to add a comment somewhere for each dialect. Unfortunately, I wasn't able to reliably add a comment for sqlite here, so sqlite always returns undefined for comment. If you can help me out here, that would be great! (or we can also just ignore it, it doesn't seem like comment is a real feature in sqlite)
- Then I added the expected comments in `test/node/src/schema.test.ts` and `test/node/src/introspect.test.ts`
- Implemented the features in `src/dialect/$DB/$DB-introspector.ts` until it passed. 

Let me know what else I can change! Thanks! :)

maktouch:
Hey @koskimas

Please check this PR when you have time 🙏🙏

Thank you

koskimas (approved):
Perfect! Thank you 👍

PR #844, opened 2024-01-15
#843 - Use JS sort method to sort migrations by name

### LINKED ISSUE #843: Postgres migrateUp sometimes failing if migration name has a "-" character
# Description
In versions of kysely  < 0.27.2, if migration names have a "-" character (beyond the usual ISO date string), running `migrateToLatest`  against a postgres database might fail even if it had previously succeeded. 
The postgres `ORDER BY` ignores "-" character when ordering strings (while javascript sort method considers it when sorting strings) causing the order of executed migrations returned from database and the order of candidate migrations to not be the same.

Since 0.27.2 this issue is less likely to occur because when
querying for executed migrations the `ORDER BY` uses the `timestamp` and secondarily the `name` column to order executed migrations (introduced in #723). 

But it can still occurr if the system where the migrations are being
runned is extremly fast or the system clock is unreliable causing the timestamps to be the same. 
That would result in the ordering of migrations being the same as if only the `name` column were used 
in the  `ORDER BY` clause.

# Example. 
We have the following pending migrations:
- 2024-01-01-create-table
- 2024-01-01.2-update-table

After running `migrator.migrateToLatest()` We get the following values in the table (the timestamps are the same)

|name                       | Timestamp                 |
|---------------------------|---------------------------|
|2024-01-01-create-table    | 2024-01-01T00:00:00.000Z  |
|2024-01-01.2-update-table  | 2024-01-01T00:00:00.000Z  |

When `migratetoLatest()` is called again  without `allowUnorderedMigrations` set to true
it will fail in version `0.27.2` and prior versions because the index of the candidate 
migrations and the executed migrations won't match

| index | candiate Migrations        | executed_migrations (as read from db)   |
--------|----------------------------|-----------------------------------------|
| 0     |  2024-01-01-create-table   | 2024-01-01.2-update-table               |
| 1     |  2024-01-01.2-update-table | 2024-01-01-create-table                 |

DavesBorges:
Simple fix to #843. Now javascript sort function is used to both sort the executed migrations and the migrations to be executed. 

Closes #843


koskimas:
This now breaks the sorting by timestamp. Out-of-order migrations can be executed in different order than the name order. That's why the primary ordering is by timestamp.

DavesBorges:
Then perhaps choose the sort method based on allowUnordered migrations?

src/migration/migrator.ts
```diff
@@ -513,7 +513,7 @@ export class Migrator {
       .orderBy(['timestamp', 'name'])
       .execute()
 
-    return executedMigrations.map((it) => it.name)
+    return executedMigrations.map((it) => it.name).sort()
   }
```
koskimas, line 518: Use the same sorting: first by timestamp, and only then by name. This is important for unordered migrations.

koskimas: The timestamp doesn't come from the file name. It's the execution time of the migration.

koskimas:
> Then perhaps choose the sort method based on allowUnordered migrations?

All you need to do is sort using the same logic as the SQL. Sort by timestamp and if the timestamps are equal, sort by name.

DavesBorges:
@koskimas  I will add more tests for the out of order migration because the current tests didn't break when I ordered  the executed migrations only by name

DavesBorges:
@koskimas I've found out why the tests haven't broken with `allowUnordereMigrations` enabled. The reason was the test migrations name were already in alphabetical order so I've added a test case where the migrations name are not in order to test that when migrating down it determines the migration order using the timestamps.

src/migration/migrator.ts
```diff
@@ -509,11 +509,17 @@ export class Migrator {
...
-      .select('name')
-      .orderBy(['timestamp', 'name'])
+      .select(['name', 'timestamp'])
       .execute()
 
-    return executedMigrations.map((it) => it.name)
+    return executedMigrations
+      .sort((a, b) => {
+        if (a.timestamp === b.timestamp) {
+          return a.name < b.name ? -1 : 1
```
igalklebanov, line 518: change to [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare).

test/node/src/migration.test.ts
```diff
@@ -792,6 +844,48 @@ for (const dialect of DIALECTS) {
...
+        before(() => {
+          originalToIsoString = Date.prototype.toISOString
+          const defaultDateIsoString = new Date(2024, 0, 11).toISOString()
+          Date.prototype.toISOString = () => defaultDateIsoString
+        })
+
+        after(() => {
+          // Reset to originalToIsoString function so that upcoming tests are not affected
+          Date.prototype.toISOString = originalToIsoString
+        })
```
igalklebanov, line 863: This is not clean, too coupled with implementation detail.
Use [sinon fake timers](https://sinonjs.org/releases/latest/fake-timers/) instead.

igalklebanov: Doesn't work for some reason. 🤷

src/migration/migrator.ts
igalklebanov: Exposed this for overriding as well.

PR #862, opened 2024-01-29
Implement `JOIN ... USING` query

### LINKED ISSUE #730: Support `JOIN ... USING ( col1, col2, ... )`

dwickern:
Fixes #730


dwickern:
To implement `JOIN ... USING`, I added a new type argument to `JoinBuilder`, `C extends string`, representing the column names which are common to both tables being joined.

I see that `JoinBuilder` is also used for the `MERGE` query. Consequently this PR allows for a `MERGE` syntax which is not valid in Postgres and probably not valid on any DBMS:
```ts
// merge into person using pet using (id)
db.mergeInto('person').using('pet', (join) => join.using(['id']))
```

There are a couple ways to solve that (if maintainers here think it's worth solving):
1. Create a `SelectJoinBuilder` which extends `JoinBuilder` and adds the `using` function. Use `SelectJoinBuilder` for `SELECT` queries and use `JoinBuilder` for `MERGE` queries
2. `MERGE` queries could assign `C = never` as the type param to `JoinBuilder`

src/query-builder/join-builder.ts
```diff
@@ -79,6 +79,37 @@ export class JoinBuilder<DB, TB extends keyof DB>
...
+   *
+   * The generated SQL (PostgreSQL):
+   *
+   * ```sql
+   * select *
+   * from "person"
+   * inner join "pet" using ("name")
+   * ```
+   */
+  using(columns: [C, ...C[]]): JoinBuilder<DB, TB, C> {
```
dwickern, line 106: Does it make sense to have an overload for a single column? I noticed that other APIs like `select` have this.
```js
.innerJoin('pet', join => join.using('col'))
.innerJoin('pet', join => join.using(['col1', 'col2', 'col3']))
```

src/operation-node/join-node.ts
```diff
@@ -15,7 +16,7 @@ export interface JoinNode extends OperationNode {
   readonly kind: 'JoinNode'
   readonly joinType: JoinType
   readonly table: OperationNode
-  readonly on?: OnNode
+  readonly on?: OnNode | JoinUsingNode
```
igalklebanov, line 19: That's a breaking change not worth doing, and also a surprise - how's `on` related to `using`?

This should make things quite simple, and no need for an extra node type.

```suggestion
  readonly on?: OnNode
  readonly using?: ReadonlyArray<ColumnNode>
```

src/query-builder/join-builder.ts
igalklebanov: Seems OK to me.
I'm not sure about `[C, ...C[]]`, I don't think we're THAT strict in other places.

igalklebanov:
> To implement `JOIN ... USING`, I added a new type argument to `JoinBuilder`, `C extends string`, representing the column names which are common to both tables being joined.

Sounds like it's not crucial and could've been placed on the method/s. We're usually quite conservative when it comes to adding more generic parameters to existing builders. There has to be a good reason..

> I see that `JoinBuilder` is also used for the `MERGE` query. Consequently this PR allows for a `MERGE` syntax which is not valid in Postgres and probably not valid on any DBMS:
> 
> ```ts
> // merge into person using pet using (id)
> db.mergeInto('person').using('pet', (join) => join.using(['id']))
> ```
> 
> There are a couple ways to solve that (if maintainers here think it's worth solving):
> 
> 1. Create a `SelectJoinBuilder` which extends `JoinBuilder` and adds the `using` function. Use `SelectJoinBuilder` for `SELECT` queries and use `JoinBuilder` for `MERGE` queries
> 2. `MERGE` queries could assign `C = never` as the type param to `JoinBuilder`

That's fine for now, not worth the extra builder/s IMHO.

peterHakio:
@dwickern Hope you have time to finish the pull request and insure that it gets merged. Since it is a very nice feature and very nice work.

PR #883, opened 2024-02-14
[Docs] Update complex join example to better illustrate the options available (symmetry with `where`)

MarkusWendorf:
Hello,

when I read the docs I failed to understand how to do OR expressions.

> You can do all the same things with the
`on` method that you can with the corresponding `where` method

The hint at the `where` documentation was not sufficient for me (that's more a me issue though).

So I included a little more information about the options provided by `on`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,6 +24,7 @@
         "chai-as-promised": "^7.1.1",
         "chai-subset": "^1.6.0",
         "esbuild": "^0.19.11",
+        "lodash": "^4.17.21",
         "mocha": "^10.2.0",
         "mysql2": "^3.6.5",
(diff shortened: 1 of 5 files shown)


koskimas:
Those files are automatically generated from the API docs. You need to:

1. modify the example in the code 
2. Build using `npm run build`
3. Run [this script](https://github.com/kysely-org/kysely/blob/master/scripts/generate-site-examples.js) to generate the example files for kysely.dev.

MarkusWendorf:
Ok got it. The script is using `lodash` which was not present in the package.json. I added it as a dev dependency, if that's alright.

koskimas (approved):
Thanks!

PR #887, opened 2024-02-24
Expose AllSelection

lorefnon:
Hello, thanks for maintaining Kysely. 

Would you mind exposing the AllSelection type from select-parser ? It is useful for consumers too.


igalklebanov:
Hey 👋 

> It is useful for consumers too.

How is it useful?

We're generally cautious in exposing internals as it may make changing things harder in the future.

PR #892, opened 2024-02-29
Add support for json_agg_strict (PostgreSQL only)

leibale:
Add support for json_agg_strict (PostgreSQL only) - this function is just like json_agg, but skips null (see https://www.postgresql.org/docs/current/functions-aggregate.html)

A few notes:
1. I've duplicated all json_agg tests and changed them to use json_agg_strict, but I'm not sure if:
    1. Should we test all cases for both json_agg and json_agg_strict?
    2. If we want to fully test both, do you prefer duplicating the code, using a loop, or using a function?
2. json_agg might return null (for example, when using left join and there is nothing to join with), but the types ignore that.. I guess fixing it is a breaking change, so I did not include it in this PR

Thanks for this awesome package! :)


leibale:
When using it with left join every field becomes nullable (I think that left join changes `DB[table]` to a "nullable version"), which is not true with json_agg_strict (which unlike json_agg will reply with `[]` for "empty join").
I'm not too sure if this will be easy to solve.. any help will be appreciated :)

igalklebanov:
Hey 👋 

Thanks! 🙏 

I'm on the fence on this one for now:

- This only collected 2 reactions in almost a year.
- It is supported just in PostgreSQL.
- It can be achieved in other ways:

```ts
const results = await db
  .selectFrom("person")
  .leftJoin("pet", "pet.owner_id", "person.id")
  .groupBy("person.first_name")
  .select((eb) => [
    "first_name",
    eb.fn
      .coalesce( // <-- ensures defaults to empty array instead of null
        eb.fn
          .jsonAgg("pet")
          .filterWhere("pet.owner_id", "is not", null), // <-- removes null rows
        sql.lit("[]"),
      )
      .as("pets"),
  ])
  .$narrowType<{ pets: Selectable<PetTable>[] }>() // <-- this makes all pet fields not nullable in a type-safe way.
  .execute();
```
https://kyse.link/ZIU7y

Issue #780, opened 2023-11-24
Add `order by` support in aggregate functions argument scope.


igalklebanov:
Hey 👋 

Following #779, we might want to allow chaining `.orderBy` where `.distinct` is invocable in `eb.fn.agg`.

Example aggregate functions where `order by` can be used inside the argument scope:

- MySQL's [`group_concat`](https://dev.mysql.com/doc/refman/8.0/en/aggregate-functions.html#function_group-concat).
- PostgreSQL's [`array_agg`](https://www.postgresql.org/docs/9.5/sql-expressions.html#SYNTAX-AGGREGATES).
- SQLite's [`group_concat`](https://www.sqlite.org/lang_aggfunc.html#group_concat).
- Oracle's [`collect`](https://docs.oracle.com/en/database/oracle/oracle-database/19/nlspg/appendix-C-collation-derivation-and-determination-rules-for-SQL-operations.html#GUID-631539B8-1F31-4ADC-9924-4D2B57E031C3).

PR #896, opened 2024-03-07
Add orderBy clause support to aggregate function builder

ivashog:
#780

see sql syntax on PostgreSQL example: https://www.postgresql.org/docs/16/sql-expressions.html#SYNTAX-AGGREGATES


igalklebanov:
Hey 👋 

Thanks! 💪 

This is also support in [MySQL](https://dev.mysql.com/doc/refman/8.3/en/aggregate-functions.html#function_group-concat) and [SQLite](https://www.sqlite.org/lang_aggfunc.html) for some aggregate functions. Let's add tests for those.

igalklebanov:
> But it is better to implement it in a separate PR, what do you say?

Yeah, different clause, different PR.

test/node/src/aggregate-function.test.ts
```diff
@@ -1106,6 +1107,56 @@ for (const dialect of DIALECTS) {
 
       await query.execute()
     })
+
+    describe(`should execute order-sensitive aggregate functions`, () => {
+      // @todo add test for mssql when implement `WITHIN GROUP ( order_by_clause )` clause
```
igalklebanov, line 1112:
```suggestion
```

test/node/src/aggregate-function.test.ts
```diff
@@ -1106,6 +1107,56 @@ for (const dialect of DIALECTS) {
...
+      if (dialect !== 'mssql') {
```
igalklebanov, line 1113: Now that we have many dialects, it's best to be explicit.

```suggestion
      if (dialect === 'postgres' || dialect === 'mysql' || dialect === 'sqlite') {
```

test/node/src/aggregate-function.test.ts
```diff
@@ -1106,6 +1107,56 @@ for (const dialect of DIALECTS) {
...
+        const isMySql = dialect === 'mysql'
+        const funcName = isMySql ? 'group_concat' : 'string_agg'
+        const funcArgs: Array<ReferenceExpression<Database, 'person'>> = [
+          'first_name',
+        ]
+        if (!isMySql) funcArgs.push(sql.lit(','))
```
igalklebanov, line 1119: nitpicky

```suggestion
        if (!isMySql) {
          funcArgs.push(sql.lit(','))
        }
```

test/node/src/aggregate-function.test.ts
```diff
@@ -1106,6 +1107,56 @@ for (const dialect of DIALECTS) {
...
+            postgres: {
+              sql: [
+                `select ${funcName}("first_name", ',' order by "first_name" desc) as "first_names"`,
+                `from "person"`,
+              ],
+              parameters: [],
+            },
+            mysql: {
+              sql: [
+                `select group_concat(\`first_name\` order by \`first_name\` desc) as \`first_names\``,
```
igalklebanov, line 1141:
```suggestion
                `select ${funcName}(\`first_name\` order by \`first_name\` desc) as \`first_names\``,
```

test/node/src/aggregate-function.test.ts
```diff
@@ -1106,6 +1107,56 @@ for (const dialect of DIALECTS) {
...
+                `from \`person\``,
+              ],
+              parameters: [],
+            },
+            mssql: NOT_SUPPORTED,
+            sqlite: {
+              sql: [
+                `select string_agg("first_name", ',' order by "first_name" desc) as "first_names"`,
```
igalklebanov, line 1149:
```suggestion
                `select ${funcName}("first_name", ',' order by "first_name" desc) as "first_names"`,
```

src/query-builder/aggregate-function-builder.ts
```diff
@@ -95,6 +103,43 @@ export class AggregateFunctionBuilder<DB, TB extends keyof DB, O = unknown>
     })
   }
 
+  /**
+   * Adds an `order by` clause inside the aggregate function for specify correct result order
+   * ([see postgres docs](https://www.postgresql.org/docs/9.5/sql-expressions.html#SYNTAX-AGGREGATES))
```
igalklebanov, line 108: This is more than enough. "Bring Your Own SQL Knowledge".

```suggestion
   * Adds an `order by` clause inside the aggregate function.
```

test/node/src/aggregate-function.test.ts
igalklebanov: We should open an issue for this, and not leave a comment in the code.

igalklebanov (commented):
Looks good and functional overall! 💪 

Left a few minor suggestions, once addressed this is good to go! 🚀

ivashog:
> Looks good and functional overall! 💪
> 
> Left a few minor suggestions, once addressed this is good to go! 🚀

@igalklebanov, thanks for review!

test/typings/test-d/aggregate-function.test-d.ts
```diff
@@ -378,6 +378,23 @@ async function testSelectWithDistinct(db: Kysely<Database>) {
...
+        .agg<string[]>('array_agg', ['pet.name'])
+        .orderBy('pet.name')
+        .as('person_pet_names'),
+      eb.fn.jsonAgg('pet').orderBy('pet.id').as('person_pets'),
+    ])
+    .executeTakeFirstOrThrow()
+
+  expectAssignable<string[]>(result.person_pet_names)
+  expectAssignable<Selectable<Pet>[]>(result.person_pets)
+}
```
igalklebanov, line 396: I don't think this adds any value. Let's remove this.

igalklebanov (approved):
LGTM! 🚀

test/typings/test-d/aggregate-function.test-d.ts
ivashog: I agree

juusaw:
Any plans to release this feature? It would be really useful for us.

Issue #910, opened 2024-03-13
Unable to create timetz with precision columns


dunkelbraun:
When creating a `timetz(p)` column with the schema builder:

```ts
await db.schema
  .createTable("test")
  .addColumn("t", "timetz(4)")
  .execute();
```

`parseDataTypeExpression`throws an Error:  

> Error: invalid column data type "timetz(4)"
 ❯ Module.parseDataTypeExpression node_modules/kysely/dist/esm/parser/data-type-parser.js:14:11

koskimas:
```ts
.addColumn("t", sql`timetz(4)`)
```

koskimas:
Oh, that's actually a supported type. Needs fixing.

dunkelbraun:
I forgot to mention that it was for the PostgreSQL dialect.

PR #911, opened 2024-03-13
Fix typo on timetz regex

dunkelbraun:
Fixes https://github.com/kysely-org/kysely/issues/910


koskimas:
`timetz` is also a valid type

koskimas (approved):
Thanks!

PR #925, opened 2024-03-29
feat: add HandleEmtpyInListsPlugin.

### LINKED ISSUE #709: "IN" on empty array causes invalid SQL
Not sure if it's by design or a bug but...

Trying to run `db.selectFrom('person').selectAll().where('id', 'in', [])` causes an invalid SQL (see image below👇 the `()` part is invalid). 

Now obviously, this query doesn't make a lot of sense, and you can simply assume there would be 0 rows returned, but IMO ot. makes sense when in the TypeScript context, and also SQL generating library that creates an invalid SQL is a bit weird. The ways I think we can overcome this "issue" really depends on the philosophy Kysely is based on:

- Leave it as is (Although I would suggest updating the docs to explain this non-trivial case)
- Throwing a runtime error
- replacing it with some sort of false statement (like `IN (null)` which should be equivalent)

Let me know what you guys think...

image

austinwoon:
Hey 👋 

closes #709.

This PR was created initially by @austinwoon, and revisited/expanded by myself.

It adds `HandleEmptyInListsPlugin` to allow you to handle `in ()` and `not in ()` with a chosen strategy.

It ships with 2 strategies out-the-box:

- `replaceWithNoncontingentExpression` - Replaces `in ()` with `1 = 0` and `not in ()` with `1 = 1` - the strategy used by Knex.js, PrismaORM, Laravel and SQLAlchemy.
- `pushValueIntoList(uniqueValue: string)` - Replaces `in ()` with `in (null)` - the strategy used by TypeORM and Sequelize, and `not in ()` with `cast({{leftOperand}} as char) in ('{{uniqueValue}}')`.

It also supports custom strategies, e.g.:

- throw an error so a request is not sent to the database.
- print a warning and apply one of the built-in strategies.
- mix the built-in strategies, e.g. `in (null)` and `1 = 1`.


src/plugin/safe-empty-array-where-in/safe-empty-array-where-in.transformer.ts
```diff
@@ -0,0 +1,64 @@
...
+
+interface InOperatorNode {
+  kind: 'OperatorNode'
+  operator: (typeof IN_OPERATORS)[number]
+}
+
+export class WithSafeArrayWhereInTransformer extends OperationNodeTransformer {
+  #BinaryOperationNode(node: OperationNode): node is BinaryOperationNode {
+    return node.kind === 'BinaryOperationNode'
+  }
```
igalklebanov, line 17: you get this "for free" using `BinaryOperationNode.is(node)`.

src/plugin/safe-empty-array-where-in/safe-empty-array-where-in.transformer.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  #isPrimitiveValueListNode(
+    node: OperationNode,
+  ): node is PrimitiveValueListNode {
+    return node.kind === 'PrimitiveValueListNode'
+  }
```
igalklebanov, line 23: you get this "for free" using `PrimitiveValueListNode.is(node)`.

src/plugin/safe-empty-array-where-in/safe-empty-array-where-in.transformer.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  #IsInOperatorNode(node: OperationNode): node is InOperatorNode {
+    return (
+      (node as InOperatorNode).kind === 'OperatorNode' &&
+      IN_OPERATORS.includes((node as InOperatorNode).operator)
+    )
+  }
+
+  protected transformWhere(node: WhereNode): WhereNode {
```
igalklebanov, line 32: what about `having`?

src/plugin/safe-empty-array-where-in/safe-empty-array-where-in.transformer.ts
```diff
@@ -0,0 +1,64 @@
...
+    const binaryOperatorNode = node.where
+
+    if (!this.#BinaryOperationNode(binaryOperatorNode)) {
+      return node
+    }
```
igalklebanov, line 37:
```suggestion
    if (!BinaryOperationNode.is(node.where)) {
      return node
    }
```

src/plugin/safe-empty-array-where-in/safe-empty-array-where-in.transformer.ts
```diff
@@ -0,0 +1,64 @@
...
+import { OperationNodeTransformer } from '../../operation-node/operation-node-transformer.js'
+import { PrimitiveValueListNode } from '../../operation-node/primitive-value-list-node.js'
+import { WhereNode } from '../../operation-node/where-node.js'
+
+const IN_OPERATORS = ['in', 'not in'] as const
+
+interface InOperatorNode {
+  kind: 'OperatorNode'
+  operator: (typeof IN_OPERATORS)[number]
+}
```
igalklebanov, line 12: no need for these.

src/plugin/safe-empty-array-where-in/safe-empty-array-where-in.transformer.ts
```diff
@@ -0,0 +1,64 @@
...
+  ): node is PrimitiveValueListNode {
+    return node.kind === 'PrimitiveValueListNode'
+  }
+
+  #IsInOperatorNode(node: OperationNode): node is InOperatorNode {
+    return (
+      (node as InOperatorNode).kind === 'OperatorNode' &&
+      IN_OPERATORS.includes((node as InOperatorNode).operator)
+    )
+  }
```
igalklebanov, line 30:
```suggestion
  #isIn(node: OperationNode): node is OperatorNode {
    return OperatorNode.is(node) && (node.operator === 'in' || node.operator === 'not in')
  }
```

src/plugin/safe-empty-array-where-in/safe-empty-array-where-in.transformer.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (!this.#BinaryOperationNode(binaryOperatorNode)) {
+      return node
+    }
+
+    const rightOperand = binaryOperatorNode.rightOperand
+    const operator = binaryOperatorNode.operator
+
+    if (
+      this.#IsInOperatorNode(operator) &&
+      this.#isPrimitiveValueListNode(rightOperand)
```
igalklebanov, line 44:
```suggestion
      PrimitiveValueListNode.is(rightOperand)
```

src/plugin/safe-empty-array-where-in/safe-empty-array-where-in.transformer.ts
```diff
@@ -0,0 +1,64 @@
...
+        return Object.freeze({
+          ...node,
+          where: {
+            ...binaryOperatorNode,
+            rightOperand: {
+              ...rightOperand,
+              values: [null],
+            },
+          },
+        })
```
igalklebanov, line 58:
```suggestion
        return Object.freeze({
          ...node,
          where: Object.freeze({
            ...binaryOperatorNode,
            rightOperand: Object.freeze({
              ...rightOperand,
              values: [null],
            }),
          }),
        })
```

src/plugin/safe-empty-array-where-in/safe-empty-array-where-in.transformer.ts
```diff
@@ -0,0 +1,64 @@
...
+      const values = rightOperand.values
+
+      if (Array.isArray(values) && values.length === 0) {
+        return Object.freeze({
+          ...node,
+          where: {
+            ...binaryOperatorNode,
+            rightOperand: {
+              ...rightOperand,
+              values: [null],
```
igalklebanov, line 55: what about situations where the value in db is `null`?

test/node/src/safe-empty-where-in.test.ts
```diff
@@ -0,0 +1,916 @@
...
+            firstName: 'John',
+          },
+          {
+            firstName: 'Mary',
+          },
+          {
+            firstName: 'Tom',
+          },
+        ])
+        .execute()
```
igalklebanov, line 60: why do we need a special table for this suite?

test/node/src/safe-empty-where-in.test.ts
```diff
@@ -0,0 +1,916 @@
...
+              `from "safeEmptyArrayPerson"`,
+              `where "firstName" not in (?)`,
+            ],
+            parameters: [null],
+          },
+        })
+
+        result = await notInQuery.execute()
+
+        expect(result).to.deep.equal([])
```
igalklebanov, line 151: should be in a separate test case.

test/node/src/safe-empty-where-in.test.ts
```diff
@@ -0,0 +1,916 @@
...
+              `from "safeEmptyArrayPerson"`,
+              `where "firstName" in (?)`,
+            ],
+            parameters: [null],
+          },
+        })
+
+        let result = await query.execute()
+
+        expect(result).to.deep.equal([])
```
igalklebanov, line 107: no need to execute really, testing the compiled sql is more than enough.

test/node/src/safe-empty-where-in.test.ts
```diff
@@ -0,0 +1,916 @@
...
+          },
+        })
+
+        let result = await query.execute()
+
+        expect(result).to.have.length(2)
+        expect(result).to.deep.equal([
+          { firstName: 'John' },
+          { firstName: 'Mary' },
+        ])
```
igalklebanov, line 201: no need to execute really, testing the compiled sql is more than enough.

test/node/src/safe-empty-where-in.test.ts
```diff
@@ -0,0 +1,916 @@
...
+              `where "firstName" not in (?, ?)`,
+            ],
+            parameters: ['John', 'Mary'],
+          },
+        })
+
+        result = await notInQuery.execute()
+
+        expect(result).to.have.length(1)
+        expect(result).to.deep.equal([{ firstName: 'Tom' }])
```
igalklebanov, line 246: make the first query have both in and not in to save time.

test/node/src/safe-empty-where-in.test.ts
```diff
@@ -0,0 +1,916 @@
...
+            firstName: 'Mary',
+          },
+          {
+            firstName: 'Tom',
+          },
+        ])
+        .execute()
+    })
+
+    describe('SELECT FROM WHERE IN TESTS', () => {
```
igalklebanov, line 63: no need for subsuite.

test/node/src/safe-empty-where-in.test.ts
```diff
@@ -0,0 +1,916 @@
...
+        })
+
+        result = await notInQuery.execute()
+
+        expect(result).to.have.length(1)
+        expect(result).to.deep.equal([{ firstName: 'Tom' }])
+      })
+    })
+
+    describe('DELETE FROM WHERE IN TESTS', () => {
```
igalklebanov, line 250: no need for subsuite.

test/node/src/safe-empty-where-in.test.ts
```diff
@@ -0,0 +1,916 @@
...
+              `delete from "safeEmptyArrayPerson"`,
+              `where "firstName" in (?)`,
+            ],
+            parameters: [null],
+          },
+        })
+
+        let result = await query.execute()
+
+        expect(result).to.deep.equal([new DeleteResult(BigInt(0))])
```
igalklebanov, line 289: no need to execute really, testing the compiled sql is more than enough.

test/node/src/safe-empty-where-in.test.ts
```diff
@@ -0,0 +1,916 @@
...
+              `delete from "safeEmptyArrayPerson"`,
+              `where "firstName" not in (?)`,
+            ],
+            parameters: [null],
+          },
+        })
+
+        result = await notInQuery.execute()
+
+        expect(result).to.deep.equal([new DeleteResult(BigInt(0))])
```
igalklebanov, line 328: make the first query have both `in` and `not in` to save time.

test/node/src/safe-empty-where-in.test.ts
```diff
@@ -0,0 +1,916 @@
...
+      })
+
+      it('should handle deleteFrom with returning in supported dialects', async () => {
```
igalklebanov, line 331: irrelevant. remove.

austinwoon: Will this use case never be used?

src/plugin/safe-empty-array-where-in/safe-empty-array-where-in.transformer.ts
austinwoon: in (null) should always return nothing, its the equivalent of `1=0`

test/node/src/safe-empty-where-in.test.ts
austinwoon: whats the convention in this project? im doing this just so to avoid collisions with other tests suites if things are run in parallel

test/node/src/safe-empty-where-in.test.ts
austinwoon: i think we should execute to make sure there are no run-time errors, which was the case before.

test/node/src/safe-empty-where-in.test.ts
igalklebanov: Testing every possible API combination for every plugin that ever alters the AST is pointless.

test/node/src/safe-empty-where-in.test.ts
austinwoon: this actually surfaced some bugs with this plugin. The AST for the case where we have `AND` or `OR` operators will not result in the transformation being applied. 

I could either:
- Apply the transformation on any `BinaryOperationNode`
- Repeat the transformation for WHERE, AND and OR nodes.

which should be the preferred option? Im afraid doing the transformation on any `BinaryOperationNode` would result in other side-effects

test/node/src/safe-empty-where-in.test.ts
austinwoon: Could we opt to keep to test for runtime errors

src/plugin/safe-empty-array-where-in/safe-empty-array-where-in.transformer.ts
austinwoon: Good point. With https://github.com/kysely-org/kysely/pull/925/commits/d0230ad19305046d8b06648dc13dc25e99368032 the transformation is now done when traversing a `BinaryOperationNode`, which covers this case.

Added tests [here](https://github.com/kysely-org/kysely/pull/925/commits/d0230ad19305046d8b06648dc13dc25e99368032)

test/node/src/safe-empty-where-in.test.ts
austinwoon: okay, will defer to your opinion on this. Deleted.

test/node/src/safe-empty-where-in.test.ts
igalklebanov: I think applying this plugin for all `in` usage is the right way to go, because `in ()` is a universal problem for some people, not just `where ... in`.

src/plugin/safe-empty-array-where-in/safe-empty-array-where-in.transformer.ts
igalklebanov: OK let's make sure this is correct across all 4 dialects:

1. have a record in data that has a `null` in that column.
2. run query against db and expect results are empty.

igalklebanov: Alternatively, could just alter the tree to `1=0` or equivalent and call it a day.

test/node/src/safe-empty-where-in.test.ts
austinwoon: Great, resolving :)

src/plugin/safe-empty-array-where-in/safe-empty-array-where-in.transformer.ts
austinwoon: https://github.com/kysely-org/kysely/pull/925/commits/015ee37730b5c2a0757d9985aabf3f0ed3e9f299 added tests here. Let me know if it is sufficient

austinwoon:
@igalklebanov ive addressed all of the PR comments, could i check if this is ready to be merged?

src/plugin/safe-empty-array-where-in/safe-empty-array-where-in.plugin.ts
```diff
@@ -0,0 +1,23 @@
...
+import { WithSafeArrayWhereInTransformer } from './safe-empty-array-where-in.transformer.js'
+
+export class WithSafeArrayWhereInPlugin implements KyselyPlugin {
+  readonly #transformer = new WithSafeArrayWhereInTransformer()
+
+  transformQuery(args: PluginTransformQueryArgs): RootOperationNode {
+    return this.#transformer.transformNode(args.node)
+  }
+
+  transformResult(
```
Blitz2145, line 18: Nit: Can just use `async` keyword here, avoid explicit `Promise.resolve` call.

src/plugin/safe-empty-array-where-in/safe-empty-array-where-in.transformer.ts
Blitz2145: @austinwoon I agree with @igalklebanov that you should switch to `1=0` and `1=1` rewrites for in/not-in clauses. Right now the transform does not seem to work correctly for the `not in` operator, since `select 1 where 1 not in (null);` does not return any rows. Following code seems to transform things correctly.

```ts
// NOTE: Translating the clause to `1 = 1` behaves semantically as truthy
if (operator.operator === 'not in') {
  return BinaryOperationNode.create(
    ValueNode.create(1),
    OperatorNode.create('='),
    ValueNode.create(1)
  );
}

// NOTE: Translating the clause to `1 = 0` behaves semantically as falsey
return BinaryOperationNode.create(
  ValueNode.create(1),
  OperatorNode.create('='),
  ValueNode.create(0)
);
```

We were testing this PR in our project, and noticed the issue, thanks for all your work!

austinwoon: Ah I see, for us that not in behaviour was okay. So you're saying the above behaviour would result in `not in (null)` returning essentially the whole table?

igalklebanov:
@austinwoon CI seems to run forever, can you check this?

src/plugin/safe-empty-array-where-in/safe-empty-array-where-in.transformer.ts
austinwoon: Sticking with original behaviour as a no-op is preferred for now. We can resolve this in another PR if you think otherwise!

austinwoon:
> @austinwoon CI seems to run forever, can you check this?

This problem seems to not be from my PR. I checked out `master` and also have a `pending` test that never stops running. The pipeline looks to be similar as well from the latest run.

could you check on this?

igalklebanov:
> > @austinwoon CI seems to run forever, can you check this?
> 
> This problem seems to not be from my PR. I checked out `master` and also have a `pending` test that never stops running. The pipeline looks to be similar as well from the latest run.
> 
> could you check on this?

Well, it is happening only in this PR. See for yourself - https://github.com/kysely-org/kysely/actions?query=workflow%3Atests.
This means, the test suite introduced in this PR is causing it.

austinwoon:
> I think that in general, this plugin (which is not a core functionality) is unnecessarily overly tested - 65 lines of source code vs. 828 lines of test code. Take `CamelCasePlugin` for example, a more complex and highly used non-core functionality, it has 329 lines of source code and only 353 lines of test code that cover all the major spots.
> 
> 
> 
> Why does it matter? well, that's more stuff to maintain (and who gets to maintain it over time?), and more time spent waiting on tests.
> 
> 
> 
> I'd cut down on setup and teardown code. If all you truly needed was an extra column, add it to the general test setup code - in this case, the existing test tables don't require any additional columns for this suite (there's a name column, and a nullable column for you to use). And as my previous comments suggested, I'd cut down on redundant/irrelevant test cases, and not execute against a database engine in every case.
> 
> 
> 
> If the `(null)` approach requires special test cases to be added, and `1=0` doesn't, and both achieve the same results, just use `1=0`. Keep It Simple. Also, if one approach might behave differently between database engines, opt for the approach that's universal. We're developing plugins for 4 core database engines, and for any SQL database engine out there.
> 
> 
> 
> By passing a `ValueNode.createImmediate(null)` instead of `null` in the transformation, you'll cut down on dialect-specific SQL string assertions, since `null`s are now hard-coded in the query and not passed as parameters - parameters add dialect-specific placeholders.

Thanks for the feedback! 

Even if we were to go with the 1=0 approach (which i did tinker with in my local branch), i wasnt thinking of reducing test suite.

The initial thought was to make this plugin have as much tests as possible since to cover all unexpected "side-effects" that could happen

I will look into rewriting the suite to cut down as much tests when I have spare time outside of work.

Also, point taken on the "core" functionality feedback. Alternatively, i can publish this under my own third party package. I understand the huge time and effort that goes into maintaining an OSS library. The last thing I want to do is add more work on reviewing/maintenance just to support non-core work. Please feel free to close this PR, I will create a separate library if you choose to do so.

austinwoon:
> ValueNode.createImmediate(null)

Thanks again for the thoughtful response. I finally made time to update the code.

I tried to do the `ValueNode.createImmediate(null)` method but that actually caused a regression spotted in the `should handle deleteFrom with returning in supported dialects` test suite. By using `ValueNode.createImmediate(null)` method, the `not in` tests actually proceeds to delete the whole table. No idea why this occurs.

Please have a look at the updated tests. I still chose to include tests for `select`, `delete` and `update` statements since those are common operations that should be tested for.

karrui:
@igalklebanov requesting for a re-review of this PR!

zaalbarxx:
Wouldn't it be better to transform it to `1 = 0` ? This is what the Laravel's query builder typically do, I think. Like:

```
    protected transformBinaryOperation(node: BinaryOperationNode): BinaryOperationNode {
        const rightOperand = node.rightOperand;
        const operator = node.operator;

        if (this.#isInOperatorNode(operator) && PrimitiveValueListNode.is(rightOperand)) {
            const values = rightOperand.values;

            if (Array.isArray(values) && values.length === 0) {
                return Object.freeze({
                    ...node,
                    leftOperand: Object.freeze(ValueNode.create(0)),
                    operator: OperatorNode.create('='),
                    rightOperand: Object.freeze(ValueNode.create(1)),
                });
            }
        }

        return node;
    }
```
I mean, I'm not sure if one is better than the other, just curious?

austinwoon:
> Wouldn't it be better to transform it to `1 = 0` ? This is what the Laravel's query builder typically do, I think. Like:
> 
> 
> 
> ```
> 
>     protected transformBinaryOperation(node: BinaryOperationNode): BinaryOperationNode {
> 
>         const rightOperand = node.rightOperand;
> 
>         const operator = node.operator;
> 
> 
> 
>         if (this.#isInOperatorNode(operator) && PrimitiveValueListNode.is(rightOperand)) {
> 
>             const values = rightOperand.values;
> 
> 
> 
>             if (Array.isArray(values) && values.length === 0) {
> 
>                 return Object.freeze({
> 
>                     ...node,
> 
>                     leftOperand: Object.freeze(ValueNode.create(0)),
> 
>                     operator: OperatorNode.create('='),
> 
>                     rightOperand: Object.freeze(ValueNode.create(1)),
> 
>                 });
> 
>             }
> 
>         }
> 
> 
> 
>         return node;
> 
>     }
> 
> ```
> 
> I mean, I'm not sure if one is better than the other, just curious?
> 
> 

Different libraries have implemented with either an equality or an in (null) check. I believe sequelize uses in (null).

Not too sure myself about perf tradeoffs. Perhaps you could try running some benchmarks to experiment!

igalklebanov:
Went over it, and expanded the use-case. You can choose which strategy to apply and there are 2 built-in strategies to choose from and support for custom ones.

This'll probably land in v0.28! 🥳

PR #926, opened 2024-03-29
chore: warn instead of throwing error if transform returns different nodes

austinwoon:
allows us to return a different node. 

Specific use case here https://discord.com/channels/890118421587578920/1223240368703799418/1223240368703799418 where I want to be able to safely detect empty array inserts and transform them into no ops with a plugin.

### DIFF
--- a/src/query-executor/query-executor-base.ts
+++ b/src/query-executor/query-executor-base.ts
@@ -37,7 +37,7 @@ export abstract class QueryExecutorBase implements QueryExecutor {
       if (transformedNode.kind === node.kind) {
         node = transformedNode as T
       } else {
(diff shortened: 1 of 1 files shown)


koskimas:
This isn't a good idea. The warning still suggests this is something you shouldn't do or things break. The result is the same. People won't do it.

Either remove the restriction alltogether or keep as-is.

austinwoon:
> This isn't a good idea. The warning still suggests this is something you shouldn't do or things break. The result is the same. People won't do it.
> 
> Either remove the restriction alltogether or keep as-is.

Ok, which would you suggest? Will leave it to you since I understand there are some principles for the library.

PR #929, opened 2024-04-02
Support accessing the last executed migration

npapagna:
This allows tools using `Migrator` to inform the current state of the database.


koskimas:
You can already use `getMigrations` to do this.

npapagna:
> You can already use `getMigrations` to do this.

Totally missed that one. Thanks a lot, @koskimas!

PR #933, opened 2024-04-04
add dependabot

igalklebanov:
Hey 👋 

Keeping up with dev dependencies' releases and triggering CI for them will help us catch if Kysely might not work for our users in certain setups, and provide fixes quickly.

We'll be hit with PR spam initially (I'll handle it), but over time it'll be OK.


igalklebanov:
Closing this, as dependabot doesn't support bun, deno, etc.

PR #953, opened 2024-04-16
add `objectStrategy` option that allows to not mutate result objects/arrays @ `ParseJSONResultsPlugin`.

igalklebanov:
Hey :wave:

This PR offers a way to avoid nasty runtime errors when result objects/arrays have non-writable props when parsing using `ParseJSONResultsPlugin`.

The default strategy is `'in-place'` (similar to how the plugin behaves in latest version).

In large objects/arrays, the new `create` strategy might significantly differ in memory consumption, as it fills new objects/arrays instead of mutating the existing ones.

### DIFF
--- a/src/plugin/parse-json-results/parse-json-results-plugin.ts
+++ b/src/plugin/parse-json-results/parse-json-results-plugin.ts
@@ -60,7 +60,7 @@ export class ParseJSONResultsPlugin implements KyselyPlugin {
 }
 
 function parseArray<T>(arr: T[], objectStrategy: ObjectStrategy): T[] {
-  const target = objectStrategy === 'create' ? [] : arr
+  const target = objectStrategy === 'create' ? new Array(arr.length) : arr
 
   for (let i = 0; i < arr.length; ++i) {
     target[i] = parse(arr[i], objectStrategy) as T
--- a/src/plugin/parse-json-results/parse-json-results-plugin.ts
+++ b/src/plugin/parse-json-results/parse-json-results-plugin.ts
@@ -15,19 +15,14 @@ export interface ParseJSONResultsPluginOptions {
    *
    * This can result in runtime errors if some objects/arrays are readonly.
    *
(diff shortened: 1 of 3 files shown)


test/node/src/parse-json-results-plugin.test.ts
```diff
@@ -0,0 +1,27 @@
+import { ParseJSONResultsPlugin } from '../../..'
+import { createQueryId } from '../../../dist/cjs/util/query-id.js'
+
+describe.only('ParseJSONResultsPlugin', () => {
```
koskimas, line 4: `describe.only`

src/plugin/parse-json-results/parse-json-results-plugin.ts
```diff
@@ -32,30 +54,32 @@ export class ParseJSONResultsPlugin implements KyselyPlugin {
...
       ...args.result,
-      rows: parseArray(args.result.rows),
+      rows: parseArray(args.result.rows, this.#objectStrategy),
     }
   }
 }
 
-function parseArray<T>(arr: T[]): T[] {
+function parseArray<T>(arr: T[], objectStrategy: ObjectStrategy): T[] {
+  const target = objectStrategy === 'create' ? [] : arr
```
koskimas, line 63: Should we use `Array.create({ length: arr.length })` instead of `[]`?

koskimas (approved):
Couple of things, otherwise looks good. Feel free to merge once at least `describe.only` is fixed.

src/plugin/parse-json-results/parse-json-results-plugin.ts
igalklebanov: Seems faster.. https://jsperf.app/fudaro

nice one!

test/node/src/parse-json-results-plugin.test.ts
igalklebanov: oops! 😅

PR #962, opened 2024-04-23
add `ControlledTransaction`.

### LINKED ISSUE #257: More fine-grained transaction API with savepoints?
```ts
try {
  const trx = await db.begin().isolationLevel('serializable').execute()
  await trx.insertInto(...).execute()

  await trx.savepoint('foobar')
  try {
    await trx.updateTable(...).execute()
    await trx.updateTable(...).execute()
  } catch (err) {
    await trx.rollbackToSavepoint('foobar')
  }

  await trx.commit()
} catch (err) {
  await trx.rollback()
}
```

igalklebanov:
closes #257.

This PR adds some optional savepoint-related methods to drivers - need to make sure we don't introduce a breaking change here!
A new `ControlledTransaction` that allows manual `commit`, `rollback` and savepoint-related commands.

```ts
const trx = await db.startTransaction().execute()

try {
  const moshe = await trx
    .insertInto('person')
    .values({ name: 'Moshe' })
    .returning('id')
    .executeTakeFirstOrThrow()

  const trxAfterFoo = await trx.savepoint('foo').execute()

  try {
    const bone = await trxAfterFoo
      .insertInto('toy')
      .values({ name: 'Bone', price: 1.99 })
      .returning('id')
      .executeTakeFirstOrThrow()

    await trxAfterFoo
      .insertInto('pet')
      .values({ 
        owner_id: moshe.id, 
        name: 'Lucky', 
        favorite_toy_id: bone.id 
      })
      .execute()
  } catch (error) {
    await trxAfterFoo.rollbackToSavepoint('foo').execute()
  }

  await trxAfterFoo.releaseSavepoint('foo').execute()

  await trx.insertInto('audit').values({ action: 'added Moshe' }).execute()

  await trx.commit().execute()
} catch (error) {
  await trx.rollback().execute()
}
```

Some design goals (that hopefully will be met once this is done) with `ControlledTransaction`:

- allow starting a controlled transaction in single connection scenario - it should re-use the connection, and not close/release it on commit/rollback by default.
~~- allow to release the connection on commit/rollback on demand via an option.~~
- once the controlled transaction is committed/rolled back, it should not allow executing any further queries with it.
- savepoint names are type-safe and autocompletion friendly - cannot release/rollback to a name that was not saved before. cannot save with a name that already exists.

### DIFF
--- a/scripts/copy-interface-documentation.js
+++ b/scripts/copy-interface-documentation.js
@@ -22,7 +22,7 @@ const OBJECT_REGEXES = [
   /^(?:export )?declare (?:abstract )?class (\w+)/,
   /^(?:export )?interface (\w+)/,
 ]
-const GENERIC_ARGUMENTS_REGEX = /<[\w"'`,{}=| ]+>/g
+const GENERIC_ARGUMENTS_REGEX = /<[\w"'`,{}=|\[\] ]+>/g
 const JSDOC_START_REGEX = /^\s+\/\*\*/
 const JSDOC_END_REGEX = /^\s+\*\//
 
@@ -123,7 +123,7 @@ function parseObjects(file) {
 function parseImplements(line) {
   if (!line.endsWith('{')) {
     console.warn(
-      `skipping object declaration "${line}". Expected it to end with "{"'`
+      `skipping object declaration "${line}". Expected it to end with "{"'`,
     )
     return []
   }
@@ -225,7 +225,7 @@ function findDocProperty(files, object, propertyName) {
     }
 
     const interfaceProperty = interfaceObject.properties.find(
-      (it) => it.name === propertyName
+      (it) => it.name === propertyName,
     )
 
     if (interfaceProperty?.doc) {
--- a/src/dialect/mssql/mssql-dialect-config.ts
+++ b/src/dialect/mssql/mssql-dialect-config.ts
@@ -65,17 +65,14 @@ export interface Tedious {
 
 export interface TediousConnection {
   beginTransaction(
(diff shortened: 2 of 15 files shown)


seivan:
I was wondering, since you already have a reference to the transaction `trxAfterFoo` and it has been named at creation with `trx.savepoint('foo').execute()` why not default to it

instead of 
```ts
trxAfterFoo.rollbackToSavepoint('foo')
```

It should technically be enough with `trxAfterFoo.rollbackToSavepoint()` or am I missing something? 

The reason I ask is because I'd like something simple like `db.startTransaction()` and `db.endTransaction()` to be in `beforeEach` and `afterEach` test, so each unit test runs with empty tables.

src/dialect/mssql/mssql-dialect-config.ts
```diff
@@ -1,3 +1,5 @@
+import { Request } from 'tedious'
```
igalklebanov, line 1: this shouldnt be here, will remove.

scripts/copy-interface-documentation.js
```diff
@@ -22,7 +22,7 @@ const OBJECT_REGEXES = [
   /^(?:export )?declare (?:abstract )?class (\w+)/,
   /^(?:export )?interface (\w+)/,
 ]
-const GENERIC_ARGUMENTS_REGEX = /<[\w"'`,{}=| ]+>/g
```
igalklebanov, line 25: up until this point, we haven't handled arrays as generics.

aq1018:
Just want to say this is a great feature, and I eagerly await for this to be released.

PR #972, opened 2024-04-27
export all operation nodes, forever.

### LINKED ISSUE #967: How to define a custom plugin using the FunctionNode?
Hello!

I was writing a custom plugin and needed to import among other nodes the `FunctionNode` but I realized it was not exported. Should it be exported or am I doing something unintended? I know all nodes are marked as internal, but I see no other way to define a custom plugin and the `OperationNodeTransformer` I use isn't marked as internal while still exposing the nodes so I'm not sure.

Thank you for the excellent project by the way!

igalklebanov:
closes #967.

Hey 👋 

Went over all nodes and made sure we export all of them.
Some were not exported.

Added a naive "good enough" script that tests we did not forget to export any nodes.


koskimas (approved):
LGTM

Issue #510, opened 2023-05-22
Support SQLITE's `ON CONFLICT` clause


johnpyp:
Sqlite supports a non-standard `ON CONFLICT` clause (https://www.sqlite.org/lang_conflict.html) that determines how inserts/updates/create tables work, taking into account the entire row. Notably, this is **distinct** from sqlite's implementation of `ON CONFLICT` for use in its `upsert` clause - https://www.sqlite.org/lang_UPSERT.html.

An example usecase of this feature is a simplified upsert, that just replaces the underlying with the new row if there's a primary key or unique constraint violation of any sort, instead of having to declare every conflict resolution separate from the main insert body:

(Note the use of `OR REPLACE` instead of `ON CONFLICT REPLACE`, which is part of the syntax to be less verbose in INSERT/UPDATE statements)

```
INSERT OR REPLACE INTO subscriptions (
  id, -- The primary key, if this is conflicted, the row gets deleted and replaced with this new one
  many,
  keys,
  that,
  will,
  always,
  be,
  upserted
) VALUES ( ... )
```

koskimas:
You can get the same result like this https://kyse.link/?p=s&i=wSDqoqqCK5mM9d5z5Ywy. Since this is trivially implemented using the current features, I don't think we should add another nonstandard method.

enrique-lozano:
> You can get the same result like this https://kyse.link/?p=s&i=wSDqoqqCK5mM9d5z5Ywy. Since this is trivially implemented using the current features, I don't think we should add another non-standard method.

But here in your example we can not inser arrays right? If we convert the `user` object to an array we can insert this array but the `doUpdateSet` fails. Maybe we should wait for #976 to get this functionality

PR #976, opened 2024-05-01
SQLite's OR CONFLICT clause for inserts

### LINKED ISSUE #916: [SQLite Dialect] Properly translate .ignore()
Currently `.ignore()` using the SQLite dialect gets translated into `INSERT IGNORE INTO ....`

Screenshot 2024-03-18 at 14 56 02

Which doesn't match the SQLite expected `INSERT OR IGNORE INTO` (see [here](https://www.sqlite.org/lang_insert.html))

Screenshot 2024-03-18 at 14 57 55

vincentiusvin:
Hello, I saw the good first issue tag and was interested in giving this a shot.

I implemented the `orAbort`, `orFail`, `orIgnore`, etc methods as discussed in #916, alongside making the SQLite override for `ignore()`.
#510 might be of interest too, since this effectively does what was asked there.

Adding support for `update or X` should be trivial, but I hope that you won't mind if I keep this PR's scope small.

~Still got a few design related questions that you can find downstairs, so I'm drafting this.~ Ready for review.

Closes #916


src/operation-node/or-conflict-node.ts
```diff
@@ -0,0 +1,30 @@
...
+import { OperationNode } from './operation-node.js'
+
+export type OrConflictNodeResolutions =
+  | 'Rollback'
+  | 'Abort'
+  | 'Fail'
+  | 'Ignore'
+  | 'Replace'
+
+export interface OrConflictNode extends OperationNode {
```
vincentiusvin, line 11: - Is making this feature an `OperationNode` the right call? 
- And the name kinda sucks since it doesn't refer to the actual syntax used. not exactly sure what else to call it though :laughing: <br/>
SQLite calls it the `ON CONFLICT Clause` https://sqlite.org/lang_conflict.html but OnConflict is definitely off the tables since we refer to upserts by it too. So I went with OR CONFLICT.

src/dialect/mysql/mysql-query-compiler.ts
```diff
@@ -72,4 +73,13 @@ export class MysqlQueryCompiler extends DefaultQueryCompiler {
       this.visitNode(node.where)
     }
   }
+
+  protected override visitOrConflict(node: OrConflictNode): void {
+    if (node.resolution === 'Ignore') {
+      this.append(' ')
```
vincentiusvin, line 79: Implemented the override for ignore per the linked issue, but I left the other calls alone. Should I also override them for consistency?

test/node/src/insert.test.ts
```diff
@@ -317,7 +317,7 @@ for (const dialect of DIALECTS) {
       }
     })
 
-    if (dialect === 'mysql') {
+    if (dialect === 'mysql' || dialect == 'sqlite') {
       it('should insert one row and ignore conflicts using insert ignore', async () => {
```
vincentiusvin, line 321: Do I need to add tests for other conflict resolution algorithms too?

src/operation-node/insert-query-node.ts
```diff
@@ -21,7 +22,7 @@ export interface InsertQueryNode extends OperationNode {
   readonly onConflict?: OnConflictNode
   readonly onDuplicateKey?: OnDuplicateKeyNode
   readonly with?: WithNode
-  readonly ignore?: boolean
+  readonly or?: OrConflictNode
```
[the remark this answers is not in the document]
igalklebanov: MySQL supports the following SQL syntax:

```sql
insert ignore into ...
```

Removing the existing property is the wrong move here, as it makes no sense for MySQL (and perhaps other dialects that have similar syntax for this) and it introduces an unwanted breaking change - nodes can be used by custom query compilers and plugins, and this breaks those that transform/visit `InsertQueryNode` and access this property.

I think it should be split - `ignore` for the MySQL way, and `or` for the SQLite way.

We might let the compiler do `insert ignore` as usual (pure WYSIWYG), or we could make `SQLiteQueryCompiler` treat `ignore` as if `or.resolution: 'ignore'` was there. I think we leaned towards the latter in the issue.

src/operation-node/or-conflict-node.ts
igalklebanov: > Is making this feature an OperationNode the right call? I didn't manage to find docs on when to make something a node and when not to.

Some questions I ask myself when presented with this dilemma.
Is it a SQL clause? does it have complex state? do we want to allow custom compilers and plugins to visit/transform it without having to override the entire parent node?

I think it is a SQL clause, and I think we should allow visiting/transforming it without having to override the entire parent node.

igalklebanov: > And the name kinda sucks since it doesn't refer to the actual syntax used. not exactly sure what else to call it though

That ship has sailed a long time ago. We have stuff like `SelectModifierNode`.

SQLite [docs](https://www.sqlite.org/lang_insert.html) use `"INSERT OR action"`, so I'd lean towards `OrActionNode`, with a `node.action` property. This could possibly be reused, outside of conflicts, to append an `OR action` to a query - action could be any string really.

src/dialect/mysql/mysql-query-compiler.ts
igalklebanov: See other comments.

src/query-compiler/default-query-compiler.ts
```diff
@@ -1739,3 +1748,13 @@ const JOIN_TYPE_SQL: Readonly<Record<JoinType, string>> = freeze({
...
+
+const OR_CONFLICT_RESOLUTION: Readonly<
+  Record<OrConflictNodeResolutions, string>
+> = freeze({
+  Abort: 'abort',
+  Fail: 'fail',
+  Ignore: 'ignore',
+  Replace: 'replace',
+  Rollback: 'rollback',
+})
```
igalklebanov, line 1760: See other comments.

This could've been avoided by just using lowercase everywhere.

test/node/src/insert.test.ts
igalklebanov: You can add a unit test for each by looping over a list of `orX` method names.

igalklebanov:
> Adding support for `update or X` should be trivial, but I hope that you won't mind if I keep this PR's scope small.

Good idea. Let's get `insert or X` out the door first.

src/operation-node/insert-query-node.ts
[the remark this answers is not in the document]
vincentiusvin: Did the override in 1c1295e90c6c86650e578e3db68108176e0e68f9. Now I'm starting to think that going with the first option and telling people to use `orIgnore()` for SQLite in the docs might be a cleaner option  :laughing:

igalklebanov:
Hey 👋 

Pushed this through the finish line.

Deprecated `InsertQueryNode.ignore` in favor of `OrActionNode`.
`DefaultQueryCompiler` would still handle it, but will warn of the deprecation.
`DefaultQueryCompiler` handles `OrActionNode` the MySQL way - since it is the more popular database - aligned with normally choosing PostgreSQL as the default handling.
`SqliteQueyrCompiler` handles `OrActionNode` the SQLite way with "or " before the `action` value.
`InsertQueryBuilder.orIgnore` is safe for accidental usage by MySQL users, the same way `InsertQueryBuilder.ignore` is safe for accidental usage by SQLite users.
`replaceInto` tests are opened for SQLite as well. The support is also reflected in the JSDocs now.

PR #977, opened 2024-05-03
add publish /w provenance workflow.

igalklebanov:
Hey 👋 

To add trust, we can follow [1](https://docs.npmjs.com/generating-provenance-statements#publishing-packages-with-provenance-via-github-actions) & [2](https://github.blog/2023-04-19-introducing-npm-package-provenance/) and publish to NPM with provenance. This requires the next versions to be published using this Github Actions workflow instead of local invocation of `npm publish`, and adding `NPM_TOKEN` to this repository's secrets on GitHub.

Now sure how I feel about the recommended snippet having an on release created trigger.


koskimas:
Adding NPM tokens to github seems too risky. Any member of kysely-org has access to it.

koskimas:
Yeah, we just shouldn't do this after all. Publishing from actions is a bad idea. Any bad actor can way too easily publish horseshit.

PR #979, opened 2024-05-05
feat: add INCLUDE clause

### LINKED ISSUE #869: `INCLUDE` clause in Postgres `createIndex`
Postgres has support for `INCLUDE` clauses in `CREATE INDEX`, can this be written in Kysely as well?

```sql
CREATE INDEX index_name
ON table_name(key_column_list)
INCLUDE(included_column_list)
```

alenap93:
Hi,
with this PR i want to add support for INCLUDE clauses in CREATE INDEX for Postgres and SQL Server.

closes #869

### DIFF
--- a/src/query-compiler/default-query-compiler.ts
+++ b/src/query-compiler/default-query-compiler.ts
@@ -907,6 +907,12 @@ export class DefaultQueryCompiler
       this.append(')')
     }
 
+    if (node.include) {
+      this.append(' INCLUDE (')
+      this.compileList(node.include)
+      this.append(') ')
+    }
+
     if (node.nullsNotDistinct) {
       this.append(' nulls not distinct')
     }
--- a/src/query-compiler/default-query-compiler.ts
+++ b/src/query-compiler/default-query-compiler.ts
@@ -908,9 +908,9 @@ export class DefaultQueryCompiler
     }
 
     if (node.include) {
-      this.append(' INCLUDE (')
+      this.append(' include (')
       this.compileList(node.include)
-      this.append(') ')
+      this.append(')')
     }
 
     if (node.nullsNotDistinct) {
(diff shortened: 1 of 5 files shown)


src/query-compiler/default-query-compiler.ts
```diff
@@ -907,6 +907,12 @@ export class DefaultQueryCompiler
       this.append(')')
     }
 
+    if (node.include) {
+      this.append(' INCLUDE (')
```
koskimas, line 911: We generate lowercase SQL. `INCLUDE` --> `include`

src/query-compiler/default-query-compiler.ts
```diff
@@ -907,6 +907,12 @@ export class DefaultQueryCompiler
...
+      this.compileList(node.include)
+      this.append(') ')
```
koskimas, line 913: Remove extra space from the end. Each conditional part adds a space at the beginning and not the end.

src/schema/create-index-builder.ts
```diff
@@ -138,6 +138,43 @@ export class CreateIndexBuilder<C = never>
...
+   * ```
+   *
+   * The generated SQL (PostgreSQL):
+   *
+   * ```sql
+   * create index "person_first_name_include_age_index" on "person" ("first_name") INCLUDE ("age")
+   * ```
+   */
+    include<CL extends string>(
+      columns: OrderedColumnName<CL>[],
```
koskimas, line 165: Why `OrderedColumnName<CL>` and not just `string` without any generics? There's no test that uses the ordered versions at least. If those are supported, add tests for them.

src/schema/create-index-builder.ts
```diff
@@ -138,6 +138,43 @@ export class CreateIndexBuilder<C = never>
...
+    ): CreateIndexBuilder<C | ExtractColumnNameFromOrderedColumnName<CL>> {
```
koskimas, line 166: Are these really supposed to be added to the list of index columns?

Issue #768, opened 2023-11-12
FileMigrationProvider doesn't work on Windows in ESM package


MChartier:
Discussed on Discord here: https://discord.com/channels/890118421587578920/1149627608175214602.

When attempting to run migrations on a Windows machine from a node project using ESM modules (as opposed to commonjs), importing the migration files will fail with a `ERR_UNSUPPORTED_ESM_URL_SCHEME` error, because the import statement is expecting a path that starts with `file:///`. But then if the caller tries to provide that instead of an absolute or relative file path, `FileMigrationProvider` will construct the path incorrectly, treating the entire URL as a relative path and concatenating it with the current path, which will also fail.

This can be fixed by updating the code in `getMigrations` in `FileMigrationProvider` to something like this:

```
const migrationPath = this.#props.path.join(this.#props.migrationFolder, fileName);
const migrationFileUrl = pathToFileURL(migrationPath);
const migration = await import(migrationFileUrl.href);
```

koskimas:
That breaks it on all other platforms.

dangnhdev:
> That breaks it on all other platforms. Since kysely is runtime agnostic and has zero dependencies, there's no way to detect the operating system and do that conditionally.

I'm getting this issue too. At least give us a hint on how to fix this please.

koskimas:
> I'm getting this issue too. At least give us a hint on how to fix this please.

`FileMigrationProvider` is 20 lines of code. Just copy-paste it to your project and change the relevant lines. How is that not the obvious solution? How is writing 20 lines of code an issue?

VincentBattu:
>  Since kysely is runtime agnostic and has zero dependencies, there's no way to detect the operating system and do that conditionally.

Naive question, but can we simply use platform from 'node:process' package to achieve this?

For example (this is quick example and, of course, the process module could be passed through the props in the constructor) :
```typescript
import { platform } from 'node:process'

export class FileMigrationProvider implements MigrationProvider {
   async getMigrations(): Promise<Record<string, Migration>> {
     // ...
          const pathParts = [
            this.#props.migrationFolder,
            fileName,
          ]
          if (platform === 'win32') {
            pathParts.unshift('file://')
          }
          const migration = await import(/* webpackIgnore: true */ this.#props.path.join(...pathParts))
    // ...
}

igalklebanov:
@VincentBattu runtime agnostic means, no dependency on NodeJS standard libraries.

VincentBattu:
[...]

PR #991, opened 2024-05-17
Possible Fix to Issue #768 ERR_UNSUPPORTED_ESM_URL_SCHEME on Windows and other posix

czuma:
i was working on a solution for this bug because work on windows and deploy on linux

i tested the solution on windows and ubuntu via (vm), works fine but only if package.json type is set to "module" don't know why the cjs commonJS build don't give me the right result.
i run the test on ubuntu and don't give me errors on the migration files so guess is "ok"

maybe i commit an error? check it and let me know, maybe we can fix this.


koskimas:
Any solution has to work with both ESM and CJS as well as on all js runtimes. Not just node.

igalklebanov:
Try using `join` from https://github.com/unjs/pathe .

PR #993, opened 2024-05-18
docs: add comments for `UpdateResult`

### LINKED ISSUE #988: Documentation for UpdateResult
[InsertResult](https://github.com/kysely-org/kysely/blob/master/src/query-builder/insert-result.ts) is very well documented, with the nuances of different dialects clearly described 👍

[UpdateResult](https://github.com/kysely-org/kysely/blob/deea8e22aae3dc4e142f055b6ec4d9a50a402bda/src/query-builder/update-result.ts), however, contains only two similarly named fields (numUpdatedRows and numChangedRows) with zero comments, which can lead to some confusion 😄

It would be nice to have similarly detailed comments in both. Unfortunately, I don't have enough information to properly document this myself in a PR. I found some info [in this issue](https://github.com/kysely-org/kysely/issues/428), but I suspect there's more to it.

Maybe @wirekang, @igalklebanov or @koskimas would be up for it? Or you could post the necessary information in this thread, which I could then prepare into a PR.

movahhedi:
Closes #988.
References #428, #431.

Based on:
https://github.com/kysely-org/kysely/blob/deea8e22aae3dc4e142f055b6ec4d9a50a402bda/src/driver/database-connection.ts#L23-L34


src/query-builder/update-result.ts
```diff
@@ -1,5 +1,14 @@
 export class UpdateResult {
+  /**
+   * The number of rows the update query updated.
+  */
   readonly numUpdatedRows: bigint
+
+  /**
+   * The number of rows the update query changed.
+   * This is **optional** and only provided by some drivers like `node-mysql2`.
```
igalklebanov, line 9:
```suggestion
   * This is **optional** and only supported in dialects such as MySQL.
```

movahhedi: Would you suggest the same change for [`database-connection.ts`](https://github.com/kysely-org/kysely/blob/deea8e22aae3dc4e142f055b6ec4d9a50a402bda/src%2Fdriver%2Fdatabase-connection.ts#L23-L34)?

https://github.com/kysely-org/kysely/blob/deea8e22aae3dc4e142f055b6ec4d9a50a402bda/src%2Fdriver%2Fdatabase-connection.ts#L23-L34

igalklebanov: Yes. changed rows also exists in `mysql`, not just `mysql2`, and its probably part of the MySQL protocol in general.

movahhedi:
Although I've written the comment, I haven't yet truly understood what "Changed Rows" means.
I've seen these words:
- Matched rows
- Affected rows
- Updated rows
- Changed rows

What do each mean? Can we make the comment any better?

igalklebanov:
> * **Matched rows**: Rows that match the condition[s] of any query.

no such concept.

> * **Affected rows**: Rows that have been affected (inserted, updated [even if not changed], or deleted) by an `INSERT`/`UPDATE`/`DELETE` query.

kind of. it varies per query kind.
https://dev.mysql.com/doc/c-api/8.0/en/mysql-affected-rows.html

> * **Updated rows**: Rows affected by an `UPDATE` query, even if the content of it hasn't changed after the execution of the query.

No such concept outside of kysely. It is the `affected rows` value under a different, friendlier, and fitting to context, name.

> * **Changed rows**: Rows affected by an `UPDATE` query that had their content changed.

I'm not sure this is only relevant to `UPDATE`. Might be also relevant to `INSERT INTO ... ON DUPLICATE KEY UPDATE`.

src/query-builder/update-result.ts
movahhedi: Applied.

src/query-builder/update-result.ts
```diff
@@ -1,5 +1,14 @@
 export class UpdateResult {
+  /**
+   * The number of rows the update query updated (even if not changed).
+  */
   readonly numUpdatedRows: bigint
+
+  /**
+   * The number of rows the update query changed.
+   * This is **optional** and only supported in dialects such as MySQL.
```
igalklebanov, line 9:
```suggestion
   * The number of rows the update query changed.
   *
   * This is **optional** and only supported in dialects such as MySQL.
```

PR #994, opened 2024-05-18
add `import` prop @ `FileMigrationProvider`.

igalklebanov:
Hey :wave:

This PR allows passing an optional custom `import` function @ `FileMigrationProvider`.
e.g. this would allow consumers to pass a function that uses `tsImport` by `tsx` to import TS files without the need to precompile them when using Node.js or a similar JavaScript runtime that doesn't have built-in TS file support.

Also adds `onDiscarded` optional prop, so external tooling can warn users about files that were ignored by the migration provider.

Also extracts file extension check to a protected method - can be overridden to only support TS files by an extending class.
Also adds `cjs` & `cts` support.

These changes were inspired by https://github.com/kysely-org/kysely-ctl/blob/main/src/kysely/ts-file-migration-provider.mts , a custom provider implementation in `kysely-ctl` that only works with TS migration files, on all runtimes.


test/node/tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "extends": "../../tsconfig-base.json",
   "include": ["src/**/*"],
   "compilerOptions": {
-    "module": "CommonJS",
-    "outDir": "dist"
+    "module": "NodeNext",
```
igalklebanov, line 5: necessary to allow importing `tsx/*`.

test/node/src/test-setup.ts
```diff
@@ -1,10 +1,10 @@
...
-import * as chaiAsPromised from 'chai-as-promised'
-import * as chaiSubset from 'chai-subset'
-import * as Cursor from 'pg-cursor'
+import chaiAsPromised from 'chai-as-promised'
+import chaiSubset from 'chai-subset'
+import Cursor from 'pg-cursor'
 import { Pool, PoolConfig } from 'pg'
 import { createPool } from 'mysql2'
-import * as Database from 'better-sqlite3'
+import Database from 'better-sqlite3'
```
igalklebanov, line 7: necessary after tsconfig changes.

vladshcherbin:
This would help so much!

PR #1024, opened 2024-06-05
Add `within group` clause support for aggregate function builder

### LINKED ISSUE #781: Add `within group (order by ...)` clause support for aggregate functions.
Hey 👋 

`within group` is a clause that allows ordering the aggregated data, usually found in concat-like functions.

This clause is supported in:

- PostgreSQL's [`rank`](https://www.postgresql.org/docs/current/functions-aggregate.html#FUNCTIONS-ORDEREDSET-TABLE).
- MSSQL's [`string_agg`](https://learn.microsoft.com/en-us/sql/t-sql/functions/string-agg-transact-sql?view=sql-server-ver16#f-generate-a-sorted-list-of-emails-per-towns).
- Oracle's [`listagg`](https://docs.oracle.com/cd/E11882_01/server.112/e41084/functions089.htm#SQLRF30030).

ivashog:
closes #781

### DIFF
--- a/src/query-builder/aggregate-function-builder.ts
+++ b/src/query-builder/aggregate-function-builder.ts
@@ -140,6 +140,8 @@ export class AggregateFunctionBuilder<DB, TB extends keyof DB, O = unknown>
   /**
    * Adds a `withing group` clause with a nested `order by` clause after the function.
    *
+   * This is only supported by some dialects like PostgreSQL or MS SQL Server.
+   *
    * ### Examples
    *
    * Most frequent person name:
@@ -150,7 +152,7 @@ export class AggregateFunctionBuilder<DB, TB extends keyof DB, O = unknown>
    *   .select((eb) => [
    *     eb.fn
    *       .agg<string>('mode')
-   *       .withinGroupOrderBy('person.name')
+   *       .withinGroupOrderBy('person.first_name')
    *       .as('most_frequent_name')
    *   ])
    *   .executeTakeFirstOrThrow()
@@ -159,7 +161,7 @@ export class AggregateFunctionBuilder<DB, TB extends keyof DB, O = unknown>
    * The generated SQL (PostgreSQL):
    *
    * ```sql
-   * select mode() within group (order by "person"."name") as "most_frequent_name"
+   * select mode() within group (order by "person"."first_name") as "most_frequent_name"
    * from "person"
    * ```
    */
--- a/src/query-builder/aggregate-function-builder.ts
+++ b/src/query-builder/aggregate-function-builder.ts
@@ -171,9 +171,10 @@ export class AggregateFunctionBuilder<DB, TB extends keyof DB, O = unknown>
   ): AggregateFunctionBuilder<DB, TB, O> {
     return new AggregateFunctionBuilder({
       ...this.#props,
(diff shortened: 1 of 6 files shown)


ivashog:
@igalklebanov, I want to consult about the API of the future `within group ()` functionality.
I see the following possible varaints:
1. Extended - `fn.agg('mode').withinGroup(wg => wg.orderBy('field'))`:
  `+` : looks like SQL
  `-` : too verbose; need to create separate `WithingGroupBuilder`
2. Simplified - `fn.agg('mode').withinGroup('field')`, because  `within group ()`  use only `order by` expression
3. Also I can try to implement both variants in one method

igalklebanov:
> @igalklebanov, I want to consult about the API of the future `within group ()` functionality. I see the following possible varaints:
> 
> 1. Extended - `fn.agg('mode').withinGroup(wg => wg.orderBy('field'))`:
>    `+` : looks like SQL
>    `-` : too verbose; need to create separate `WithingGroupBuilder`
> 2. Simplified - `fn.agg('mode').withinGroup('field')`, because  `within group ()`  use only `order by` expression
> 3. Also I can try to implement both variants in one method

Hey 👋 

We had a similar situation with `filter (where ...)`. All current SQL specs only support `where` clause there, so we decided, for simplicity to just go with `filterWhere(...)`.

So what I'd go with, for now, is `withinGroupOrderBy(...)`. WDYT?

ivashog:
> Hey 👋
> 
> We had a similar situation with `filter (where ...)`. All current SQL specs only support `where` clause there, so we decided, for simplicity, to just go with `filterWhere(...)`.
> 
> So I'd go with `withinGroupOrderBy(...)` for now. WDYT?

OK! We think in the same direction)

https://github.com/kysely-org/kysely/pull/1024/commits/c81d766336bb8bee0d276894621833d3df52b8f6#diff-0e4fd806a08c99c93c0331a3d0e75d4c726e4face483d2e21a9febe03e944593R167

src/query-builder/aggregate-function-builder.ts
```diff
@@ -95,6 +102,81 @@ export class AggregateFunctionBuilder<DB, TB extends keyof DB, O = unknown>
...
+   * ```
+   *
+   * The generated SQL (PostgreSQL):
+   *
+   * ```sql
+   * select mode() within group (order by "person"."name") as "most_frequent_name"
+   * from "person"
+   * ```
+   */
+  withinGroupOrderBy<OE extends SimpleReferenceExpression<DB, TB>>(
```
ivashog, line 167: Maybe should expand the `orderBy` argument type for support custom SQL expressions  => `withinGroupOrderBy<OE extends SimpleReferenceExpression<DB, TB> | Expression<O>>`

@igalklebanov, WDYT?

igalklebanov: It might be a good idea, but let's avoid scope creep in this PR, and open a separate issue - preferably with usage examples.

Let's make sure `withinGroupOrderBy` supports the same input types the other `orderBy`-like methods do, for now.

igalklebanov:
Hey 👋 

Pushed this through the finish line. It was very good so not a lot of changes - just some minor simplifications. 💪

Issue #1012, opened 2024-05-29
`eb(ref, op, val)` does not accept `sql` template tag escaped `op` argument.


igalklebanov:
Hey 👋 

Seems like `eb(ref, op, val)` does not accept `sql` template tag escaped operators in 2nd argument.

![image](https://github.com/kysely-org/kysely/assets/14938291/a7364256-9c17-42c9-9100-b17c32a20514)
![image](https://github.com/kysely-org/kysely/assets/14938291/e0840d36-047e-423d-86bb-350067b24483)
https://tsplay.dev/NrYbaw

Also, these are uncaught in the kysely playground. (cc @wirekang)

koskimas:
Isn't this a duplicate of https://github.com/kysely-org/kysely/issues/891? `eb` assumes the operator is arithmetic and returns the operand type and not bool.

igalklebanov:
Oops. Forgot about it. 🫢

wirekang:
> Also, these are uncaught in the kysely playground. 

[...]

Issue #891, opened 2024-02-28
`eb` assumes unknown operators to be arithmetic operators


koskimas:
Consider this snippet

```ts
where(eb => eb('first_name', sql`%>>`, 'Jennifer'))
```

This will cause a type error since the operator is assumed to be arithmetic and the output type of the expression is `string`.

One way to solve this could be

```ts
where(eb => eb('first_name', sql<boolean>`%>>`, 'Jennifer'))
```

or we bring back the `cmpr` function 😬

koskimas:
Or we could make boolean the default expression type. Unknown operators would create a boolean expression. That's probably a better assumption (?)

igalklebanov:
> Or we could make boolean the default expression type. Unknown operators would create a boolean expression. That's probably a better assumption (?)

#1579 introduced the following:

```ts
expectAssignable<Expression<SqlBool>>(eb('age', sql`OP`, 1))
expectAssignable<Expression<SqlBool>>(eb('age', sql<SqlBool>`OP`, 1))
expectAssignable<Expression<number>>(eb('age', sql<number>`OP`, 1))
expectAssignable<Expression<string>>(eb('age', sql<string>`OP`, 1))
```

PR #1029, opened 2024-06-07
feat: add the starts-with postgres comparison operator

boehs:
https://www.postgresql.org/docs/current/spgist-builtin-opclasses.html#SPGIST-BUILTIN-OPCLASSES-TABLE

There are a lot more operators as well that are arguably even more esoteric, not sure this can scale forever but.

see #891 and #1012. worth noting that `sql<boolean>` doesn’t currently work, but what you can do is

```ts
eb("tag", sql`^@`, ".").$castTo<boolean>()
```

### DIFF
--- a/src/operation-node/operator-node.ts
+++ b/src/operation-node/operator-node.ts
@@ -21,6 +21,7 @@ export const COMPARISON_OPERATORS = [
   'not ilike',
   '@>',
   '<@',
+  '^@',
   '&&',
   '?',
   '?&',


koskimas:
> worth noting that sql<boolean> does not currently work. 

Why made you think it would? It's mentioned as one potential solution for the issue. As in "this is what we could do to solve this issue".

boehs:
I just mistakenly understood what that was getting at, I’ll update the PR body

Issue #1036, opened 2024-06-12
CamelCasePlugin messes up complex type mappings with setTypeParser


ersinakinci:
I'm using Postgres and the `pg` driver, and I want to map `numeric` to complex objects using [bignumber.js](https://github.com/MikeMcl/bignumber.js):

```ts
// Map numeric to BigNumber.
PG.types.setTypeParser(1700, "text", (value) =>
  null ? null : BigNumber(value)
);
```

However, when I use CamelCasePlugin, it seems to serialize my mapped values and strip them of any methods:

```ts
import {
  Kysely,
  PostgresDialect,
  CamelCasePlugin,
  Transaction,
  sql,
} from "kysely";
import PG from "pg";
import { BigNumber } from "bignumber.js";

const { Pool } = PG;

// Map numeric to BigNumber.
PG.types.setTypeParser(1700, "text", (value) =>
  null ? null : BigNumber(value)
);

export const db = new Kysely<any>({
  dialect: new PostgresDialect({
    pool: new Pool({
      connectionString:
        "postgres://postgres:postgres@localhost:5432/scrappy-cash",
    }),
  }),
  plugins: [new CamelCasePlugin()],
  log(event) {
    if (event.level === "query") {
      console.log(event.query.sql);
      console.log(event.query.parameters);
    }
  },
});

export type DB = Kysely<any>;
export type DBClient = Kysely<any> | Transaction<any>;

const res = await sql`SELECT 0.1::numeric as num`.execute(db);

// Expected: num 0.1
// Actual: num { s: 1, e: -1, c: [ 10000000000000 ] }
console.log("num", res.rows[0].num);

// Expected: num op 0.3
// Actual: TypeError: res.rows[0].num.plus is not a function
console.log("num op", res.rows[0].num.plus(0.2).toString());
```

I originally [created an issue upstream](https://github.com/brianc/node-postgres/issues/3255) before realizing that CamelCasePlugin was the culprit.

ersinakinci:
Interesting. I messed around a bit and found that setting `maintainNestedObjectKeys: true` on the plugin config results in the type mapping working correctly.

ersinakinci:
Here's the culprit: https://github.com/kysely-org/kysely/blob/9519ffd2b6ebb6a3e6ed95724f8c171fc4bec6cc/src/plugin/camel-case/camel-case-plugin.ts#L171-L176

[...]

PR #1037, opened 2024-06-12
Go up prototype chain when checking isPlainObject

ersinakinci:
Implement a more robust version of `isPlainObject` that checks the prototype chain, [inspired by Lodash's `isPlainObject`](https://github.com/lodash/lodash/blob/a67a085cc0612f5b83d78024e507427dab25ca2c/src/isPlainObject.ts).

Fixes https://github.com/kysely-org/kysely/issues/1036.

Sorry for not writing a test, I'm a first-time contributor and the idea of adding a test for my particular use case--complex objects returned from `PG.types.setTypeMapParser`--was daunting. However, all tests pass on my machine.

### DIFF
--- a/src/util/object-utils.ts
+++ b/src/util/object-utils.ts
@@ -64,7 +64,8 @@ export function isPlainObject(obj: unknown): obj is Record<string, unknown> {
     !Array.isArray(obj) &&
     !isDate(obj) &&
     !isBuffer(obj) &&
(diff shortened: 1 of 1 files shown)


koskimas:
Can't merge untested code. Expecially for such a crucial function.

koskimas:
Fixed in master.

PR #1041, opened 2024-06-14
fix `MssqlDialect` streaming not handling backpressure.

### LINKED ISSUE #879: MSSQL Streaming does not handle backpressure
First of all thank you for this great library and support for mssql. You literally saved my life.

I'm working with rather huge datasets (3.9M rows joined with other tables) which cannot be paged without risking timeouts. Unfortunately, I cannot influence the schema or indices so I have to work with what I got and that is streaming the query results. This is were I noticed that my app crashes with OOM depending on the number of selected columns (and rows of course).

Short digression about how streaming mssql works right now:
There is an array that is filled with rows as they come in from the tedious request.
A loop (with setTimeout 0) checks if the array is larger than the requested batch size (let's call it `n`) and resolves a deferred with the same number `n` of rows.

Now this is the problem as only the **first** `n` rows are passed to the application. If the application is fast in handling the batch this creates no problem. But otherwise the row buffer fills up and is not cleared fast enough, thus consuming all the memory.

I solved this problem locally by using a (nodejs) readable stream and pausing the tedious request if the stream cannot buffer more rows. This works really great but I could only test this for reading data as I have only read-only access to the database.

Are you interested in a PR? I'm not sure about other runtimes that kysely supports.

### LINKED ISSUE #971: Tedious v18 includes different TypeScript annotations
Installing the latest version of tedious (18.2.0) will result in a typescript error when creating a new MssqlDialect.

```
Type 'Connection' is not assignable to type 'TediousConnection | Promise<TediousConnection>'.
  Type 'Connection' is not assignable to type 'TediousConnection'.
    Types of property 'beginTransaction' are incompatible.
      Type '(callback: BeginTransactionCallback, name?: string | undefined, isolationLevel?: number | undefined) => void' is not assignable to type '(callback: (error?: Error | undefined) => void, transactionId?: string | undefined, isolationLevel?: number | undefined) => void'.
        Types of parameters 'callback' and 'callback' are incompatible.
          Types of parameters 'error' and 'err' are incompatible.
            Type 'Error | null | undefined' is not assignable to type 'Error | undefined'.
              Type 'null' is not assignable to type 'Error | undefined'.
```

If the proposed solution is just to add `| null` to the definition in `mssql-dialect-config.d.ts`, I'm happy to open a PR.

igalklebanov:
fixes #879.
fixes #971.

Now MSSQL driver handles backpressure properly using `tedious`'s `pause` and `resume` methods.
Also aligns the dialect's types with `tedious` newly exported types.


src/dialect/mssql/mssql-dialect-config.ts
igalklebanov, line 1: `tedious` started shipping their own types instead of letting definitely typed handle it. and the new types are not 100% aligned with the old ones this code relied on.

src/dialect/mssql/mssql-driver.ts
```diff
@@ -238,6 +242,19 @@ class MssqlConnection implements DatabaseConnection {
     return tediousIsolationLevel
   }
 
+  #cancelRequest<O>(request: MssqlRequest<O>): Promise<void> {
+    return new Promise((resolve) => {
+      request.request.once('requestCompleted', resolve)
+
+      const wasCanceled = this.#connection.cancel()
```
igalklebanov, line 249: this is `true` when the underlying `tedious` request has completed already - all rows were received from database.

src/dialect/mssql/mssql-driver.ts
```diff
@@ -238,6 +242,19 @@ class MssqlConnection implements DatabaseConnection {
...
+
+      if (!wasCanceled) {
+        request.request.off('requestCompleted', resolve)
+        resolve(undefined)
+      }
```
igalklebanov, line 254: this is the case where there was nothing to cancel, the request completed already.

koskimas (approved):
LGTM

belgattitude:
Very helpful... thank you so much

fehnomenal:
Very nice, in my tests this works :+1:

PR #1062, opened 2024-06-26
Add `Date` as a valid return type for `max` and `min`

samclearman:
At least with postgres / pg, it's possible for the value of a field generated by the `max` and `min` functions to be a Date.  This PR updates the constraints on the types of these functions to reflect that.


igalklebanov:
Hey 👋 

Thanks! 💪 

Let's add a unit test that verifies a `Date` is returned by `pg`.

src/query-builder/function-module.ts
igalklebanov, line 1: Should also add `Date` to the unions in L536 and L595.

igalklebanov:
Pushed this through the finish line. Thanks for raising this. 🙏

Issue #1061, opened 2024-06-25
Mssql connection pooling issue. Multiple server round trips per query.


tmrclark:
## The problem

I was looking through OpenTelemtry traces of an application that uses Kysely with mssql and noticed that for every user created Kysley query, 3 SQL queries get sent to the server in succession. 

The first is sent when a connection is acquired from the [tarn](https://github.com/vincit/tarn.js) connection pool. Every time this happens, tarn calls the provided [`validate`](https://github.com/tmrclark/kysely/blob/master/src/dialect/mssql/mssql-driver.ts/#L218) method, which in this case  [`sends a query`](https://github.com/tmrclark/kysely/blob/master/src/dialect/mssql/mssql-driver.ts/#L204).

The seconds is the user defined query.

The third happens with the connection is released back to the connection pool. The [`releaseConnection`](https://github.com/tmrclark/kysely/blob/master/src/dialect/mssql/mssql-driver.ts/#L86) method calls the [`PRIVATE_RELEASE_METHOD`](https://github.com/tmrclark/kysely/blob/master/src/dialect/mssql/mssql-driver.ts/#L274), which calls the [`reset`](https://github.com/tediousjs/tedious/blob/master/src/connection.ts/#L3190) method defined in [tedious](https://github.com/tediousjs/tedious), which in turn executes [this query](https://github.com/tediousjs/tedious/blob/master/src/connection.ts/#L2469).

The 2 bookending queries create additional application latency by adding 2 round trips to the database server. I am assuming that this behavior is not expected and that there should only be one round trip to the database per user defined query. If I am wrong, please let me know.

## How to reproduce

I created a [demo repo](https://github.com/tmrclark/kysley-mssql-pool-bug-demo) with steps to reproduce this with trace logging and example traces. 

## Possible solution

I think the easiest and safest way to address this is to allow the user to configure the pool validation function and if the tedious `reset` method should be called when a connection is released. The default behavior can be the same as it is now, but this gives the opportunity to opt out.

tmrclark:
@igalklebanov @koskimas Any input on ways this could be addressed beyond what I suggest? I hope to implement a fix when time allows some time this week or the next.

koskimas:
Unfortunately I know nothing about the dialect or mssql in general. @igalklebanov is needed for this one.

igalklebanov:
Hey 👋 

Are there any risks?
Does the `mssql` library (a higher level abstraction over `tedious`) allow/do something similar?

[...]

PR #1073, opened 2024-07-08
Add additional config options to MSSQL dialect

tmrclark:
This PR is part of https://github.com/kysely-org/kysely/issues/1061.

Adds a new optional `resetConnectionOnRelease` config option to `config.tedious` that controls if the underlying tedious `reset` function is called when a connection is released back to the connection pool. The value defaults to true to maintain current behavior.

Adds new optional `validateConnections` argument to `config.tarn` that controls if tarn will execute a query to validate a connection before it is acquired. Defaults to `true` to maintain current behavior.

This is non-breaking change that will require no action for those currently using the MSSQL dialect, as it maintains current behavior.

### DIFF
--- a/src/dialect/mssql/mssql-dialect-config.ts
+++ b/src/dialect/mssql/mssql-dialect-config.ts
@@ -55,9 +55,9 @@ export interface Tedious {
   Request: TediousRequestClass
   TYPES: TediousTypes
   /**
-   * This controls weather the dialect calls the underlying tedious connection's
-   * [`reset` method](https://github.com/tediousjs/tedious/blob/master/src/connection.ts/#L3190)
-   * when a connection is released back to the pool.
+   * Controls whether connections are reset to their initial states when released back to the pool. Resetting a connection performs additional requests to the database.
+   * See {@link https://tediousjs.github.io/tedious/api-connection.html#function_reset | connection.reset}.
+   *
    * Defaults to `true`.
    */
   resetConnectionOnRelease?: boolean
@@ -152,9 +152,9 @@ export interface Tarn {
    */
   options: Omit<TarnPoolOptions<any>, 'create' | 'destroy' | 'validate'> & {
     /**
-     * Controls if tarn will validate connections before acquiring them from the pool.
-     * Connections are validated by executing the query `SELECT 1` on the connection.
-     * Defaults to `true`
+     * Controls whether connections are validated before being acquired from the pool. Connection validation performs additional requests to the database.
+     *
+     * Defaults to `true`.
      */
     validateConnections?: boolean
   }
@@ -199,7 +199,7 @@ export interface TarnPoolOptions<R> {
   min: number
   propagateCreateError?: boolean
   reapIntervalMillis?: number
-  validate(resource: R): boolean
+  validate?: (resource: R) => boolean
 }
 
 export interface TarnPendingRequest<R> {
--- a/src/dialect/mssql/mssql-dialect-config.ts
+++ b/src/dialect/mssql/mssql-dialect-config.ts
@@ -1,5 +1,3 @@
-import type { MssqlConnection } from './mssql-driver.js'
-
 export interface MssqlDialectConfig {
   /**
    * This dialect uses the `tarn` package to manage the connection pool to your
--- a/src/dialect/mssql/mssql-dialect-config.ts
+++ b/src/dialect/mssql/mssql-dialect-config.ts
@@ -199,7 +199,7 @@ export interface TarnPoolOptions<R> {
   min: number
   propagateCreateError?: boolean
   reapIntervalMillis?: number
-  validate?: (resource: R) => boolean
+  validate?(resource: R): boolean
 }
 
 export interface TarnPendingRequest<R> {
--- a/src/dialect/mssql/mssql-dialect-config.ts
+++ b/src/dialect/mssql/mssql-dialect-config.ts
@@ -1,4 +1,4 @@
-import { MssqlConnection } from './mssql-driver'
+import type { MssqlConnection } from './mssql-driver.js'
 
 export interface MssqlDialectConfig {
   /**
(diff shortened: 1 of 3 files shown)


src/dialect/mssql/mssql-dialect-config.ts
```diff
@@ -3,7 +3,8 @@ export interface MssqlDialectConfig {
    * This dialect uses the `tarn` package to manage the connection pool to your
    * database. To use it as a peer dependency and not bundle it with Kysely's code,
    * you need to pass the `tarn` package itself. You also need to pass some pool options
-   * (excluding `create`, `destroy` and `validate` functions which are controlled by this dialect),
+   * (excluding `create` and `destroy` functions which are controlled by this dialect whereas `validate`
+   * may be provided, otherwise it is also controlled by the dialect),
```
igalklebanov, line 7: let's revert this one.

src/dialect/mssql/mssql-driver.ts
```diff
@@ -56,7 +56,10 @@ export class MssqlDriver implements Driver {
       },
       // @ts-ignore `tarn` accepts a function that returns a promise here, but
       // the types are not aligned and it type errors.
-      validate: (connection) => connection.validate(),
+      validate:
+        this.#config.tarn.options.validateConnections === false
+          ? () => true
+          : (connection) => connection.validate(),
```
igalklebanov, line 62: `tarn`'s default is `() => true`. https://github.com/Vincit/tarn.js/blob/master/src/Pool.ts#L132

```suggestion
        this.#config.tarn.options.validateConnections === false
          ? undefined
          : (connection) => connection.validate(),
```

src/dialect/mssql/mssql-driver.ts
```diff
@@ -56,7 +56,10 @@ export class MssqlDriver implements Driver {
       },
       // @ts-ignore `tarn` accepts a function that returns a promise here, but
       // the types are not aligned and it type errors.
```
igalklebanov, line 58: is the `@ts-ignore` still relevant? shouldn't it be placed above L62 instead?

src/dialect/mssql/mssql-driver.ts
```diff
@@ -84,7 +87,9 @@ export class MssqlDriver implements Driver {
   }
 
   async releaseConnection(connection: MssqlConnection): Promise<void> {
-    await connection[PRIVATE_RELEASE_METHOD]()
+    if (this.#config.tedious.resetConnectionOnRelease !== false) {
+      await connection[PRIVATE_RELEASE_METHOD]()
+    }
```
igalklebanov, line 92: this leaks "lower-level" detail.
let's do it @ `connection[PRIVATE_RELEASE_METHOD]` instead.

src/dialect/mssql/mssql-dialect-config.ts
```diff
@@ -143,7 +150,14 @@ export interface Tarn {
    * Tarn.js' pool options, excluding `create`, `destroy` and `validate` functions,
    * which must be implemented by this dialect.
    */
-  options: Omit<TarnPoolOptions<any>, 'create' | 'destroy' | 'validate'>
+  options: Omit<TarnPoolOptions<any>, 'create' | 'destroy' | 'validate'> & {
+    /**
+     * Controls if tarn will validate connections before acquiring them from the pool.
+     * Connections are validated by executing the query `SELECT 1` on the connection.
+     * Defaults to `true`
```
igalklebanov, line 157:
```suggestion
     * Controls whether connections are validated before being acquired from the pool. Validation might require additional requests to the database.
     * Defaults to `true`.
```

src/dialect/mssql/mssql-dialect-config.ts
```diff
@@ -54,6 +54,13 @@ export interface Tedious {
   ISOLATION_LEVEL: TediousIsolationLevel
   Request: TediousRequestClass
   TYPES: TediousTypes
+  /**
+   * This controls weather the dialect calls the underlying tedious connection's
+   * [`reset` method](https://github.com/tediousjs/tedious/blob/master/src/connection.ts/#L3190)
+   * when a connection is released back to the pool.
+   * Defaults to `true`.
```
igalklebanov, line 61:
```suggestion
   * Controls whether connections are reset to their initial states when released back to the pool. Resetting a connection performs additional requests to the database.
   * See {@link https://tediousjs.github.io/tedious/api-connection.html#function_reset | connection.reset}.
   * 
   * Defaults to `true`.
```

src/dialect/mssql/mssql-driver.ts
tmrclark: Making this change would require updating the [`validate` type](https://github.com/tmrclark/kysely/blob/mssql-pooling-updates/src/dialect/mssql/mssql-dialect-config.ts/#L202). We can also update it to remove the need for the `@ts-ignore` if we change it to `validate?: (resource: R) => boolean | Promise<boolean>`

src/dialect/mssql/mssql-driver.ts
tmrclark: We would need to make `connection[PRIVATE_RELEASE_METHOD]()` aware of the config option in order to do that. It could be passed as a function argument `connection[PRIVATE_RELEASE_METHOD](this.#config.tedious.resetConnectionOnRelease)`, or it could be passed to the `MssqlConnection` constructor.

Any opinion?

igalklebanov: Let's pass it to the constructor.

src/dialect/mssql/mssql-dialect-config.ts
```diff
@@ -185,7 +199,7 @@ export interface TarnPoolOptions<R> {
   min: number
   propagateCreateError?: boolean
   reapIntervalMillis?: number
-  validate(resource: R): boolean
+  validate?: (resource: R) => boolean
```
igalklebanov, line 202: nitpicky

```suggestion
  validate?(resource: R): boolean
```

igalklebanov (approved):
LGTM! 🚀

PR #1079, opened 2024-07-14
Support update table1, table2, ... query support. Closes #192

### LINKED ISSUE #192: Update statements with joins don't seem to work
Hello!  Just switched from Prisma and loving the library so far.  Just ran into an issue where I'm hoping there's a workaround or I'm doing something wrong.

Basically, I have an existing MySQL statement that reads:
```sql
update OverallMV O join OverallMV O2 on O.Rank = O2.Rank and O.UserID <> O2.UserID set O.Tied = true
```

So I recreated it in Kysely as:
```ts
await trx.updateTable("OverallMV as O").innerJoin(
      "OverallMV as O2",
      (join) =>
        join
          .onRef("O.Rank", "=", "O2.Rank")
          .onRef("O.UserID", "<>", "O2.UserID"),
    ).set({ Tied: 1 }).execute();
```

However, this fails with a `ER_PARSE_ERROR` since it puts the set before the join:
```sql
update `OverallMV` as `O` set `Tied` = 1 inner join `OverallMV` as `O2` on `O`.`Rank` = `O2`.`Rank` and `O`.`UserID` <> `O2`.`UserID`
```

Any ideas here?

Side note, I'm wondering if this use case was never tested as I would also expect to choose which table to update in the set.  So more like `.set({ "O.Tied": 1 })`, which does not currently work but `.set({ Tied: 1 })` does.

koskimas:
Adds support for queries like this:

```sql
update
  person,
  pet
set
  person.first_name = ?,
  pet.name = ?
where
  person.id = pet.owner_id
  and person.name = ?
```

This kind of queries are supported by (at least) MySQL. As discussed in #192 we can't really make the join methods work with both MySQL and PostgreSQL since postgres joins add tables to the `TB` generic and MySQL would need to add to both `UT` and `TB` generics.

The `set({ first_name: 'foo' })` version of the `set` method doesn't support table qualified references. The users need to use the `set('table.column', value)` version.

### DIFF
--- a/site/docs/examples/update/0030-my-sql-joins.mdx
+++ b/site/docs/examples/update/0030-my-sql-joins.mdx
@@ -1,0 +1,40 @@
+---
+title: 'MySQL joins'
+---
+
+# MySQL joins
+
+MySQL allows you to join tables directly to the "main" table and update
+rows of all joined tables. This is possible by passing all tables to the
+`updateTable` method as a list and adding the `ON` conditions as `WHERE`
+statements. You can then use the `set(column, value)` variant to update
+columns using table qualified names.
+
+The `UpdateQueryBuilder` also has `innerJoin` etc. join methods, but those
+can only be used as part of a PostgreSQL `update set from join` query.
+Due to type complexity issues, we unfortunately can't make the same
+methods work in both cases.
+
+import {
+  Playground,
+  exampleSetup,
+} from '../../../src/components/Playground'
+
+import {
+  mySqlJoins
+} from './0030-my-sql-joins'
+
+<div style={{ marginBottom: '1em' }}>
+  <Playground code={mySqlJoins} setupCode={exampleSetup} />
+</div>
+
+:::info[More examples]
+The API documentation is packed with examples. The API docs are hosted [here](https://kysely-org.github.io/kysely-apidoc/),
+but you can access the same documentation by hovering over functions/methods/classes in your IDE. The examples are always
+just one hover away!
+
+For example, check out these sections:
+ - [set method](https://kysely-org.github.io/kysely-apidoc/classes/UpdateQueryBuilder.html#set)
+ - [returning method](https://kysely-org.github.io/kysely-apidoc/classes/UpdateQueryBuilder.html#returning)
+ - [updateTable method](https://kysely-org.github.io/kysely-apidoc/classes/Kysely.html#updateTable)
+:::
--- a/src/operation-node/update-query-node.ts
+++ b/src/operation-node/update-query-node.ts
@@ -12,6 +12,7 @@ import { ExplainNode } from './explain-node.js'
 import { LimitNode } from './limit-node.js'
 import { TopNode } from './top-node.js'
 import { OutputNode } from './output-node.js'
+import { ListNode } from './list-node.js'
 
 export type UpdateValuesNode = ValueListNode | PrimitiveValueListNode
 
@@ -38,10 +39,15 @@ export const UpdateQueryNode = freeze({
     return node.kind === 'UpdateQueryNode'
   },
 
-  create(table: OperationNode, withNode?: WithNode): UpdateQueryNode {
+  create(
+    tables: ReadonlyArray<OperationNode>,
+    withNode?: WithNode,
+  ): UpdateQueryNode {
     return freeze({
       kind: 'UpdateQueryNode',
-      table,
+      // For backwards compatibility, use the raw table node when there's only one table
+      // and don't rename the property to something like `tables`.
+      table: tables.length === 1 ? tables[0] : ListNode.create(tables),
       ...(withNode && { with: withNode }),
     })
   },
--- a/src/plugin/with-schema/with-schema-transformer.ts
+++ b/src/plugin/with-schema/with-schema-transformer.ts
@@ -1,5 +1,6 @@
 import { AliasNode } from '../../operation-node/alias-node.js'
 import { IdentifierNode } from '../../operation-node/identifier-node.js'
+import { ListNode } from '../../operation-node/list-node.js'
 import { OperationNodeTransformer } from '../../operation-node/operation-node-transformer.js'
 import { OperationNode } from '../../operation-node/operation-node.js'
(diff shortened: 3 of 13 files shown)


site/docs/examples/update/0030-my-sql-joins.mdx
```diff
@@ -0,0 +1,40 @@
...
+# MySQL joins
+
+MySQL allows you to join tables directly to the "main" table and update
+rows of all joined tables. This is possible by passing all tables to the
+`updateTable` method as a list and adding the `ON` conditions as `WHERE`
+statements. You can then use the `set(column, value)` variant to update
+columns using table qualified names.
+
+The `UpdateQueryBuilder` also has `innerJoin` etc. join methods, but those
+can only be used as a part of PostgreSQL a `update set from join` query.
```
igalklebanov, line 14:
```suggestion
can only be used as part of a PostgreSQL `update ... set ... from ... join` query.
```

src/query-creator.ts
```diff
@@ -464,39 +464,57 @@ export class QueryCreator<DB> {
...
-    ExtractTableAlias<DB, TR>,
-    ExtractTableAlias<DB, TR>,
+    ExtractTableAlias<DB, TE>,
+    ExtractTableAlias<DB, TE>,
     UpdateResult
   >
 
-  updateTable<TR extends AnyAliasedTable<DB>>(
-    table: TR,
+  updateTable<TE extends TableExpression<DB, never>>(
```
koskimas, line 476: Whoops, this shouldn't be allowed. Fixing.

koskimas:
I'm rebasing this to 0.28, but I think that branch is broken. The `QueryNode.cloneWithEndModifier` is missing.

igalklebanov:
> I'm rebasing this to 0.28, but I think that branch is broken. The `QueryNode.cloneWithEndModifier` method is missing.

Yeah, it was 3 commits behind or something. I forgot to rebase it.

koskimas:
> > I'm rebasing this to 0.28, but I think that branch is broken. The `QueryNode.cloneWithEndModifier` method is missing.
> 
> Yeah, it was 3 commits behind or something. I forgot to rebase it.

Actually, I just accidentally picked up the `modifyEnd` method when rebasing 😬

site/docs/examples/update/0030-my-sql-joins.mdx
koskimas: Fixed.

src/operation-node/update-query-node.ts
```diff
@@ -38,10 +39,15 @@ export const UpdateQueryNode = freeze({
...
+  create(
+    tables: ReadonlyArray<OperationNode>,
+    withNode?: WithNode,
+  ): UpdateQueryNode {
     return freeze({
       kind: 'UpdateQueryNode',
-      table,
+      // For backwards compatibility, use the raw table node when there's only one table
+      // and don't rename the property to something like `tables`.
+      table: tables.length === 1 ? tables[0] : ListNode.create(tables),
```
igalklebanov, line 50: +1

src/query-compiler/default-query-compiler.ts
```diff
@@ -798,6 +798,10 @@ export class DefaultQueryCompiler
     }
 
     if (node.joins) {
+      if (!node.from) {
+        throw new Error("Joins in an update query are only supported as a part of a PostgreSQL 'update set from join' query. If you want to create a MySQL 'update join set' query, see https://kysely.dev/docs/examples/update/my-sql-joins")
```
igalklebanov, line 802: nitpicky

maybe we should expose a `KyselyError`?

src/plugin/with-schema/with-schema-transformer.ts
```diff
@@ -157,14 +158,14 @@ export class WithSchemaTransformer extends OperationNodeTransformer {
...
-    if (table) {
-      this.#collectSchemableId(table.table, schemableIds)
+    if (ListNode.is(node)) {
+      for (const table of node.items) {
+        this.#collectSchemableIdsFromTableExpr(table, schemableIds)
+      }
+    } else if (TableNode.is(node)) {
+      this.#collectSchemableId(node.table, schemableIds)
+    } else if (AliasNode.is(node) && TableNode.is(node.node)) {
+      this.#collectSchemableId(node.node.table, schemableIds)
```
igalklebanov, line 168: micro-optimization: intuition tells me the frequency will be `TableNode` > `AliasNode` > `ListNode`, maybe flip the order of `if`s?

igalklebanov (approved):
Hey 👋 

Overall LGTM as always! 🚀

I left some nitpicky comments..

I wonder if we could make `set`'s object variant work for this scenario. Was it a type-complexity decision or something else?

src/operation-node/update-query-node.ts
igalklebanov: Alternatively, we could populate both `table` and `tables` and deprecate `table`.

koskimas:
> I wonder if we could make set's object variant work for this scenario. Was it a type-complexity decision or something else?

I'm afraid that auto-completion would lead postgres etc. users to use the table qualified names even though they only work on mysql. The object version works for columns that are unambiquous. For example this works:

```ts
db.updateTable(['person', 'pet'])
  .set({ first_name: 'foo', name: 'bar' })
```

`first_name` is a `person` column and `name` is a `pet` column.

src/query-compiler/default-query-compiler.ts
koskimas: Maybe, but I'd add that as a separate PR since we should use that everywhere if we so decide.

src/query-creator.ts
koskimas: Arbitrary expressions ARE okey it seems... You just naturally can't update them. Should we allow them or not? I'm leaning towards allowing them in the array but not as the lone target.

src/plugin/with-schema/with-schema-transformer.ts
koskimas: Yep, it's cleaner that way.

koskimas: Fixed.

PR #1080, opened 2024-07-15
Speedup types with huge databases. Fixes #867

### LINKED ISSUE #867: Case of extremely slow type checking
I have a query pattern that degrades tsc type checking performance to an outsized degree. It involves a nested query in `selectFrom`, and a simple `where` clause.

This is my most minimal reproduction:

```ts
const demoQuery = (db: Kysely<DB>) =>
  db
    .selectFrom((eb) =>
      eb
        .selectFrom("my_table")
        // Removing this `where` makes a dramatic difference in compilation time.
        .where((eb2) => eb2("my_table.id", "in", [1, 2, 3]))
        .selectAll()
        .as("test")
    )
    .selectAll();
```

With the `where` line indicated by the comment, running `tsc --diagnostics --noEmit` takes about 30 seconds, whereas after removing it it takes about a second.

TypeScript diagnostics with `where`:

```
Files:              435
Lines:           119235
Identifiers:     112580
Symbols:         258239
Types:           185820
Instantiations: 4743126
Memory used:    581215K
I/O read:         0.19s
I/O write:        0.00s
Parse time:       0.87s
Bind time:        0.27s
Check time:      30.30s
Emit time:        0.00s
Total time:      31.44s
```

TypeScript diagnostics WITHOUT `where`:

```
Files:              435
Lines:           119235
Identifiers:     112577
Symbols:          76912
Types:             2539
Instantiations:   10897
Memory used:    133318K
I/O read:         0.04s
I/O write:        0.00s
Parse time:       0.78s
Bind time:        0.23s
Check time:       0.25s
Emit time:        0.00s
Total time:       1.26s
```

My DB has a large number of tables, so I'm sure that's part of the issue, but I never had this problem until I added this particular pathological query.

I'm using:

- `kysely` v0.27.2
- `typescript` v5.3.3
- Node v21.6.1

---

Edit: [Here's a Github gist with all of the files needed to reproduce this issue.](https://gist.github.com/agj/b638a49ab55a38e5c8b4b21c228748d1)

koskimas:
This speeds up type checks with huge database interfaces.

The added type test case took 12 seconds to check before the fixes and about 0.4 seconds after the fixes.

Once again, I have no idea why things got faster, but it's related to function overloads. Squashing all overloads of `selectFrom`, `updateTable`, `mergeInto` and `deleteFrom` into a single signature fixes the perf issues 🤷

This is a slightly breaking change. It breaks some generics use cases, but those were pretty much unusable already anyways.

### DIFF
--- a/src/query-creator.ts
+++ b/src/query-creator.ts
@@ -11,17 +11,8 @@ import { SelectQueryNode } from './operation-node/select-query-node.js'
 import { UpdateQueryNode } from './operation-node/update-query-node.js'
 import {
   parseTable,
-  parseTableExpression,
   parseTableExpressionOrList,
-  TableExpression,
-  From,
   TableExpressionOrList,
-  FromTables,
-  TableReference,
-  TableReferenceOrList,
-  ExtractTableAlias,
-  AnyAliasedTable,
-  PickTableWithAlias,
   SimpleTableReference,
   parseAliasedTable,
 } from './parser/table-parser.js'
@@ -52,6 +43,10 @@ import {
 import { MergeQueryBuilder } from './query-builder/merge-query-builder.js'
 import { MergeQueryNode } from './operation-node/merge-query-node.js'
 import { MergeResult } from './query-builder/merge-result.js'
+import { SelectFrom } from './parser/select-from-parser.js'
+import { DeleteFrom } from './parser/delete-from-parser.js'
+import { UpdateTable } from './parser/update-parser.js'
+import { MergeInto } from './parser/merge-into-parser.js'
 
 export class QueryCreator<DB> {
   readonly #props: QueryCreatorProps
@@ -167,39 +162,17 @@ export class QueryCreator<DB> {
    *   (select 1 as one) as "q"
    * ```
    */
-  selectFrom<TE extends keyof DB & string>(
-    from: TE[],
-  ): SelectQueryBuilder<DB, ExtractTableAlias<DB, TE>, {}>
-
-  selectFrom<TE extends TableExpression<DB, never>>(
-    from: TE[],
-  ): SelectQueryBuilder<From<DB, TE>, FromTables<DB, never, TE>, {}>
-
-  selectFrom<TE extends keyof DB & string>(
+  selectFrom<TE extends TableExpressionOrList<DB, never>>(
     from: TE,
-  ): SelectQueryBuilder<DB, ExtractTableAlias<DB, TE>, {}>
-
-  selectFrom<TE extends AnyAliasedTable<DB>>(
-    from: TE,
-  ): SelectQueryBuilder<
-    DB & PickTableWithAlias<DB, TE>,
-    ExtractTableAlias<DB & PickTableWithAlias<DB, TE>, TE>,
-    {}
-  >
-
-  selectFrom<TE extends TableExpression<DB, never>>(
-    from: TE,
-  ): SelectQueryBuilder<From<DB, TE>, FromTables<DB, never, TE>, {}>
-
-  selectFrom(from: TableExpressionOrList<any, any>): any {
+  ): SelectFrom<DB, never, TE> {
     return createSelectQueryBuilder({
       queryId: createQueryId(),
       executor: this.#props.executor,
       queryNode: SelectQueryNode.createFrom(
-        parseTableExpressionOrList(from),
+        parseTableExpressionOrList(from as TableExpressionOrList<any, any>),
         this.#props.withNode,
       ),
-    })
+    }) as SelectFrom<DB, never, TE>
   }
 
   /**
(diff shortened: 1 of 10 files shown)


test/typings/test-d/huge-db.test-d.ts
igalklebanov, line 1: 💪

src/query-creator.ts
```diff
@@ -545,17 +476,9 @@ export class QueryCreator<DB> {
...
+  mergeInto<TR extends SimpleTableReference<DB>>(
     targetTable: TR,
-  ): MergeQueryBuilder<
-    DB & PickTableWithAlias<DB, TR>,
-    ExtractTableAlias<DB & PickTableWithAlias<DB, TR>, TR>,
-    MergeResult
-  >
+  ): MergeInto<DB, TR>
 
   mergeInto<TR extends SimpleTableReference<DB>>(targetTable: TR): any {
```
igalklebanov, line 483: do we still need the split between declaration and implementation?

src/query-creator.ts
```diff
@@ -464,41 +427,9 @@ export class QueryCreator<DB> {
...
-    from: TE,
-  ): UpdateQueryBuilder<
-    DB & PickTableWithAlias<DB, TE>,
-    ExtractTableAlias<DB & PickTableWithAlias<DB, TE>, TE>,
-    ExtractTableAlias<DB & PickTableWithAlias<DB, TE>, TE>,
-    UpdateResult
-  >
+  ): UpdateTable<DB, TE>
 
   updateTable(tables: TableExpressionOrList<any, any>): any {
```
igalklebanov, line 434: do we still need the split between declaration and implementation?

src/query-creator.ts
```diff
@@ -414,23 +389,11 @@ export class QueryCreator<DB> {
...
-  ): DeleteQueryBuilder<DB, ExtractTableAlias<DB, TR>, DeleteResult>
-
-  deleteFrom<TR extends TableReference<DB>>(
-    table: TR,
-  ): DeleteQueryBuilder<From<DB, TR>, FromTables<DB, never, TR>, DeleteResult>
+  deleteFrom<TE extends TableExpressionOrList<DB, never>>(
+    from: TE,
+  ): DeleteFrom<DB, TE>
 
-  deleteFrom(tables: TableReferenceOrList<DB>): any {
```
igalklebanov, line 433: do we still need the split between declaration and implementation?

src/query-creator.ts
```diff
@@ -167,29 +162,9 @@ export class QueryCreator<DB> {
...
-    ExtractTableAlias<DB & PickTableWithAlias<DB, TE>, TE>,
-    {}
-  >
-
-  selectFrom<TE extends TableExpression<DB, never>>(
-    from: TE,
-  ): SelectQueryBuilder<From<DB, TE>, FromTables<DB, never, TE>, {}>
+  ): SelectFrom<DB, never, TE>
 
   selectFrom(from: TableExpressionOrList<any, any>): any {
```
igalklebanov, line 169: do we still need the split between declaration and implementation?

igalklebanov (approved):
Hey 👋 

LGTM! 🚀 

Left a question..

Let's make sure these new types are exported @ `index.ts`.

src/query-creator.ts
koskimas: Not needed specifically, but the other option is to have `as any` casts inside the implementation. I think that's still better so I'll combine them.

koskimas:
> Let's make sure these new types are exported

Do you mean the `SelectFrom`, `UpdateTable` etc. types? I can't see how those would be useful for the users. Did you have some use case in mind?

igalklebanov:
> > Let's make sure these new types are exported
> 
> Do you mean the `SelectFrom`, `UpdateTable` etc. types? I can't see how those would be useful for the users. Did you have some use case in mind?

Come to think of it, we don't need to.

vicary:
@koskimas Unfortunately we have a moderately complex project in product that was
1. Extremely slow in IDE types where we want to upgrade, but
2. Uses string unions for common types where it breaks our code

A simplified example use case would looks like this:
Table A, B, C has a common interface of `abstract interface Node { id: string }`

The code below works before, but is breaking now.
```ts
async function getData(table: "A" | "B" | "C") {
  return await db.selectFrom(table)
    .select("id")
    .execute();
}
```

Our current workaround is `table as "A"` to force one of the union values arbitrarily, definitely not pretty. But the type gain is massive so please document it if fixing the new generic types are impossible.

igalklebanov:
Hey @vicary 👋 

This is undocumented usage of Kysely, that worked for you unintentionally.
We are working on introducing an official solution for this pattern. See https://github.com/kysely-org/kysely/pull/1434.

PR #1081, opened 2024-07-16
Improve relations recipe

koskimas:
I don't know why it took so long to figure out that this is how we should tell people to write their relation helper functions:

```ts
function pets(foreignKey: Expression<string>) {
  return jsonArrayFrom(
    db.selectFrom('pet')
      .select(['pet.id', 'pet.name'])
      .where('pet.owner_id', '=', foreignKey)
      .orderBy('pet.name')
  )
}
```

Instead of the old suggestion:

```ts
function pets(eb: ExpressionBuilder<DB, 'person'>) {
  return jsonArrayFrom(
    eb.selectFrom('pet')
      .select(['pet.id', 'pet.name'])
      .whereRef('pet.owner_id', '=','person.id')
      .orderBy('pet.name')
  )
}
```

The former only takes in the dependencies (the foreign key) as an argument and provides just as much type-safety as the old way.

The old way had a number of issues: It specified the exact `ExpressionBuilder` type, which doesn't work if the parent query has any joins or uses an alias for the `person` table.

Another benefit is that we can work with the `db` type instead of the more complex `eb` type. Especially if the relations are deeply nested, the deepest `eb` types become really slow, as they inherit their DB types from the parent `ExpressionBuilders` and etc.

Also, this is in line with how we instruct people to write any helper functions:

```ts
function doStuff(expr: Expression<number>) { 
  return sql<T>`something that uses ${expr}`
}
```

Another benefit: You can't accidentally select stuff from the parent scope.

### DIFF
--- a/site/docs/recipes/0001-relations.md
+++ b/site/docs/recipes/0001-relations.md
@@ -78,16 +78,12 @@ Simple right 😅. Yeah, not so much. But it does provide full control over the 
 Fortunately we can improve and simplify this a lot using Kysely. First let's define a couple of helpers:
 
 ```ts
-function jsonArrayFrom<O>(
-  expr: Expression<O>
-): RawBuilder<Simplify<O>[]> {
-  return sql`(select coalesce(json_agg(agg), '[]') from ${expr} as agg)`
+function jsonArrayFrom<O>(expr: Expression<O>) {
+  return sql<Simplify<O>[]>`(select coalesce(json_agg(agg), '[]') from ${expr} as agg)`
 }
 
-export function jsonObjectFrom<O>(
-  expr: Expression<O>
-): RawBuilder<Simplify<O>> {
-  return sql`(select to_json(obj) from ${expr} as obj)`
+function jsonObjectFrom<O>(expr: Expression<O>) {
+  return sql<Simplify<O>>`(select to_json(obj) from ${expr} as obj)`
 }
 ```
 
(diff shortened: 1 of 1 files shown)


site/docs/recipes/0001-relations.md
```diff
@@ -78,16 +78,12 @@ Simple right 😅. Yeah, not so much. But it does provide full control over the
 Fortunately we can improve and simplify this a lot using Kysely. First let's define a couple of helpers:
 
 ```ts
-function jsonArrayFrom<O>(
-  expr: Expression<O>
-): RawBuilder<Simplify<O>[]> {
-  return sql`(select coalesce(json_agg(agg), '[]') from ${expr} as agg)`
+function jsonArrayFrom<O>(expr: Expression<O>): {
```
igalklebanov, line 81:
```suggestion
function jsonArrayFrom<O>(expr: Expression<O>) {
```

site/docs/recipes/0001-relations.md
```diff
@@ -78,16 +78,12 @@ Simple right 😅. Yeah, not so much. But it does provide full control over the
...
+  return sql<Simplify<O>[]>`(select coalesce(json_agg(agg), '[]') from ${expr} as agg)`
 }
 
-export function jsonObjectFrom<O>(
-  expr: Expression<O>
-): RawBuilder<Simplify<O>> {
-  return sql`(select to_json(obj) from ${expr} as obj)`
+export function jsonObjectFrom<O>(expr: Expression<O>): {
```
igalklebanov, line 85:
```suggestion
export function jsonObjectFrom<O>(expr: Expression<O>) {
```

igalklebanov (approved):
LGTM! 🚀

igalklebanov:
Hey 👋 

Shouldn't this land on `master` instead of the version branch?

koskimas:
> Shouldn't this land on master instead of the version branch?

True. I'll rebase.

PR #1083, opened 2024-07-17
Improve expressions recipe

koskimas:
Use the new suggested way to create helper functions in the expressions recipe.


igalklebanov (approved):
LGTM! 🚀

PR #1085, opened 2024-07-21
add reusable helpers recipe and implement missing expression features

koskimas:
Add a new recipe about creating reusable helpers.

This PR also implements some obvious things I discovered while implementing a bunch of different kinds of helper functions.

This is just a first draft and any input is welcome.


koskimas:
I don't understand why CI is failing. It's suddenly not finding `mocha` or `playwright`. Probably all executables are suddenly not found. @igalklebanov Any ideas? I tried disabling the npm cache but it didn't help.

igalklebanov:
@koskimas first CrowdStrike killing many services, now this. 😨

site/docs/recipes/0001-reusable-helpers.md
```diff
@@ -0,0 +1,234 @@
...
+
+Kysely doesn't have a built-in `upper` function but there are at least three ways you could write this:
+
+```ts
+const lastName = 'STALLONE'
+
+const persons = await db
+  .selectFrom('person')
+  .select(['id', 'first_name'])
+  // 1. `sql` template tag. This is the least type-safe option
```
igalklebanov, line 23:
```suggestion
  // 1. `sql` template tag. This is the least type-safe option. You're providing the column name without any type-checking.
```

site/docs/recipes/0001-reusable-helpers.md
```diff
@@ -0,0 +1,234 @@
...
+  .where(
+    sql<string>`upper(last_name)`, '=', lastName
+  )
+  // 2. `sql` template tag with `ref` for type-safe references
+  .where(({ eb, ref }) => eb(
+    sql<string>`upper(${ref('last_name')})`, '=', lastName
+  ))
+  // 3. The `fn` function
```
igalklebanov, line 31:
```suggestion
  // 3. The `fn` function helps you avoid missing parentheses errors and uses refs as 1st class arguments.
```

site/docs/recipes/0001-reusable-helpers.md
koskimas: Good additions! I'll add these on separate comment lines though to keep the code more readable on mobile.

site/docs/recipes/0001-reusable-helpers.md
```diff
@@ -0,0 +1,234 @@
...
+const lastName = 'STALLONE'
+
+const persons = await db
+  .selectFrom('person')
+  .select(['id', 'first_name'])
+  // 1. `sql` template tag. This is the least type-safe option
+  .where(
+    sql<string>`upper(last_name)`, '=', lastName
+  )
+  // 2. `sql` template tag with `ref` for type-safe references
```
igalklebanov, line 27:
```suggestion
  // 2. `sql` template tag with `ref`. Anything passed to `ref` gets checked against the accumulated query context, which uses your `Database` interface as source of truth.
```

site/docs/recipes/0001-reusable-helpers.md
```diff
@@ -0,0 +1,234 @@
...
+    sql<string>`upper(${ref('last_name')})`, '=', lastName
+  ))
+  // 3. The `fn` function
+  .where(({ eb, fn }) => eb(
+    fn<string>('upper', ['last_name']), '=', lastName
+  ))
+  .execute()
+```
+
+but each option could be more readable or type-safe.
```
igalklebanov, line 38:
```suggestion
...but each option could be more readable or type-safe.
```

site/docs/recipes/0001-reusable-helpers.md
```diff
@@ -0,0 +1,234 @@
...
+  )
+```
+
+In all our examples we've used the following syntax:
+
+```ts
+.where(eb => eb(left, operator, right))
+```
+
+When the expression builder `eb` is used as a function, it creates a binary expression. All binary expression with a comparison operator are represented as a `Expression<SqlBool>`. You don't always need to return `eb(left, operator, right)` from the callback though. Since `Expressions` are composable and reusable, you can return any `Expression<SqlBool>`.
```
igalklebanov, line 170:
```suggestion
When the expression builder `eb` is used as a function, it creates a binary expression. All binary expressions with a comparison operator are represented as `Expression<SqlBool>`. You don't always need to return `eb(left, operator, right)` from the callback though. Since `Expressions` are composable and reusable, you can return any `Expression<SqlBool>`.
```

igalklebanov (approved):
LGTM! 🚀 

Love the new `$asScalar` thing.

PR #1102, opened 2024-08-05
fix typings: remove conflicts from intersecting selects 

mattkrick:
When 2 select statements use the same alias, the DB will return 2 discrete columns, but the _driver_ will merge those into an object using a conflict strategy of last write wins.

For example, `SELECT 1 as ans, true as ans` will return `{ans: 1, ans: true}`, but the driver will return this value as `{ans: true}` (I believe ALL database drivers do this, please correct me if I'm wrong!)

Currently, kysely intersects these types instead of picking the value of the last one, e.g. `{ans: number & boolean}`, which becomes `{ans: never}`.

This PR selects the type of the most recently selected alias. This is useful in real world scenarios, e.g. using `to_json` to overwrite a column that that was captured in a `selectAll()`.

### DIFF
--- a/src/parser/select-parser.ts
+++ b/src/parser/select-parser.ts
@@ -77,7 +77,7 @@ type FlattenSelectExpression<SE> = SE extends DynamicReferenceBuilder<infer RA>
   ? { [R in RA]: DynamicReferenceBuilder<R> }[RA]
   : SE
 
-type ExtractAliasFromSelectExpression<SE> = SE extends string
+export type ExtractAliasFromSelectExpression<SE> = SE extends string
   ? ExtractAliasFromStringSelectExpression<SE>
   : SE extends AliasedExpression<any, infer EA>
     ? EA
--- a/src/query-builder/select-query-builder.ts
+++ b/src/query-builder/select-query-builder.ts
@@ -366,7 +366,11 @@ export interface SelectQueryBuilder<DB, TB extends keyof DB, O>
 
   select<CB extends SelectCallback<DB, TB>>(
     callback: CB,
-  ): SelectQueryBuilder<DB, TB, O & CallbackSelection<DB, TB, CB>>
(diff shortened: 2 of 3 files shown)


src/parser/select-parser.ts
```diff
@@ -77,7 +77,7 @@ type FlattenSelectExpression<SE> = SE extends DynamicReferenceBuilder<infer RA>
   ? { [R in RA]: DynamicReferenceBuilder<R> }[RA]
   : SE
 
-type ExtractAliasFromSelectExpression<SE> = SE extends string
+export type ExtractAliasFromSelectExpression<SE> = SE extends string
```
mattkrick, line 80: Using this is MUCH more performant than `keyof Selection<...>`

src/query-builder/select-query-builder.ts
```diff
@@ -84,6 +85,28 @@ import { parseFetch } from '../parser/fetch-parser.js'
 import { TopModifier } from '../operation-node/top-node.js'
 import { parseTop } from '../parser/top-parser.js'
 
+type RemoveIndex<T> = keyof {
```
mattkrick, line 88: this is required so we don't Omit index types, e.g. `{[x: string]: unknown}`.
I did this to preserve the integrity of the unit test here: https://github.com/kysely-org/kysely/blob/ec68d6bd0b2da7899a169b3f89e0c29bbaf1bf7c/test/typings/test-d/select.test-d.ts#L39-L43.

If performance is an issue, we could modify that test & remove this

koskimas:
Thanks for the effort, but this adds unplanned changes to the extremely critical code path. And it does it by glueing stuff on top. We can't risk this kind of a change here.

mattkrick:
I'm OK with that, can you please suggest an alternative way to achieve the results as seen in the unit tests?
Today, there's no way to use `selectAll` + select `to_json`.

PR #1130, opened 2024-08-25
feat: type-safe JSON insertions/updates, `eb.jval` & `sql.jval`.

igalklebanov:
Hey :wave:

`JSONColumnType` demands manual serialization before passing values in insertions/updates. Once you do that, it is not that type-safe - you can pass any string value really.

This PR deprecates `JSONColumnType`, and introduces the stricter `Json` type that enforces type-safe insertions/updates.

Using a branded type as the `InsertType` and `UpdateType` AND as the return value of new serialization helpers in `ExpressionBuilder` and `sql` template tag, users can now insert/update and serialize their JSON objects/arrays safely into JSON columns.

Compilers/plugins can serialize the objects/arrays differently, as the value is passed around in a `ValueNode` with a `serialized` flag. The default serialization method is, well, `JSON.stringify`.

```ts
db.insertInto('person_metadata')
  .values((eb) => ({
    website: eb.jval({ url: 'https://kysely.dev' }),
    // or...
    experience: sql.jval([{
      establishment: 'The University of Life',
    }]),
  }))
```

If you want the non-type-safe old `JSONColumnType`.. we don't encourage it, but you can continue using `JSONColumnType` (until we decide to remove it) or just use:

```ts
import type { ColumnType } from 'kysely'

type MetadataColumn = ColumnType<YourJSONType, string, string>
```

For a readonly `Json`, use:

```ts
import type { ColumnType, Json, Serialized } from 'kysely'

type MetadataColumn = Json<YourJSONType, Serialized<YourJSONType>, never>
```


pawelwilczewski:
Sorry if this is not the right place to ask, but as a first-time user of Kysely and since I couldn't find a resource about this anywhere, what is the issue with typing the column using the TypeScript type directly? i.e. 

```
export type MyJson = {
  value: string;
};

export type SomeTable = {
  content: MyJson;
};
```

This seemed to work fine for me using pg, and I was considering "downgrading" this experience to using `JSONColumnType` and came across this PR.

ethanresnick:
Love this conceptually, but really think the function name should be more explicit (e.g., `eb.jsonVal`). Imo, the clarity (including for AI models) is absolutely worth the extra three characters.

PR #1170, opened 2024-10-06
add more control through configuration @ ParseJSONResultsPlugin.

igalklebanov:
Hey :wave:

This PR revisits `ParseJSONResultsPlugin` and attempts to provide more fine-grained control to consumers.

- `skipKeys` allows you to specify columns to not parse in a result object.
- `reviver` allows passing a custom reviver function to the `JSON.parse` invocations. This can allow you to instantiate native Dates, or omit keys.
- by default, `__proto__` keys are deleted to deny prototype pollution.
- `isJSON` allows overriding what is or what isn't considered a JSON string before attempting to `JSON.parse` it.

### DIFF
--- a/src/plugin/parse-json-results/parse-json-results-plugin.ts
+++ b/src/plugin/parse-json-results/parse-json-results-plugin.ts
@@ -1,6 +1,6 @@
 import { QueryResult } from '../../driver/database-connection.js'
 import { RootOperationNode } from '../../query-compiler/query-compiler.js'
-import { isPlainObject, isString } from '../../util/object-utils.js'
+import { freeze, isPlainObject, isString } from '../../util/object-utils.js'
 import { UnknownRow } from '../../util/type-utils.js'
 import {
   KyselyPlugin,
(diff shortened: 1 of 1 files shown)


myndzi:
Coming from https://github.com/kysely-org/kysely/issues/1275 for context

I know you are doing some absolute _black magic_ with the types, so I'm not sure how cleanly this can be done with respect to revivers (though possibly it'd be cleaner?)

I'd much prefer explicit opt-in than an opt-out list by column names. `skipKeys` seems to exist at the wrong level of abstraction here; it doesn't entirely solve the problem because it's not defined at the table/column level. What are the chances of implementing a syntax more like this?

```ts
const rows = db.selectFrom('table').select(eb => eb.json('column_name')).execute();
```

Since the expression builder has the capability to pass forward its column's expected type, this should allow for syntax like:

```ts
const rows = db.selectFrom('table').select(eb => eb.json('column_name', reviver)).execute();
```

Obviously this is less convenient, but it is more safe and flexible, since you can define a type-safe reviver (or validation) function, type guard, etc.

For convenience's sake, if you wanted to globally define JSON parsing behavior at the plugin level, you could potentially do something like this:

```ts
interface ParseJSONResultsPluginOptions<Database> {
  revivers: {[Table in keyof Database]: {[Column in keyof Table]: Table[Column] extends Reviver<infer P> ? P : never}}
}
```
(not fully sure if that syntax would work, and of course it could be extracted into further generics)

The idea being that, if you supply the plugin with your database interface (which is expected to match the same interface that you provide to the Kysely instance, and this should be enforceable?) - then you can define revivers for specific columns on specific tables, both of which are fully type-safe, and ensure that the reviver supplied returns the type defined on the table's interface.

PR #1175, opened 2024-10-10
feat: add default to current timestamp

### LINKED ISSUE #1162: Improve DX around column default values and common raw SQL usage.
Hey 👋 

Temporal columns with a default value being current date and time are very common.
Forcing users to do:

```ts
import { sql } from 'kysely'

.addColumn('created_at', 'datetime', (cb) => cb.defaultTo(sql`CURRENT_TIMESTAMP`)) // MySQL / SQLite / PostgreSQL / MSSQL
// or
.addColumn('created_at', 'timestampz', (cb) => cb.defaultTo(sql`now()`)) // PostgreSQL
```

everywhere, is not ideal.

These also come up in issues/discord more often than desired.
There are also other `defaultTo` use cases that use raw SQL and are quite common.

Worth considering:

1. deprecate `defaultTo`.
2. introduce `defaultLit` for literals (strings, numbers, booleans, etc.).
3. introduce `defaultRaw` that accepts any raw SQL string, but has autocompletion for the common stuff like `CURRENT_TIMESTAMP`, `NOW()`, `gen_random_uuid()`, etc. no longer have to import `sql` template tag. Better DX.

austinwoon:
Closes #1162


package.json
```diff
@@ -7,10 +7,17 @@
...
+  "typesVersions": {
+    "<4.6": {
+      "*": [
+        "outdated-typescript.d.ts"
+      ]
+    }
+  },
   "exports": {
     ".": {
       "import": "./dist/esm/index.js",
```
nick-w-nick, line 23: According to the TS [documentation](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#exports-is-prioritized-over-typesversions), the `exports` key takes precedence over the `typesVersions` key.

This change was technically made in TS 4.9, but one of the maintainers indicated [here](https://github.com/microsoft/TypeScript/pull/50890#issuecomment-1256406529) that non-`exports` keys cannot contribute to the resolution when an `exports` key is present, so this should still be applicable.

```suggestion
  "exports": {
    ".": {
      "types@<4.6": "outdated-typescript.d.ts",
      "types": "./dist/esm/index.d.ts",
      "import": "./dist/esm/index.js",
```

igalklebanov: Hey 👋 

You're commenting on the wrong PR.

Anyway, this still works for the version being targeted. If the cutoff eventually happens on a version where `typesVersions` is no good anymore, we'll change that.

PR #1178, opened 2024-10-12
fix(docs): add missing import to migrations example script

mxkxf:
When copying and pasting the example migration script, there is an unknown `Database` type:

Screenshot 2024-10-12 at 18 10 55

This PR adds the import relative to the current directory. Not a biggie but might help for folks new to the tech.


site/docs/migrations.mdx
```diff
@@ -158,6 +158,7 @@ import {
   PostgresDialect,
   FileMigrationProvider,
 } from 'kysely'
+import { Database } from './types';
```
igalklebanov, line 161:
```suggestion
import { Database } from './types'
```

PR #1181, opened 2024-10-13
feat: export missing types

### LINKED ISSUE #1180: Missing types of core functionality
Hello @koskimas,

first of all, i think this is an incredibly query builder!
But sadly, it's only a query builder. 
I will try to build an orm around this lovely library like [TypeOrm](https://typeorm.io/) or [MikroOrm](https://mikro-orm.io/).
For that, i need core types and functionality.

xpr0gamers:
closes #1180

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,10 +2,7 @@ export * from './kysely.js'
 export * from './query-creator.js'
 
 export * from './expression/expression.js'
-export {
-  ExpressionBuilder,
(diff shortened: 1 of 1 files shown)


koskimas:
Unfortunately we can't export all of that. It's not exported for a reason: the internal types and utilities can change at any given moment. If we exported them, we'd need to keep them backwards compatible. That'd slow down development significantly.

PR #1266, opened 2024-11-18
docs: clarify that after rolling back a transaction the exception gets thrown again

twiddler:
I needed to know whether I can `try … catch …` exceptions that are thrown in the callback to `db.transcation().execute(…)` or whether they get swallowed. The docs are not clear about this.

I added a few words to the docs to clarify. (This is the [relevant code](https://github.com/kysely-org/kysely/blob/25b6715af7f941e4fb455f63fc2dd5584965b4a3/src/kysely.ts#L572).)


igalklebanov (approved):
Hey 👋 

LGTM! 🚀

Issue #1281, opened 2024-11-26
[CamelCasePlugin] underscore might added twice when `underscoreBeforeDigits` is set to `true` globally


bakasmarius:
**Given**:
I have a PostgreSQL database that has some columns with underscores before the digits, i.e.:
![image](https://github.com/user-attachments/assets/e1e0c6c1-30aa-4021-8a29-4bd081baaa33)
and I have configured Kysely DB to use `CamelCasePlugin`
and I have set `CamelCasePlugin` options to use `{ underscoreBeforeDigits: true }` (so that working with columns that have an underscore + digit, such as `address_row_1`, would not cause any errors)

**When**:
I select from a subquery

**Then**:
_Actual behavior_: `addressRow1` becomes `address_row__1` in a final SQL, thus resulting in an error, because such a column does not exist (I think because `createSnakeCaseMapper` executes conversion twice)
_Expected behavior_: `addressRow1` becomes `address_row_1` in a final SQL resulting in no errors

**Steps to reproduce**:
In `Kysely` repository, in `camel-case.test.ts`, add this test, run it and see that it's failing:

<details>
   <summary>camel-case.test.ts</summary>

   ```
    it('should respect underscoreBeforeDigits and not add a second underscore in a nested query', async () => {
      interface CamelTestDatabase {
        camelTest: CamelTest
      }

      interface CamelTest {
        id: Generated<number>
        firstName: string
        lastName: string
        addressRow1: string
        addressRow2: string
      }

      let dbForMigration = new Kysely<CamelTestDatabase>({
        ...ctx.config,
        plugins: [new CamelCasePlugin()],
      })

      await dbForMigration.schema.dropTable('test_camel').ifExists().execute()
      await createTableWithId(dbForMigration.schema, dialect, 'test_camel')
        .addColumn('first_name', 'varchar(255)')
        .addColumn('last_name', 'varchar(255)')
        .addColumn('address_row_1', 'varchar(255)')
        .addColumn('address_row_2', 'varchar(255)')
        .execute()

      let db = new Kysely<CamelTestDatabase>({
        ...ctx.config,
        plugins: [new CamelCasePlugin({ underscoreBeforeDigits: true })],
      })

      if (dialect === 'mssql' || dialect === 'sqlite') {
        db = db.withPlugin(new ParseJSONResultsPlugin())
      }

      const originalQuery = db.selectFrom('camelTest').select('addressRow1')
      const nestedQuery = db
[...]

PR #1290, opened 2024-12-11
fix: implements a underscore validation to camelcase plugin

RyanV-Souza:
This is my proposed solution to issue #1281 . If there are alternative approaches or suggestions, I would be happy to discuss them.

### DIFF
--- a/src/plugin/camel-case/camel-case.ts
+++ b/src/plugin/camel-case/camel-case.ts
@@ -31,7 +31,7 @@ export function createSnakeCaseMapper({
       // If underScoreBeforeDigits is true then, well, insert an underscore
       // before digits :). Only the first digit gets an underscore if
       // there are multiple.
-      if (underscoreBeforeDigits && isDigit(char) && !isDigit(prevChar)) {
+      if (underscoreBeforeDigits && isDigit(char) && !isDigit(prevChar) && !hasUnderscore(out)) {
         out += '_' + char
         continue
       }
@@ -114,6 +114,10 @@ function isAllUpperCaseSnakeCase(str: string): boolean {
 
 function isDigit(char: string): boolean {
   return char >= '0' && char <= '9'
+}
+
+function hasUnderscore(str: string): boolean {
+  return str.endsWith('_')
 }
 
 function memoize(func: StringMapper): StringMapper {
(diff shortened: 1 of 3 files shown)


src/plugin/camel-case/camel-case.ts
```diff
@@ -116,6 +116,10 @@ function isDigit(char: string): boolean {
   return char >= '0' && char <= '9'
 }
 
+function hasUnderscore(str: string): boolean {
+  return str.endsWith('_')
+}
+
```
igalklebanov, line 122: This one doesn't warrant a function yet.

```suggestion
```

src/plugin/camel-case/camel-case.ts
```diff
@@ -31,7 +31,7 @@ export function createSnakeCaseMapper({
       // If underScoreBeforeDigits is true then, well, insert an underscore
       // before digits :). Only the first digit gets an underscore if
       // there are multiple.
-      if (underscoreBeforeDigits && isDigit(char) && !isDigit(prevChar)) {
+      if (underscoreBeforeDigits && isDigit(char) && !isDigit(prevChar) && !hasUnderscore(out)) {
```
igalklebanov, line 34:
```suggestion
      if (underscoreBeforeDigits && isDigit(char) && !isDigit(prevChar) && !out.endsWith('_')) {
```

test/node/src/camel-case.test.ts
```diff
@@ -309,6 +309,93 @@ for (const dialect of DIALECTS) {
...
+        .execute()
+   
+      let db = new Kysely<CamelTestDatabase>({
+        ...ctx.config,
+        plugins: [new CamelCasePlugin({ underscoreBeforeDigits: true })],
+      })
+   
+      if (dialect === 'mssql' || dialect === 'sqlite') {
+        db = db.withPlugin(new ParseJSONResultsPlugin())
+      }
```
igalklebanov, line 345: Let's reuse the existing setup as much as possible - adding necessary columns to existing interfaces and DDL queries.

RyanV-Souza:
Thanks for the review, I've updated the code to be in line with the reviews and the good practices of the repository

PR #1311, opened 2025-01-05
fix: resolve edge case with varying shapes in multiple item inserts causing `undefined` in parameters.

### LINKED ISSUE #1310: Batch inserts use "undefined" instead of falling back on default value
When doing batch inserts (multiple values in a single INSERT), missing values are being replaced with DEFAULT values, but undefined values are not. Single inserts handle both missing and undefined values correctly.

Edit: The original repro wasn't great; the real repro is here: https://kyse.link/0feir

naorpeled:
closes #1310

### DIFF
--- a/test/node/src/insert.test.ts
+++ b/test/node/src/insert.test.ts
@@ -729,7 +729,7 @@ for (const dialect of DIALECTS) {
         {
           first_name: 'Foo',
           last_name: 'Spam',
-          middle_name: undefined,
+          middle_name: 'Bo',
           gender: 'other',
         },
         {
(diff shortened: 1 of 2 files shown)


koskimas:
Looking good otheriwse but: when inserting a single row, the value should actually be missing and not `DEFAULT`. The change now caused `DEFAULT` being use even when you insert a single row with `undefined` value.

Most databases support `DEFAULT` but not all. For example on sqlite we need to use `null` instead. Mapping `undefined` to null is not a good idea.

naorpeled:
> Looking good otheriwse but: when inserting a single row, the value should actually be missing and not `DEFAULT`. The change now caused `DEFAULT` being used even when you insert a single row with an `undefined` value (see all the failing tests).
> 
> Most databases support `DEFAULT` but not all. For example on sqlite we need to use `null` instead. Mapping `undefined` to null is not a good idea when it can be avoided.

Makes sense,
how would you separate those two cases in the cleanest way?

koskimas:
> how would you separate those two cases in the cleanest way?

Don't know. Maybe just a special case when `rows.length === 1`? Add a comment why the special case is there in case you end up with this solution.

naorpeled:
@koskimas let me know what you think 🙏

koskimas (approved):
LGTM 👍

test/node/src/insert.test.ts
```diff
@@ -742,20 +742,20 @@ for (const dialect of DIALECTS) {
...
-          parameters: ['Foo', 'Spam', 'Bo', 'other', 'Baz', 'other'],
+          sql: 'insert into "person" ("first_name", "last_name", "middle_name", "gender") values (@1, @2, default, @3), (@4, default, default, @5)',
+          parameters: ['Foo', 'Spam', 'other', 'Baz', 'other'],
         },
         sqlite: {
-          sql: 'insert into "person" ("first_name", "last_name", "middle_name", "gender") values (?, ?, ?, ?), (?, null, null, ?)',
-          parameters: ['Foo', 'Spam', 'Bo', 'other', 'Baz', 'other'],
+          sql: 'insert into "person" ("first_name", "last_name", "middle_name", "gender") values (?, ?, null, ?), (?, null, null, ?)',
+          parameters: ['Foo', 'Spam', 'other', 'Baz', 'other'],
         },
```
igalklebanov, line 759: should probably make a separate test case for this.

naorpeled: It felt redundant to me as there's another test case that already covers this as far as I remember

naorpeled: Won't be able to continue this PR until next week as I don't have my laptop on me, let's talk over Discord and decide what to do

igalklebanov:
Hey @naorpeled @koskimas  👋 

The [approved changes](https://github.com/kysely-org/kysely/pull/1311/commits/8811f39a79b3851bd518e15e01f99b90698d0c4c) resolved the issue but introduced an unnecessary regression in the form of needlessly longer queries.

Here's an added test case that reproduces #1310 output with those changes:
image

I have reverted those changes and found the issue in:

```ts
  const hasMissingColumns = rowColumns.length < columns.size
```

This check was `false` because `columns.size` was equal to `rowColumns.length`, but the row had a missing column actually, replaced by an unindexed column with an `undefined` value.

PR #1315, opened 2025-01-08
Add a reference to `ref` in `table`'s docstring

AlexErrant:
There's a `.table()` fn, but there's no `.col()` fn. Thought it would be nice to include a reference to `ref` to make discoverability easier.


igalklebanov (approved):
LGTM! 🚀

PR #1316, opened 2025-01-09
Support json_agg(column_ref)

### LINKED ISSUE #1280: Support `json_agg(column_ref)`.
Hey 👋 

Currently `eb.fn.jsonAgg` supports table names and expressions as inputs. In such cases, the return type is a JSON object array.

`json_agg` also supports passing a column reference. In such cases, the return type is a literal array.
We should support this.

```ts
const result = await db
  .selectFrom('person')
  .innerJoin('pet', 'pet.owner_id', 'person.id')
  .select((eb) =>
    eb.fn.jsonAgg('pet.name').as('person_pets')
  )
  .executeTakeFirstOrThrow()
```

The return type would be `{ person_pets: string[] }`.

SimonSimCity:
Fixes #1280.

Wow, my first PR in this repo 🤩

If there was something wrong about what I wrote - e.g. if the types are too narrow or wide, please let me know.

Here's what I know about this:
* Postgresql returns an array if there are elements in the linked table.
* Postgresql returns null otherwise.

According to the docs, this method does only exist for Postgresql, and I therefore didn't look into any other of the options.

### DIFF
--- a/src/query-builder/function-module.ts
+++ b/src/query-builder/function-module.ts
@@ -664,11 +664,13 @@ export interface FunctionModule<DB, TB extends keyof DB> {
   any<T>(expr: Expression<ReadonlyArray<T>>): ExpressionWrapper<DB, TB, T>
 
   /**
-   * Creates a json_agg function call.
+   * Creates a `json_agg` function call.
    *
-   * This function is only available on PostgreSQL.
+   * This is only supported by some dialects like PostgreSQL.
    *
-   * You can use it either on a table:
+   * ### Examples
+   *
+   * You can use it on table expressions:
    *
    * ```ts
    * await db.selectFrom('person')
@@ -687,42 +689,26 @@ export interface FunctionModule<DB, TB extends keyof DB> {
    * group by "person"."first_name"
    * ```
    *
-   * ... or on a column:
+   * or on columns:
    *
    * ```ts
    * await db.selectFrom('person')
    *   .innerJoin('pet', 'pet.owner_id', 'person.id')
-   *   .select((eb) => ['first_name', eb.fn.jsonAgg('pet.names').as('pet_names')])
-   *   .groupBy('person.first_name')
-   *   .execute()
-   *
-   * await db.selectFrom('person')
    *   .select((eb) => [
    *     'first_name',
-   *     eb
-   *       .selectFrom('pet')
-   *       .select((eb) => [eb.fn.jsonAgg("pet.owner_id")])
-   *       .whereRef('pet.owner_id', '=', 'person.id')
-   *       .as('petnames'),
+   *     eb.fn.jsonAgg('pet.name').as('pet_names'),
    *   ])
+   *   .groupBy('person.first_name')
    *   .execute()
    * ```
    *
    * The generated SQL (PostgreSQL):
    *
    * ```sql
-   * select "first_name", json_agg("pet"."names") AS "pet_names"
+   * select "first_name", json_agg("pet"."name") AS "pet_names"
    * from "person"
    * inner join "pet" ON "pet"."owner_id" = "person"."id"
    * group by "person"."first_name"
-   *
-   * select "first_name",
-   *   (
-   *     select json_agg("pet"."owner_id")
-   *     from "pet"
-   *     where "pet"."owner_id" = "person"."id"
-   *   ) as "petnames"
-   * from "person"
    * ```
    */
   jsonAgg<T extends (TB & string) | Expression<unknown>>(
@@ -736,6 +722,7 @@ export interface FunctionModule<DB, TB extends keyof DB> {
         ? O[]
         : never
   >
+
   jsonAgg<RE extends StringReference<DB, TB>>(
     column: RE,
   ): RawBuilder<ExtractTypeFromStringReference<DB, TB, RE>[] | null>
--- a/src/query-builder/function-module.ts
+++ b/src/query-builder/function-module.ts
@@ -725,7 +725,11 @@ export interface FunctionModule<DB, TB extends keyof DB> {
 
   jsonAgg<RE extends StringReference<DB, TB>>(
     column: RE,
-  ): RawBuilder<ExtractTypeFromStringReference<DB, TB, RE>[] | null>
+  ): AggregateFunctionBuilder<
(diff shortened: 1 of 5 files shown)


SimonSimCity:
Since it's likely that people will use this functionality with `coalesce`, I thought of adding a helper function, similar to the existing `jsonArrayFrom`. I don't know how you have it in regards to other dialects, because I would only be able to add it for Postgres. Here's an example of how this could be used: https://github.com/kysely-org/kysely/issues/1280#issuecomment-2581016249

igalklebanov:
> Since it's likely that people will use this functionality with `coalesce`, I thought of adding a helper function, similar to the existing `jsonArrayFrom`. I don't know how you have it in regards to other dialects, because I would only be able to add it for Postgres. Here's an example of how this could be used: [#1280 (comment)](https://github.com/kysely-org/kysely/issues/1280#issuecomment-2581016249)

I'll need to think about the helper. You're right about the usage pattern due to nullability.

1. How would you name such a helper without making stuff confusing (due to the existence of `jsonArrayFrom`)?
2. Can it be replicated with other dialects? people might get disappointed with other dialects not having something similar.
3. Not that big of a lift compared to other helpers:

```ts
(eb) => eb.coalesce(eb.jsonAgg("col"), sql<string[]>`[]`)
```

vs. 

```ts
import { someHelper } from 'kysely/helpers/postgres'

(eb) => someHelper(eb.ref("col"))
```

The existing helpers were born out of a LOT of signals from the community needing something after they ditched some ORM and did not know how to do nested stuff with SQL.

src/query-builder/function-module.ts
```diff
@@ -696,6 +723,10 @@ export interface FunctionModule<DB, TB extends keyof DB> {
         : never
   >
 
+  jsonAgg<RE extends StringReference<DB, TB>>(
+    column: RE,
+  ): RawBuilder<ExtractTypeFromStringReference<DB, TB, RE>[] | null>
```
igalklebanov, line 728: Why is a `RawBuilder` returned here and not `AggregateFunctionBuilder`?

test/typings/test-d/postgres-json.test-d.ts
```diff
@@ -169,6 +169,39 @@ async function testPostgresJsonAgg(db: Kysely<Database>) {
...
+        .as('pet_names'),
+    ])
+    .execute()
+
+  expectType<
+    {
+      first_name: string
+      pet_names: number[] | null
+    }[]
+  >(r6)
```
igalklebanov, line 204: This one doesn't add much value, let's remove it.

```suggestion
```

src/query-builder/function-module.ts
igalklebanov: e.g. this is valid SQL:

```sql
SELECT name,
  json_agg(laptop.model order by laptop.model DESC) as models
from employee 
left join laptop on employee.empId = laptop.assignedTo 
group by employee.empId;
```
https://onecompiler.com/postgresql/435v97hnu

igalklebanov:
This could really use a unit test.

src/query-builder/function-module.ts
SimonSimCity: It's due to my inexperience with this library 😅 Thanks for pointing this out. Should I add a query like this to the tests as well?

test/typings/test-d/postgres-json.test-d.ts
SimonSimCity: Could be that this typing-wise doesn't make a difference, but to me this is another use-case. It's using the json-agg in a subquery rather than in a joined table. In a joined table you need a `groupBy`, but in a subquery you don't.

SimonSimCity:
Could you assist me in writing a unit test?

test/typings/test-d/postgres-json.test-d.ts
igalklebanov: Typings tests are there to ensure we got things right on the type-level. Not to document the SQL spec.

The relationship between a root query and it's sub-queries has been thoroughly tested elsewhere.
Group by doesn't affect anything on the type-level. This test case doesn't add any value here.

Excessive tests means:

1. more stuff to maintain.
2. more stuff to wait for locally or in CI.

src/query-builder/function-module.ts
igalklebanov: > It's due to my inexperience with this library 😅

When you're new to a codebase, it's perfectly fine to not get things right the first time.

I'm just asking questions to get a feel for your thought process here.

You either have a good reason for your choice or by asking, I could find something that could be improved, so the next time, it's clearer what the correct choice should be.

igalklebanov:
> Could you assist me in writing a unit test? What unit do you think should be tested though not to make it an integration test?

Sure. We could also pair up on this if you want (DM on Discord).

We have a suite for aggregate functions, probably there.

SimonSimCity:
@igalklebanov I've just taken some time to look at writing a unit test - it was not as hard as I assumed. Is this how you envisioned it?

src/query-builder/function-module.ts
```diff
@@ -24,6 +24,7 @@ import { SelectQueryBuilderExpression } from '../query-builder/select-query-buil
 import { isString } from '../util/object-utils.js'
 import { parseTable } from '../parser/table-parser.js'
 import { Selectable } from '../util/column-type.js'
+import { RawBuilder } from '../raw-builder/raw-builder.js'
```
igalklebanov, line 27: shouldn't this be removed now?

igalklebanov (approved):
> @igalklebanov I've just taken some time to look at writing a unit test - it was not as hard as I assumed. Is this how you envisioned it?

Yeah, looks good. Most times there's also a sql output check between execution and query, but I see this suite doesn't do these so it's fine.

> I saw that you rebased my PR on the branch 0.28, which had some tests failing - at least at the stage it was back then. Now, as I've rebased it on master, all tests run through.

Yeah since this PR introduces new functionality, we wanna keep it out of any documentation so it's parked in the next major release branch.

src/query-builder/function-module.ts
SimonSimCity: Sure, not needed 😎👍

PR #1317, opened 2025-01-11
add FUNDING.md

igalklebanov:
Hey 👋 

This PR adds a `FUNDING.md` file explicitly stating we do not accept donations and why.


koskimas (approved):
Good!

Issue #1112, opened 2024-08-10
Support CROSS JOIN and CROSS JOIN LATERAL


mertalev:
Kysely supports `LEFT JOIN LATERAL` and `INNER JOIN LATERAL`, but doesn't support [lateral subqueries](https://www.postgresql.org/docs/current/queries-table-expressions.html#QUERIES-LATERAL) of the syntax `SELECT * FROM a, LATERAL ( ) l`, or the equivalent `SELECT * FROM a CROSS JOIN LATERAL ( ) l`. #140 discussed the reference of multiple tables in `selectFrom` as an alternative to cross joins, but this doesn't work for subqueries as far as I can tell.

This is the kind of query I'd like to use this feature for:
```sql
WITH RECURSIVE cte AS (
  (
    SELECT city, "assetId"
    FROM exif
    ORDER BY city
    LIMIT 1
  )

  UNION ALL

  SELECT l.city, l."assetId"
  FROM cte c
  CROSS JOIN LATERAL (
    SELECT city, "assetId"
    FROM exif
    WHERE city > c.city
    ORDER BY city
    LIMIT 1
  )  l
)
select "assets".*, jsonb_strip_nulls(to_jsonb(exif)) as "exifInfo"
from "assets"
inner join "exif" on "assets"."id" = "exif"."assetId"
inner join "cte" on "assets"."id" = "cte"."assetId";
```

koskimas:
Doesn't `leftJoinLateral` work just as well in this case?

mertalev:
In the example I shared with the `exif` and `assets` tables both having 2 million rows (one-one), using `LEFT JOIN LATERAL ... ON true` took at least 60s - I canceled it at that point. `CROSS JOIN` took 7ms.

koskimas:
How about `inner join on true`? Isn't that 100% equivalent to cross join?

koskimas:
But I guess we should add cross join and cross join lateral to Kysely.

mertalev:
> How about `inner join on true`? Isn't that 100% equivalent to cross join?

Nice suggestion! This one does have the same query plan and performance as cross join.

bakasmarius:
> How about `inner join on true`? Isn't that 100% equivalent to cross join?

Probably no because on MS SQL Server 2016 I get this error when I try to run `inner join MyTable as mt on true`:
`SQL Error [4145] [S0001]: An expression of non-boolean type specified in a context where a condition is expected`

[...]

PR #1325, opened 2025-01-18
Add support for cross join and cross join lateral

ersinakinci:
Implements https://github.com/kysely-org/kysely/issues/1112

`CROSS JOIN LATERAL` in Postgres: https://www.postgresql.org/docs/current/queries-table-expressions.html#QUERIES-LATERAL


koskimas (approved):
LGTM. Thank you!

test/node/src/join.test.ts
```diff
@@ -778,6 +809,41 @@ for (const dialect of DIALECTS) {
...
+            )
+            .select(['first_name', 'p.name'])
+            .orderBy(['first_name', 'p.name'])
+
+          testSql(query, dialect, {
+            postgres: {
+              sql: `select "first_name", "p"."name" from "person" cross join lateral (select "name" from "pet" inner join "person" as "owner" on "owner"."id" = "pet"."owner_id" where "owner"."gender" = "person"."gender") as "p" order by "first_name", "p"."name"`,
+              parameters: [],
+            },
+            mysql: NOT_SUPPORTED,
```
igalklebanov, line 833: should also be supported in MySQL according to [spec](https://dev.mysql.com/doc/refman/8.4/en/join.html).

test/node/src/join.test.ts
```diff
@@ -711,6 +711,37 @@ for (const dialect of DIALECTS) {
...
+          .selectAll()
+          .orderBy('person.first_name')
+
+        testSql(query, dialect, {
+          postgres: {
+            sql: `select * from "person" cross join "pet" order by "person"."first_name"`,
+            parameters: [],
+          },
+          mysql: {
+            sql: `select * from "person" cross join "pet" order by "person"."first_name"`,
```
igalklebanov, line 728:
```suggestion
            sql: 'select * from `person` cross join `pet` order by `person`.`first_name`',
```

PR #1335, opened 2025-01-28
feat: Add disableTransactions option to Migrator

### LINKED ISSUE #352: feature request: non-transactional DDL for concurrent index creation
👋 First just wanted to say this project looks awesome! Pretty close to what I wish Knex was if I started it now, with the affordances of modern TypeScript instead of 10 years ago when we were stuck in ES5 and Promises weren't even really an official language construct yet.

I started to play around with it a bit and couldn't find a way to do non-transactional DDL, though there's a chance I was missing something. This is necessary when running statements like `CREATE INDEX CONCURRENTLY` which [cannot be run inside a transaction](https://www.postgresql.org/docs/current/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY) in PostgreSQL.

Knex does this with a `config` property that can be set in a transaction file:

https://knexjs.org/guide/migrations.html#transactions-in-migrations

```ts
export const config = { transaction: false }
```

Wanted to see if this was something on the radar or if you'd be open to a similar API for configuring per-file migrations, or another approach that'd offer similar support.

reidswan:
## Summary 

Adds `disableTransactions` option to the Migrator.

This flag controls whether the migrator will run a migration in a transaction or not, for dialects that support a transactional DDL. 
   
When true, migrations will not be run in a transaction, even if the dialect supports transactions in their DDL.

When false, migrations will be run in a transaction if that is supported by the dialect.

## Testing

- Added unit tests to `test/node`.
- Tested manually by creating a postgres migration to `CREATE INDEX CONCURRENTLY` on a big table. This succeeded when `disableTransactions` was true, and failed (as expected) when `disableTransactions` was false.

Closes #352

### DIFF
--- a/src/migration/migrator.ts
+++ b/src/migration/migrator.ts
@@ -824,7 +824,14 @@ export interface MigratorProps {
   readonly nameComparator?: (name0: string, name1: string) => number
 
   /**
-   *
+   * Whether to skip running the migration in a transaction. Only relevant to databases
+   * that support a transactional DDL (i.e. postgres and mssql).
+   * 
+   * When true, migrations will not be run in a transaction, even if the dialect 
+   * supports transactions in their DDL.
+   * 
+   * When false, migrations will be run in a transaction if that is supported by
+   * the dialect.
    */
   readonly disableTransactions?: boolean
 }
--- a/src/migration/migrator.ts
+++ b/src/migration/migrator.ts
@@ -523,7 +523,7 @@ export class Migrator {
       }
     }
 
-    if (adapter.supportsTransactionalDdl) {
+    if (adapter.supportsTransactionalDdl && !this.#props.disableTransactions) {
       return this.#props.db.transaction().execute(run)
     } else {
       return this.#props.db.connection().execute(run)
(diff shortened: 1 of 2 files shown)


src/migration/migrator.ts
```diff
@@ -822,6 +822,18 @@ export interface MigratorProps {
...
+
+  /**
+   * Whether to skip running the migration in a transaction. Only relevant to databases
+   * that support a transactional DDL (i.e. postgres and mssql).
+   * 
+   * When true, migrations will not be run in a transaction, even if the dialect 
+   * supports transactions in their DDL.
+   * 
+   * When false, migrations will be run in a transaction if that is supported by
+   * the dialect.
```
igalklebanov, line 834:
```suggestion
   * When `true`, don't run migrations in transactions even if the dialect supports transactional DDL.
   *
   * Default is `false`.
   *
   * This is useful when some migrations include queries that would fail otherwise.
```

igalklebanov (approved):
LGTM! 🚀 

Thanks! 🙏

PR #1338, opened 2025-02-02
feat: add `SafeNullComparisonPlugin` plugin

rafaelalmeidatk:
As mentioned on [this Discord thread](https://discord.com/channels/890118421587578920/1312819986061983876) I migrated a big project from knex and one of the biggest blocks was the fact that kysely didn't automatically handle nullable types. The behavior is understandable since the library follows more of a WYSIWYG concept, but changing searching and updating all the queries would make the migration unpractical so I thought this could be a good plugin to have

### DIFF
--- a/src/plugin/safe-null-comparison/safe-null-comparison-transformer.ts
+++ b/src/plugin/safe-null-comparison/safe-null-comparison-transformer.ts
@@ -1,0 +1,47 @@
+import { BinaryOperationNode } from '../../operation-node/binary-operation-node.js'
+import { OperationNodeTransformer } from '../../operation-node/operation-node-transformer.js'
+import { OperatorNode } from '../../operation-node/operator-node.js'
+import { ValueNode } from '../../operation-node/value-node.js'
+
+export class SafeNullComparisonTransformer extends OperationNodeTransformer {
+  protected transformBinaryOperation(
+    node: BinaryOperationNode,
+  ): BinaryOperationNode {
+    const { operator, leftOperand, rightOperand } =
+      super.transformBinaryOperation(node)
+
+    if (
+      !ValueNode.is(rightOperand) ||
+      rightOperand.value !== null ||
+      !OperatorNode.is(operator)
+    ) {
+      return node
+    }
+
+    const op = operator.operator
+    if (op !== '=' && op !== '!=' && op !== '<>') {
+      return node
+    }
+
+    return BinaryOperationNode.create(
+      leftOperand,
+      OperatorNode.create(op === '=' ? 'is' : 'is not'),
+      rightOperand,
+    )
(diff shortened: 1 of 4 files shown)


noor-tg:
Can you make it standalone package ?

site/docs/plugins.md
```diff
@@ -25,3 +25,7 @@ A plugin that removes duplicate joins from queries. You can read more about it i
 ### Handle `in ()` and `not in ()` plugin
 
 A plugin that allows handling `in ()` and `not in ()` with a chosen strategy. [Learn more](https://kysely-org.github.io/kysely-apidoc/classes/HandleEmptyInListsPlugin.html).
+
+### Safe null comparison plugin
+
+A plugin that automatically converts `=`, `!=` and `<>` to the equivalent `is` and `is not` predicates depending on the value of the variable. [Learn more](https://kysely-org.github.io/kysely-apidoc/classes/SafeNullComparisonPlugin.html).
```
igalklebanov:
```suggestion
A plugin that converts `=`, `!=` and `<>` to the equivalent `is` and `is not` operators when the right-hand side value is `null`. [Learn more](https://kysely-org.github.io/kysely-apidoc/classes/SafeNullComparisonPlugin.html).
```

src/plugin/safe-null-comparison/safe-null-comparison-transformer.ts
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    const op = operator.operator
+    if (op !== '=' && op !== '!=' && op !== '<>') {
+      return node
+    }
+
+    return BinaryOperationNode.create(
+      leftOperand,
+      OperatorNode.create(op === '=' ? 'is' : 'is not'),
```
igalklebanov, line 29:
```suggestion
    const op = operator.operator
    
    if (op !== '=' && op !== '!=' && op !== '<>') {
      return node
    }

    return BinaryOperationNode.create(
      leftOperand,
      OperatorNode.create(op === '=' ? 'is' : 'is not'),
```

test/node/src/safe-null-comparison-plugin.test.ts
```diff
@@ -0,0 +1,244 @@
...
+  destroyTest,
+  initTest,
+  TestContext,
+  testSql,
+  insertDefaultDataSet,
+  DIALECTS,
+} from './test-setup.js'
+
+for (const dialect of DIALECTS) {
+  describe(`${dialect}: safe null comparison`, () => {
```
igalklebanov:
```suggestion
for (const dialect of DIALECTS) {
  describe(`${dialect.variant}: safe null comparison`, () => {
```

PR #1342, opened 2025-02-06
feat: Allow read-only transactions in Postgres and MySQL

### LINKED ISSUE #1341: Support read only transactions
Postgres and MySQL allow to start a transaction as read-only, can we have a way to do the same via kysely? e.g.

```sql
start transaction isolation level read committed read only
```

https://www.postgresql.org/docs/current/sql-start-transaction.html
https://dev.mysql.com/doc/refman/8.4/en/commit.html

(I'll be happy to PR this myself)

---

I am slowly migrating MikroORM to Kysely, I am actually almost there, and this is one of the few final issues I cannot workaround easily (without a custom dialect with lots of unnecessary overrides).

B4nan:
Closes #1341

### DIFF
--- a/src/dialect/mysql/mysql-driver.ts
+++ b/src/dialect/mysql/mysql-driver.ts
@@ -73,15 +73,15 @@ export class MysqlDriver implements Driver {
     connection: DatabaseConnection,
     settings: TransactionSettings,
   ): Promise<void> {
-    if (settings.isolationLevel || settings.readOnly) {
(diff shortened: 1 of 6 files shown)


B4nan:
@igalklebanov should be ready for review now

igalklebanov (approved):
LGTM! 🚀 

Thanks! 🙏 

Added an `AccessMode` type and validation, extracted validation elsewhere. Rearranged/simplified/aligned transaction characteristics test cases between classic and controlled transaction suites.

PR #1344, opened 2025-02-06
fix: isAutoIncrementing: true for sqlite integer primary key

tgriesser:
Fixes: https://github.com/RobinBlomberg/kysely-codegen/issues/105

**Summary:**
- If no `AUTOINCREMENT` is not detected for a table, additionaly checks for `INTEGER PRIMARY KEY`, which sqlite will use to determine whether to autoincrement by default: https://www.sqlite.org/autoinc.html

**Additional Refactor:**
- Changed the `#getTableMetadata` function to execute a total of 2 queries while introspecting the sqlite DB:

**Previously:**

```sql
select name from sqlite_master where type in ('table', 'view') and ...

-- The following 2 queries are repeated for N number of tables returned in first query:
select sql, type from sqlite_master where name = $table
select name, type, notnull, dflt_value from pragma_table_info($table)
```

**Now:**

```sql
select name, sql, type from sqlite_master where type in ('table', 'view') and ...

-- single union for all the columns
select 
  $table as table, name, type, notnull, dflt_value from pragma_table_info($table) 
union all 
  $table2 as table, name, type, notnull, dflt_value from pragma_table_info($table) 
union all ...  
order by table, cid
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -163,6 +163,6 @@ PR's description. E.g. `closes #123` will link the PR to issue/pull request #123
 
 1. install docker.
 
-1. run `docker-compose up` in your terminal to spin up database instances.
+1. run `docker compose up` in your terminal to spin up database instances.
 
 1. run `npm test` in another terminal to run tests.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "kysely",
-  "version": "0.27.4",
+  "version": "0.27.5",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "kysely",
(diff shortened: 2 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -163,6 +163,6 @@ PR's description. E.g. `closes #123` will link the PR to issue/pull request #123
 
 1. install docker.
 
-1. run `docker-compose up` in your terminal to spin up database instances.
+1. run `docker compose up` in your terminal to spin up database instances.
```
igalklebanov, line 166: 🙏

igalklebanov (approved):
LGTM! 🚀 

Thanks! 🙏 

Added a test case, and re-arranged types a bit. 🤜 🤛

Issue #1345, opened 2025-02-06
Support `.limit(Infinity)` or `.limit(null)`  for an unlimited query


everett1992:
I wrote a function with an optional limit parameter. I want the query to be limited if the option is set, and without limit otherwise.

My first instinct was to write ` .limit(limit ?? Infinity)` which makes the intention clear, is chainable, and passes type checking, but it threw SqliteError datatype mismatch from sqlite-driver.

I found I could use a let binding and imperative code, or $if. I like the $if code but it has a performance warning.
I don't like let binding, it introduces noise and requires you to name a local variable.

```
// Using $if
await db.selectFrom('people')
  .$if(limit != null, qb => qb.limit(limit!))
  .execute()
```
```
// using let
let qb = db.selectFrom('people')
if (limit != null) qb = qb.limit(limit)
await qb.execute()
```

Would you consider supporting `.limit(Infinity)` or `.limit(null)` as a no-op? The query builder would not add any LIMIT clause.

igalklebanov:
Hey 👋

This goes against the "What You See Is What You Get" core design principle. Kysely outputs the SQL you want it to. Giving it JS-only values or omitting clauses based on values under the hood is not the way.

You could implement a plugin that removes limit nodes if the value is `null` or `Infinity`.

igalklebanov:
Actually, PostgreSQL supports `limit all` and `limit null`, so extending the value to `'all'` and/or `null` is a reasonable enhacement idea.

PR #1347, opened 2025-02-08
feat: expands limit in select accepting null value

alenap93:
Expands limit in select accepting null value. This PR includes tests.

This PR doesn't add _limit ALL_ : https://github.com/brianc/node-postgres/issues/2673

Closes:
#1345


igalklebanov (approved):
LGTM! 🚀 

Thanks! 🙏 

Added some text about this in the relevant jsdoc and moved the test case closer to other `limit` test cases.

PR #1349, opened 2025-02-08
refactor: improve sqlite-introspector metadata query

tgriesser:
Minor change to follow up on #1344	

The change to using a `union all` means that although there's now only one query, the query grows to be as long as the number of tables in the DB, which feels a little off.

Decided to see if there was a way to handle this that was a little cleaner and came up with:

```sql
with "table_list" as (
  select "name", "sql", "type" from "sqlite_master"
  where "type" in ('table', 'view')
  and "name" not like 'sqlite_%' order by "name"
)
select
  "tl"."name" as "table",
  "p"."cid",
  "p"."name",
  "p"."type",
  "p"."notnull",
  "p"."dflt_value",
  "p"."pk"
from
  "table_list" as "tl",
  pragma_table_info(tl.name) as "p"
order by "tl.name", "p.cid"
```

Which feels a little nicer than an really large union query

### DIFF
--- a/src/dialect/sqlite/sqlite-introspector.ts
+++ b/src/dialect/sqlite/sqlite-introspector.ts
@@ -11,6 +11,7 @@ import {
   DEFAULT_MIGRATION_TABLE,
 } from '../../migration/migrator.js'
 import { sql } from '../../raw-builder/sql.js'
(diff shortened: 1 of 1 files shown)


igalklebanov (approved):
LGTM! 🚀 

Thanks! 🙏

Issue #509, opened 2023-05-21
Support `create table...index/key`.


igalklebanov:
Hey 👋 

Currently we don't support adding indexes in `CreateTableBuilder`.
This is supported in MySQL, and other non-builtin dialects.

Also, we should align with #507 & #508 if the proposed enhancements are introduced.

PR #1352, opened 2025-02-10
feat: add addIndex to CreateTableBuilder

alenap93:
This PR adds addIndex to CreateTableBuilder. Now is possible to create index inside create table (for dialects like mysql).

Closes:
#509

PS: If you could suggest me a better name for new node, review this 😉 .


igalklebanov:
Hey 👋 

Sorry for the delay! Thanks! 🙏 

Pushed it through. Re-using the existing `AddIndexNode`. Good stuff.

PR #1355, opened 2025-02-12
document that `refTuple` doesn't check or escape inputs

AlexErrant:


koskimas:
The values are references. If you don't ignore the types, you can't pass in arbitrary strings. If you ignore the types, even `ref` has the same vulnerability. Don't ignore the types.

AlexErrant:
I'm going to make a noise of protest. This is the same documentation that's on [`ref`](https://github.com/kysely-org/kysely/blob/deedd28c4c422185e4ec79f6da5432bc263ec7fd/src/raw-builder/sql.ts#L152-L153), [`table`](https://github.com/kysely-org/kysely/blob/deedd28c4c422185e4ec79f6da5432bc263ec7fd/src/raw-builder/sql.ts#L204-L205), [`id`](https://github.com/kysely-org/kysely/blob/deedd28c4c422185e4ec79f6da5432bc263ec7fd/src/raw-builder/sql.ts#L245-L246), [`lit`](https://github.com/kysely-org/kysely/blob/deedd28c4c422185e4ec79f6da5432bc263ec7fd/src/raw-builder/sql.ts#L281-L284), and [`raw`](https://github.com/kysely-org/kysely/blob/deedd28c4c422185e4ec79f6da5432bc263ec7fd/src/raw-builder/sql.ts#L312-L313). Why do these methods deserve an all caps `WARNING!` but not `refTuple`? If anything, `refTuple` deserves it _more_ because it's easily overlooked!

Also, the fact that you intentionally omit this warning implies it is safe since other dangerous methods are documented as dangerous. Having _inconsistent_ security docs is not good.

Tagging @igalklebanov because he +1ed your comment and I _**strongly**_ disagree.

igalklebanov:
Hey 👋 

Generally, people should avoid ignoring the type-system. I agree 100% with @koskimas on that. I also don't like the way we just spam the same text everywhere. And there are a lot of these reference methods that might also need that text - not great as far as maintainability goes.

But, `Kysely<any>` is a thing. It is being used in migrations, seeds, in very dynamic scenarios (e.g. in a SQL GUI application that manages our users' users' schema and queries the db).
Also, there's no harm in adding another warning text, if we're already doing the same in other places.

An ambitious task would be to create a system of JSDoc injection, that would allow us to DRY this up. Is it worth the complexity? I dunno.

AlexErrant:
To be clear: I am _not_ ignoring the type system. I've found an edge case in a framework where there's typesafety _without_ input validation across a server-browser communication layer. At least, not explicitly. Excuse me for being intentionally vague here. I'm _not_ doing `@ts-ignore`, `any`, etc. and can still run into this issue.

> I also don't like the way we just spam the same text everywhere. And there are a lot of these reference methods that might also need that text - not great as far as maintainability goes.

In [one location](https://github.com/kysely-org/kysely/blob/deedd28c4c422185e4ec79f6da5432bc263ec7fd/src/raw-builder/sql.ts#L281-L284) it's not the same text - it adds "_You almost always want to use normal substitutions that get sent to the db as parameters._" I'm not sure an effort to DRY this up is worthwhile.

That said, I do think a central location where the dangerous methods are listed/documented makes sense.

Separately, I was thinking about making an ESLint rule for warning when passing a variable to a dangerous method and not the usual string literal:

```ts
const myTable = "myTable" as const
qb.table(myTable) // yields eslint error

qb.table("myTable") // no eslint error
```

Shouldn't be too hard (famous last words).

PR #1379, opened 2025-03-06
Redundant export UpdateValuesNode removed

Ciantic:


koskimas:
@igalklebanov You've marked this as breaking since someone might've used the exported type in userland code? I think it's highly unlikely since it's not used internally anywhere.

igalklebanov:
We've been exporting it. 

I haven't found anyone using it yet on my searches. There's nothing blocking it from being merged. Just wanted to make sure we list it in the breaking changes list.

PR #1392, opened 2025-03-16
feat: escape single quotes in string literals.

### LINKED ISSUE #1124: Escape string literals by default
I believe Kysely should escape string literals by default, or at least provide an option to handle this. I brought this up [before](https://github.com/kysely-org/kysely/issues/485), and escaping was added to identifiers, but I don't believe it was applied to literals.

igalklebanov:
Hey :wave:

closes #1124.

This PR introduces some string literal SQL injection denial in the form of escaping `'` to deny the most common string literal attacks:

```sql
-- when sql.lit(`'; drop table users --`)
where column = ''; drop table users --'
-- when sql.lit(`' or 1=1 --`)
where column = '' or 1=1 --'
```

and make them:

```sql
-- when sql.lit(`'; drop table users --`)
where column = '''; drop table users --'
-- when sql.lit(`' or 1=1 --`)
where column = ''' or 1=1 --'
```

If you find any additional measures we could take, please submit an issue OR swing by the Discord.
Some things are out of bounds, like SQL parsing, semantic checks, etc.

### DIFF
--- a/src/query-compiler/default-query-compiler.ts
+++ b/src/query-compiler/default-query-compiler.ts
@@ -118,6 +118,8 @@ import { CollateNode } from '../operation-node/collate-node.js'
 import { QueryId } from '../util/query-id.js'
 import { RenameConstraintNode } from '../operation-node/rename-constraint-node.js'
 
+const LIT_WRAP_REGEX = /'/g
+
 export class DefaultQueryCompiler
   extends OperationNodeVisitor
   implements QueryCompiler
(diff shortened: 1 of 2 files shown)


igalklebanov:
Need to handle strings with `'` already being escaped.
Possibly find a way to opt-out of sanitation.

koskimas (approved):
LGTM

igalklebanov:
@koskimas I wonder if we should add an easy opt-out mechanism in case we did something wrong, and until we fix it. The cheapest (no API changes to dialects and compilers) would be a Kysely-specific environment variable check - but that might be inconvenient to Deno users (might make `--allow-env` a requirement suddenly).

I'm also unsure about the "already escaped literals" problem. Is it a problem? what if the literal is not escaped, just has a bunch of `'`s grouped together? Can we tell?

PR #1395, opened 2025-03-19
fix: allow empty array for sql.join

Sealos:
We are trying to support using ANY() with empty arrays such as in this code:
```ts
if (filter.excludedOperatorIds) {
    query = query.where(sqlKy`NOT operator_id = ANY(ARRAY[${sqlKy.join(filter.excludedOperatorIds)}])`);
}

if (filter.includedOperatorIds) {
    query = query.where(sqlKy`operator_id = ANY(ARRAY[${sqlKy.join(filter.includedOperatorIds)}])`);
}
```

While this code works for any non-empty array, it crashes as the join fails calling a new Array(-1).

This change should maintain the current behavior while supporting empty arrays. If this PR is valid, I can add a test for it.


dsonet:
This seems to be an invalid use.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Array
image

igalklebanov (approved):
LGTM! 🚀

Issue #1425, opened 2025-04-15
Postgres introspector fails on mixed-case table names as of 0.28.0


neil176:
Hello, we are using the postgres introspector to generate types for kysely and have encountered an issue trying to upgrade to 0.28.0.

The error surfaces as something like `relation public.mytable does not exist` in the case of a table `public.MyTable` existing. 

The issue is specific to databases containing mixed case table names, and is caused by this line of sql:

https://github.com/kysely-org/kysely/blob/a43b270cb6c88208d1d3f79b958e774deebe63f2/src/dialect/postgres/postgres-introspector.ts#L63-L67

I believe the issue is needing to insert some extra `"` characters to wrap schema and then table, as the function is somewhat peculiar:

https://www.postgresql.org/docs/current/functions-info.html#FUNCTIONS-INFO-CATALOG
>pg_get_serial_sequence ( table text, column text ) → text
>Returns the name of the sequence associated with a column, or NULL if no sequence is associated with the column. If the column is an identity column, the associated sequence is the sequence internally created for that column. For columns created using one of the serial types (serial, smallserial, bigserial), it is the sequence created for that serial column definition. In the latter case, the association can be modified or removed with ALTER SEQUENCE OWNED BY. (This function probably should have been called pg_get_owned_sequence; its current name reflects the fact that it has historically been used with serial-type columns.) The first parameter is a table name with optional schema, and the second parameter is a column name. Because the first parameter potentially contains both schema and table names, it is parsed per usual SQL rules, meaning it is lower-cased by default. The second parameter, being just a column name, is treated literally and so has its case preserved. The result is suitably formatted for passing to the sequence functions (see [Section 9.17](https://www.postgresql.org/docs/current/functions-sequence.html)).

I can get to work on a PR for this or if it's convenient for someone more familiar with contributing here I would appreciate it.

koskimas:
Oof that's bad 😬 Thank you so much for raising this and for the fix! I'll merge that as soon as you have a test case done and release a patch version.

koskimas:
0.28.1 released

neil176:
Thank you for the quick replies, help with the testing, and fast release here! We really appreciate it and are excited to upgrade - the transaction functionality has been a much-requested feature internally and the `IN ()` plugins will be super helpful with converting some code over from prisma!

PR #1426, opened 2025-04-15
pg introspector - Wrap schema.table in double quotes for case handling

neil176:
Fixes https://github.com/kysely-org/kysely/issues/1425

Does so by wrapping the namespace and relname with the builtin `quote_ident` function to match the behavior of  `pg_get_serial_sequence` lowercasing by default.

Adds a test assertion to the introspector for a `some_table.MixedCaseTable` that I confirmed fails without the fix.

### DIFF
--- a/src/dialect/postgres/postgres-introspector.ts
+++ b/src/dialect/postgres/postgres-introspector.ts
@@ -62,7 +62,7 @@ export class PostgresIntrospector implements DatabaseIntrospector {
         ),
         sql<
           string | null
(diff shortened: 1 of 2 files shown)


neil176:
@koskimas  I am having trouble with starting the mssql container to run the tests while working on them locally (it's probably an issue with my docker vm). Do you know a way to only run the postgres tests?

igalklebanov (approved):
LGTM! 🚀

PR #1435, opened 2025-04-21
fix: SQLite's introspector is printing deprecation errors for `orderBy(array)`.

### LINKED ISSUE #1424: After upgrading to v0.28 when I run a migration to create the database schema Kysely displays a deprecation warning.
After upgrading to v0.28 when I run a migration to create the database schema Kysely displays a deprecation warning.
```
"orderBy(array) is deprecated, use multiple orderBy calls instead."
```
This is the query that seems to be throwing the warning:
```
[2025-06-14 19:06:29] DEBUG: select "name", "sql", "type" from "sqlite_master" where "type" in (?, ?) and "name" not like ? order by "name"
[2025-06-14 19:06:29] DEBUG: [ 'table', 'view', 'sqlite_%' ]
```
This seems to be a query that is internal to Kysely's migration code and that I have no control over.

Additional information:
I'm using the following packages:
```
"kysely": "0.28.0",
"better-sqlite3": "11.9.1",
```

igalklebanov:
Hey :wave:

closes #1424

Sorry about that. Annoying.

### DIFF
--- a/src/query-builder/order-by-interface.ts
+++ b/src/query-builder/order-by-interface.ts
@@ -45,24 +45,6 @@ export interface OrderByInterface<DB, TB extends keyof DB, O> {
    * order by "id", "fn" desc
    * ```
    *
-   * Multiple columns/expressions per call:
-   *
-   * ```ts
-   * await db
-   *   .selectFrom('person')
-   *   .select('person.first_name as fn')
-   *   .orderBy(['id', 'fn'])
-   *   .execute()
-   * ```
-   *
-   * The generated SQL (PostgreSQL):
-   *
-   * ```sql
-   * select "person"."first_name" as "fn"
-   * from "person"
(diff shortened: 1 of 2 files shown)


koskimas (approved):

src/query-builder/order-by-interface.ts
```diff
@@ -45,24 +45,6 @@ export interface OrderByInterface<DB, TB extends keyof DB, O> {
...
-   * ```
-   *
-   * The generated SQL (PostgreSQL):
-   *
-   * ```sql
-   * select "person"."first_name" as "fn"
-   * from "person"
-   * order by "id", "fn"
-   * ```
-   *
```
igalklebanov, line 65: we don't want to encourage this deprecated overload's usage. it is a lot slower at compile-time.

PR #1436, opened 2025-04-22
ensure jsdocs check's dts is source of truth for docs examples.

igalklebanov:
Hey 👋 

So we are testing our jsdocs code examples, and made changes to the examples. That's awesome.
To get that up and running, we created a `d.ts` file that contains all the types needed for our jsdocs examples.

I forgot to also make sure the docs site's examples' types are aligned with this new `d.ts` file.

This PR makes it so that the check's `d.ts` file is auto generated as the docs site's playground types.

Without these changes, the docs site's examples are broken and show lots of type errors.


koskimas (approved):
LGTM!

PR #1463, opened 2025-05-17
fix(mssql): socket error handling during connect.

igalklebanov:


koskimas (approved):

src/util/deferred.ts
```diff
@@ -1,29 +1,31 @@
...
 
   get promise(): Promise<T> {
     return this.#promise
   }
 
-  resolve = (value: T | PromiseLike<T>): void => {
-    if (this.#resolve) {
-      this.#resolve(value)
-    }
+  reject = (reason?: any): void => {
```
koskimas, line 24: Why did you remove types from these?

src/util/deferred.ts
```diff
@@ -1,29 +1,31 @@
...
   #reject?: (reason?: any) => void
+  #resolve?: (value: T | PromiseLike<T>) => void
 
   constructor() {
     this.#promise = new Promise<T>((resolve, reject) => {
-      this.#reject = reject
-      this.#resolve = resolve
+      this.#reject = (reason?: any) => {
+        reject(reason)
+        this.#reject = this.#resolve = undefined
```
koskimas, line 11: Should you clear #reject and #resolve before the `reject(reason)` call? I don't really remember how promises worked in this situation, but can `reject(reason)` trigger code that calls `this.reject` or `this.resolve`?

belgattitude:
Thank you for that one ☝️

Issue #1401, opened 2025-03-24
Values(***) should handle empty array


pankajaksha:
while inserting a array , the values column throws SQL error .
Shouldn't it handle this empty array ?

await this.postgresClient
      .withSchema(this.schema)
      .insertInto(this.table)
      .values(entities) <-- throws error is entities is an empty array
      .execute()

igalklebanov:
Hey 👋 

It should not. You should not send a query in the first place (check `entities.length`). Having Kysely not send it for you is against its design principles - Kysely builds the queries you want it to, and executes them.

PR #1473, opened 2025-05-28
Add non empty array check to insertInto

KROT47:
This should prevent runtime error when empty array is passed to `insertInto().values()` since now TS error will pop up.

Currently no TS error is thrown and this may lead to runtime error and this is bad.

It [was said](https://github.com/kysely-org/kysely/issues/1401) that end user should handle the check but it is easy to forget about it.

Maybe we also should export a util like `getNonEmptyArray` and update the docs. It will be a little complex but I think it is better than risk of runtime errors.

What do you think?

### DIFF
--- a/test/node/src/insert.test.ts
+++ b/test/node/src/insert.test.ts
@@ -1,4 +1,4 @@
-import { AliasedRawBuilder, InsertResult, Kysely, sql } from '../../../'
+import { AliasedRawBuilder, InsertObject, InsertResult, Kysely, sql } from '../../../'
 
 import {
   clearDatabase,
@@ -14,6 +14,7 @@ import {
   DIALECTS,
   limit,
 } from './test-setup.js'
+import { getNonEmptyArray } from './utils'
 
 for (const dialect of DIALECTS) {
   describe(`${dialect}: insert into`, () => {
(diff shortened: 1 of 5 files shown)


igalklebanov:
Hey 👋 

Thanks! 🙏

We've considered this in the past.

1. It is too big a breaking change.
2. It is a major DX downgrade for many people. TypeScript should be natively narrowing to `[T, ...T[]]` after a length check, it doesn't.
3. There's no indication how widespread the problem of forgetting to check array lengths is.
4. People who forget to check their array lengths will also forget to use a `[T, ...T[]]` helper.

I believe the *lesser evil* solution is in an opt-in mechanism, e.g., a plugin like [`HandleEmptyInListsPlugin`](https://kysely-org.github.io/kysely-apidoc/classes/HandleEmptyInListsPlugin.html), that allows to alter the query to `insert into <table> select * from <table> where 1 = 0;` that doesn't insert anything, as a strategy.

test/node/src/insert.test.ts
```diff
@@ -35,6 +36,36 @@ for (const dialect of DIALECTS) {
       await destroyTest(ctx)
     })
 
+    it('should throw ts and runtime error if empty array is passed to values', async () => {
```
igalklebanov, line 39: this is not the place to test compile-time. there are `tsd` suites for that.

Issue #482, opened 2023-05-11
jsonArrayFrom dates have Date type but are returned as a string


jvandenaardweg:
Hi there,

Thanks for the time and effort on this library, really enjoy working with it on my current project!

I do however have a question. I notice the `Date`'s retrieved by using `jsonArrayFrom` are returned as a `string` instead of a `Date`. The types expect i should get a `Date` back, but in fact i'm getting a `string`. Is this expected behaviour?

Due to not giving away my database architecture in the open, i've reduced it down to it's most basic form below.

Typescript sees both `createdAt` with a `Date` type. But only the `createdAt` from the `jsonArrayFrom` is actually a `string`, which is not what we expect.

```typescript
const a = await kysely
  .selectFrom('a')
  .select((eb) => [
    jsonArrayFrom(
      eb
        .selectFrom('b')
        .select([
          'b.createdAt',
        ])
        .whereRef('b.aId', '=', 'a.id'),
    ).as('b'),
  ])
  .select(['a.createdAt'])
  .where('id', '=', id)
  .executeTakeFirst();

console.log(
  'typeof a?.createdAt',
  typeof a?.createdAt,
  a?.createdAt,
);
// typeof a?.createdAt object 2023-05-02T21:57:15.466Z
// This is a Date, correct!

a?.b.forEach((b) =>
  console.log('typeof b.createdAt', typeof b.createdAt, b.createdAt),
);
// typeof b.createdAt string 2023-05-11 14:16:36.483000
// This is a string, not correct
```

The types:

```typescript
export type A = {
  id: Generated<number>;
  createdAt: Generated<Timestamp>;
};

export type B = {
  id: Generated<number>;
  aId: number;
  createdAt: Generated<Timestamp>;
};
```

Some context: i'm using Kysely with the Planetscale Dialect together with `prisma-kysely` to have the correct types.

edit: can confirm `jsonObjectFrom` is also having this same behaviour.

koskimas:
Yep, this is expected.

jvandenaardweg:
Ok, is it something that can be fixed? Either as a fix in this library or a workaround I can use? As it's not really type-safe here, which is a bit of a bummer

koskimas:
JSON is JSON. There's no `Date` type in JSON. There's nothing we can do. The word `JSON` is in the name of the function, so you shouldn't be that surprised.

jvandenaardweg:
[...]

PR #1474, opened 2025-05-29
feat: add Date type cast to string for json objects

KROT47:
Kinda not very nice when receiving date types for aggregated json objects when they are not dates but strings. Leads to runtime errors...

This PR seems reasonable for me, any comments and updates are appreciated)

Inspired from [this issue](https://github.com/kysely-org/kysely/issues/482) and bad DX

P.S. love kysely, thanks guys!

BTW check out KyselyJson, it is very useful for me in my projects, what do you think?

### DIFF
--- a/src/helpers/mssql.ts
+++ b/src/helpers/mssql.ts
@@ -1,7 +1,7 @@
 import { Expression } from '../expression/expression.js'
 import { RawBuilder } from '../raw-builder/raw-builder.js'
 import { sql } from '../raw-builder/sql.js'
-import { Simplify } from '../util/type-utils.js'
+import { CastDatesToStrings, Simplify } from '../util/type-utils.js'
 
 /**
  * An MS SQL Server helper for aggregating a subquery into a JSON array.
@@ -77,7 +77,7 @@ import { Simplify } from '../util/type-utils.js'
  */
 export function jsonArrayFrom<O>(
   expr: Expression<O>,
-): RawBuilder<Simplify<O>[]> {
+): RawBuilder<CastDatesToStrings<Simplify<O>>[]> {
   return sql`coalesce((select * from ${expr} as agg for json path, include_null_values), '[]')`
 }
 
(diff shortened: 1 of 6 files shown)


igalklebanov:
Hey 👋 

Thanks! 🙏 

Quoting the issue you mentioned:

> Recursively traversing the output type and transforming all non-JSON properties to JSON would be terribly slow and it would be done for all outputs of those JSON functions, not just the problematic ones. We could destroy typescript performance for everyone that uses those functions.
> 
> Since the typescript compiler is super slow, we need to constantly balance between type-safety and keeping IDEs and your compilation step fast. If I fixed this for you, a bunch of other users would come here pissed off demanding the opposite fix.
> 
> If someone is able to implement the tranformation AND is able to provide proof through comprehensive tests that the performance is good, I'd merge that PR.


IDE-crashing slow types are worse DX than something with an easy workaround.

This is not it, for now.

KROT47:
> > Recursively traversing the output type and transforming all non-JSON properties to JSON would be terribly slow and it would be done for all outputs of those JSON functions, not just the problematic ones. We could destroy typescript performance for everyone that uses those functions.

Yep, recursive is bad! But as you can see there is no recursive traversing in kysely types. There is only exported until which does recursive stuff for end users only!

@igalklebanov please, check PR again

igalklebanov:
Thanks for the follow-up.

This situation is a clear example of why our contribution guidelines—specifically discussing the implementation in the issue and getting assigned before submitting code—are essential.

Work on a solution had already begun internally based on an agreed-upon approach. Since your PR was submitted without following the required process, we were unable to collaborate on it and have continued with the original plan.

Following the established workflow is a requirement for all contributors. It's the only way we can effectively coordinate efforts and ensure we're building the library in a consistent and maintainable way.

I appreciate the time you invested, and I hope this clarifies the process for any future contributions you might consider.

---

This response was generated using Gemini 2.5 Pro

PR #1477, opened 2025-06-02
fix(helpers): JSON helpers' return types incorrectly have `Date`s and other JS/Node-native instances that require data type metadata.

### LINKED ISSUE #482: jsonArrayFrom dates have Date type but are returned as a string

igalklebanov:
Hey :wave:

closes #482.

We're going for a shallow approach - no recursion:

1. Recursion would be very expensive at compile time.

1. JSON results of nested JSON helpers should already be covered by their return type calculation.

1. If you have JSON columns, you should have already defined them properly - no JS/Node native, JSON invalid, types in their `SelectType`.

1. PostgreSQL arrays are in the scope of this change. e.g. a `timestamp[]` column would be `string[]`, not `Date[]`.

Typings test seems to be running for same amount of time. 😬 

Will add some more cases tomorrow.

### DIFF
--- a/src/helpers/mssql.ts
+++ b/src/helpers/mssql.ts
@@ -1,7 +1,11 @@
 import { Expression } from '../expression/expression.js'
 import { RawBuilder } from '../raw-builder/raw-builder.js'
 import { sql } from '../raw-builder/sql.js'
-import { Simplify } from '../util/type-utils.js'
+import {
+  ShallowDehydrateObject,
+  ShallowDehydrateValue,
+  Simplify,
+} from '../util/type-utils.js'
 
 /**
  * An MS SQL Server helper for aggregating a subquery into a JSON array.
@@ -77,7 +81,7 @@ import { Simplify } from '../util/type-utils.js'
  */
 export function jsonArrayFrom<O>(
   expr: Expression<O>,
-): RawBuilder<Simplify<O>[]> {
+): RawBuilder<Simplify<ShallowDehydrateObject<O>>[]> {
   return sql`coalesce((select * from ${expr} as agg for json path, include_null_values), '[]')`
 }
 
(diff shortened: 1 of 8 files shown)


test/ts-benchmarks/select-from.bench.ts
igalklebanov, line 1: these might have worsened before this PR.

PR #1480, opened 2025-06-05
fix(Migrator): Enable Transactional Dry-Runs by supporting Transaction Instances with disableTransactions: true

jlucaso1:
Hey team!

This PR addresses an issue where the `Migrator` would unexpectedly fail if initialized with a Kysely `Transaction` instance (`trx`) while `disableTransactions` was set to `true`. The Migrator was attempting to call `trx.connection()`, which isn't a supported operation on a `Transaction` object.

**The Problem:**

When `disableTransactions` is true, the `Migrator` falls back to using `this.#props.db.connection().execute(...)`. If `this.#props.db` *is* a `Transaction` instance, this leads to:
`Error: calling the connection method for a Transaction is not supported`

This prevented a useful pattern: performing a "transactional dry-run" of migrations, where migrations are executed within an outer transaction that is subsequently rolled back.

**The Solution:**

This change modifies the `Migrator` to check if the provided `db` instance `isTransaction`. If it is, and `disableTransactions` is also `true`, the `Migrator` will now directly use the provided `Transaction` instance (`this.#props.db`) to run the migration logic (`run(this.#props.db)`), bypassing the problematic `.connection()` call.

**Use Case Example (Transactional Migration Validation):**

This fix makes it much easier to validate migrations without permanently altering the database, which is super handy for pre-push hooks or CI checks:

```typescript
import { db } from './your-db-setup'; // Your main Kysely instance
import { Migrator, FileMigrationProvider } from 'kysely';

async function validateMigrationsDryRun() {
  console.log('Attempting transactional dry-run of migrations...');
  try {
    await db.transaction().execute(async (trx) => {
      const migrator = new Migrator({
        db: trx, // Pass the transaction
        provider: new FileMigrationProvider(...),
        disableTransactions: true, // Tell Migrator not to manage its own TX
      });

      const { error, results } = await migrator.migrateToLatest();

      if (error) {
        console.error('Migration validation failed:', error);
        throw error; // This will cause the outer transaction to rollback
      }

      console.log('Migrations appear valid!');
      // IMPORTANT: We must throw an error to force the rollback for a dry run
      throw new Error('DRY_RUN_ROLLBACK');
    });
  } catch (e: any) {
    if (e.message === 'DRY_RUN_ROLLBACK') {
      console.log('Dry run successful, all changes rolled back.');
    } else {
      console.error('Migration validation encountered an error:', e);
      // Potentially exit with error for CI
    }
  }
}
```

**Testing:**
I've added a test case to cover this specific scenario, ensuring the `Migrator` behaves as expected when used with a `Transaction` and `disableTransactions: true`.

### DIFF
--- a/src/migration/migrator.ts
+++ b/src/migration/migrator.ts
@@ -526,7 +526,12 @@ export class Migrator {
       }
     }
 
-    if (adapter.supportsTransactionalDdl && !this.#props.disableTransactions) {
+    if (this.#props.db.isTransaction && this.#props.disableTransactions) {
+      return run(this.#props.db)
+    } else if (
+      adapter.supportsTransactionalDdl &&
+      !this.#props.disableTransactions
+    ) {
       return this.#props.db.transaction().execute(run)
     } else {
       return this.#props.db.connection().execute(run)
--- a/test/node/src/migration.test.ts
+++ b/test/node/src/migration.test.ts
@@ -825,6 +825,55 @@ for (const dialect of DIALECTS) {
           expect(transactionSpy.called).to.be.false
         }
       })
+
+      it('should allow migrations to run inside a user-managed transaction when disableTransactions is true', async () => {
+        const migrationName = 'trx_connection_method_fail_case'
+        const executedUpMethodsForThisTest: string[] = []
+
+        const testSpecificProvider: any = {
+          getMigrations: async () => ({
+            [migrationName]: {
+              async up(_db: Kysely<any>): Promise<void> {
+                executedUpMethodsForThisTest.push(migrationName)
+              },
+              async down(_db: Kysely<any>): Promise<void> {},
+            },
+          }),
+        }
+
+        await ctx.db.transaction().execute(async (trx) => {
+          const migratorInTransaction = new Migrator({
+            db: trx,
+            provider: testSpecificProvider,
+            disableTransactions: true,
+          })
+
+          const { results, error } = await migratorInTransaction.migrateUp()
+
+          expect(error).to.be.undefined
+          expect(results).to.eql([
+            { migrationName, direction: 'Up', status: 'Success' },
+          ])
+        })
+
+        expect(executedUpMethodsForThisTest).to.eql(
+          [migrationName],
+          "The migration's 'up' method should have been executed.",
+        )
+
+        const checkerMigrator = new Migrator({
+          db: ctx.db,
+          provider: testSpecificProvider,
+        })
+
+        const migrations = await checkerMigrator.getMigrations()
+        const testMigrationInfo = migrations.find(
+          (m) => m.name === migrationName,
+        )
+
+        expect(testMigrationInfo).to.exist
+        expect(testMigrationInfo!.executedAt).to.be.instanceOf(Date)
+      })
     })
 
     describe('migrateDown', () => {


src/migration/migrator.ts
```diff
@@ -526,7 +526,12 @@ export class Migrator {
       }
     }
 
-    if (adapter.supportsTransactionalDdl && !this.#props.disableTransactions) {
+    if (this.#props.db.isTransaction && this.#props.disableTransactions) {
```
igalklebanov: When a transaction is passed, anything but running with it would fail. Regardless of `disableTransactions`.

```suggestion
    if (this.#props.db.isTransaction) {
```

test/node/src/migration.test.ts
```diff
@@ -825,6 +825,55 @@ for (const dialect of DIALECTS) {
           expect(transactionSpy.called).to.be.false
         }
       })
+
+      it('should allow migrations to run inside a user-managed transaction when disableTransactions is true', async () => {
```
igalklebanov:
```suggestion
      it('should run migrations using a provided transaction', async () => {
```

test/node/src/migration.test.ts
```diff
@@ -825,6 +825,55 @@ for (const dialect of DIALECTS) {
...
+          provider: testSpecificProvider,
+        })
+
+        const migrations = await checkerMigrator.getMigrations()
+        const testMigrationInfo = migrations.find(
+          (m) => m.name === migrationName,
+        )
+
+        expect(testMigrationInfo).to.exist
+        expect(testMigrationInfo!.executedAt).to.be.instanceOf(Date)
```
igalklebanov: Redundant. Already tested elsewhere.

```suggestion
```

test/node/src/migration.test.ts
```diff
@@ -825,6 +825,55 @@ for (const dialect of DIALECTS) {
...
+            disableTransactions: true,
+          })
+
+          const { results, error } = await migratorInTransaction.migrateUp()
+
+          expect(error).to.be.undefined
+          expect(results).to.eql([
+            { migrationName, direction: 'Up', status: 'Success' },
+          ])
+        })
```
igalklebanov: nitpicky

```suggestion
        const trx = await db.startTransaction().execute()
        
        const migratorInTransaction = new Migrator({
          db: trx,
          provider: testSpecificProvider,
        })

        const { results, error } = await migratorInTransaction.migrateUp()
        
        await trx.rollback().execute().catch(() => {})

        expect(error).to.be.undefined
        expect(results).to.eql([
          { migrationName, direction: 'Up', status: 'Success' },
        ])
```

igalklebanov (commented):
Hey 👋 

Thanks! 🙏 

Left some comments.

src/migration/migrator.ts
```diff
@@ -526,7 +526,12 @@ export class Migrator {
...
+      return run(this.#props.db)
```
igalklebanov, line 552:
```suggestion
      if (!adapter.supportsTransactionalDdl) {
        throw new Error('Transactional DDL is not supported in this dialect. Passing a transaction to this migrator would result in failure or unexpected behavior.')
      }

      return run(this.#props.db)
```

test/node/src/migration.test.ts
```diff
@@ -825,6 +825,55 @@ for (const dialect of DIALECTS) {
           expect(transactionSpy.called).to.be.false
         }
       })
+
+      it('should allow migrations to run inside a user-managed transaction when disableTransactions is true', async () => {
+        const migrationName = 'trx_connection_method_fail_case'
+        const executedUpMethodsForThisTest: string[] = []
+
+        const testSpecificProvider: any = {
```
igalklebanov:
```suggestion
        const testSpecificProvider: MigrationProvider = {
```

tarqwara:
Any updates on this?

PR #1482, opened 2025-06-07
Query cancelation

### LINKED ISSUE #783: Support query cancellation
I have some long-running analytical queries. At some point in the middle of the query running, the end user can express that they're no longer interested in getting more query results, by clicking a 'cancel' button or leaving the page. I'd like, at that point, to stop running the query in the underlying database, to save on compute costs. (For reference, my underlying database is Snowflake, which has pretty good [support for cancellation](https://docs.snowflake.com/en/developer-guide/node-js/nodejs-driver-execute#canceling-statements) in its driver.) 

In some of these cases, I'm reading the query results from Kysely in a streaming fashion; in other cases, I'm just waiting for a promise with the full result.

So, this request is to somehow support query cancellation in kysely, ideally for both streaming and non-streaming queries.

I don't really know what the API should look like for this in kysely, but...

- It seems like maybe the `DatabaseConnection` interface could be extended to provide an optional `cancelQuery` method
- For the non-streaming case, `execute()` and friends could accept an object argument like `{ signal: AbortSignal }`. If/when the signal is aborted before the query finishes, the underlying `cancelQuery` would be called.
- For the streaming case, there's already a way for the consumer of the iterator to signal disinterest in further results (with an explicit or implicit call to `iterator.return()`), but it would be nice for kysely to cancel the currently-running query at that time (whereas I think all that happens today is that the connection is released after the current batch of data finishes loading). It might also be nice for `stream()` to accept a signal too, for convenience, in case the caller already has a signal that they’re using to create one lifecycle for multiple different async operations (which could contain a mix of streaming and non-streaming queries).

Thoughts?

NB: edited to suggest AbortSignal instead (brain fart) and to clarify that this isn't limited to streaming.

elitan:
fixes #783


igalklebanov:
Hey 👋

Thank you for this contribution and the considerable effort that went into it. I genuinely appreciate you taking the time.

I'm going to close this PR for now, and I want to be transparent about why.

First, on a practical level, a pull request of this size (3k+ lines of code) that modifies core components is very difficult to review effectively. For a change this significant, the expectation is to first discuss the proposed implementation plan in the linked issue. This allows us to align on the approach and break the work into smaller, manageable pieces before a large amount of code is written.

On a more philosophical note, a core value of this project is to be a space where we can enjoy the craft of software development together—at a human pace, without deadlines or the pressure to "speedrun" a solution. We prioritize deliberate, manual exploration of the codebase, collaborative problem-solving, and learning from each other.

It seems like this PR may have been generated with the help of AI, and while that's a powerful tool, a large, automated submission can bypass the human-centric process we value here. Your comment, for instance, about needing "Knowledge of Kysely's architecture" as a prerequisite feels impersonal and runs counter to our goal of helping contributors learn the codebase collaboratively.

The best path forward is to use the existing issue as our home base for this.
If you'd still like to take this on, could you please comment on the issue and ask to be assigned? This helps us coordinate who is actively working on what.

Once you're assigned, let's move the detailed implementation discussion to our ⁠#contribution channel on Discord. Real-time chat is a much better place for the kind of back-and-forth brainstorming that will help us break this down into a series of small, focused PRs. 

Again, I truly appreciate the energy you put into this. I hope this explanation makes sense, and I look forward to collaborating with you in a way that aligns with the project's spirit.

---

This response was generated using Gemini 2.5 Pro

PR #1484, opened 2025-06-08
feat(core): ground work for query cancellation + PostgreSQL implementation.

igalklebanov:
Hey 👋 

part of #783 

This PR sets the foundations for dialects to implement query cancellation.
An `executable<O>` interface is created and all relevant builders implement it.
All `execute` and `stream` function accept an options object that might include `signal`.
This `signal` is passed down to the executor, and handled there. 
The connection receives a hint at query time that it might need to prepare itself for cancelation - e.g. in PostgreSQL you need to make a request to get the `pid` of the connection.
The connection has a `cancelQuery` method that, in dialects that support it, cancel the query on the database side.

Verifying this works with PostgreSQL.


src/driver/database-connection.ts
```diff
@@ -1,15 +1,21 @@
...
 export interface DatabaseConnection {
-  executeQuery<R>(compiledQuery: CompiledQuery): Promise<QueryResult<R>>
+  executeQuery<R>(
+    compiledQuery: CompiledQuery,
+    options?: ExecuteQueryOptions,
+  ): Promise<QueryResult<R>>
+
   streamQuery<R>(
     compiledQuery: CompiledQuery,
-    chunkSize?: number,
```
igalklebanov: optionality here was wrong; all streaming query builders always defaulted to `100`.

src/kysely.ts
```diff
@@ -538,20 +541,13 @@ export class Kysely<DB>
    *
    * See {@link https://github.com/kysely-org/kysely/blob/master/site/docs/recipes/0004-splitting-query-building-and-execution.md#execute-compiled-queries splitting build, compile and execute code recipe} for more information.
    */
-  executeQuery<R>(
+  async executeQuery<R>(
     query: CompiledQuery<R> | Compilable<R>,
-    // TODO: remove this in the future. deprecated in  0.28.x
-    queryId?: QueryId,
```
igalklebanov: breaking change following deprecation.

igalklebanov:
went for a more centralized approach in the latest commit. haven't implement for streaming yet. hopefully it makes sense there too. we want as little non-database related stuff happening in the drivers.

igalklebanov:
tests looking good for querying.. need to do streaming tests as well.

igalklebanov:
Found out `processID` exists on [`Client`](https://github.com/brianc/node-postgres/blame/master/packages/pg/lib/client.js#L298) but is undocumented. We can use it if exists instead of querying for it (slow).

Found out `cancel` exists on [`Client`](https://github.com/brianc/node-postgres/blame/master/packages/pg/lib/client.js#L460), but is undocumented. I attempted to use it, but realized there is no easy way to verify we're not canceling the wrong query after connecting.

Found out `Client` constructor exists on [`Pool`](https://github.com/brianc/node-postgres/blame/master/packages/pg-pool/index.js#L95) but is undocumented. Also found out `options` (aka `config`) exists on `Pool`, and is exposed in [`@types/pg`](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/70371) too. You can instantiate clients outside the pool's restrictions, with the same configuration. In case this doesn't exist, we now allow passing `Client` constructor to the dialect just in case it doesn't exist.

If `Client` exists (through `Pool` or through dialect config), use it. It doesn't register as a pool connection, so we won't have to wait for an idle connection.
If `Client` doesn't exist, get a pool connection. Not ideal.

Either way, do a cancel query for the PID, if we're still on the original query.

igalklebanov:
Got MySQL working as well in the following branch - https://github.com/kysely-org/kysely/pull/1505
It required killing the connection. `kill query` doesn't work as well as PostgreSQL's way, had to use `kill connection`.

src/dialect/pglite/pglite-dialect-config.ts
igalklebanov, line 1: The names were conflicting with the general `QueryOptions` this PR introduces.

igalklebanov:
Merged as https://github.com/kysely-org/kysely/pull/1796 into `next` branch.

PR #1494, opened 2025-06-25
feat(Introspect): add support for postgres & mysql foreign tables

williamluke4:
Hey 👋 

This PR adds foreign table introspection for PostgreSQL and MySQL dialects. An `isForeign` property is added to each table object.

### DIFF
--- a/src/dialect/database-introspector.ts
+++ b/src/dialect/database-introspector.ts
@@ -26,6 +26,11 @@ export interface DatabaseMetadataOptions {
    * such as the migration tables.
    */
   withInternalKyselyTables: boolean
+
+  /**
+   * If this is true, the metadata contains the foreign tables.
+   */
+  withForeignTables?: boolean
 }
 
 export interface SchemaMetadata {
@@ -43,6 +48,7 @@ export interface DatabaseMetadata {
 export interface TableMetadata {
   readonly name: string
   readonly isView: boolean
+  readonly isForeignTable: boolean
   readonly columns: ColumnMetadata[]
   readonly schema?: string
 }
--- a/src/dialect/database-introspector.ts
+++ b/src/dialect/database-introspector.ts
@@ -26,11 +26,6 @@ export interface DatabaseMetadataOptions {
    * such as the migration tables.
    */
   withInternalKyselyTables: boolean
-
-  /**
-   * If this is true, the metadata contains the foreign tables.
-   */
-  withForeignTables?: boolean
 }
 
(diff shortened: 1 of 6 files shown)


src/dialect/database-introspector.ts
```diff
@@ -26,6 +26,11 @@ export interface DatabaseMetadataOptions {
    * such as the migration tables.
    */
   withInternalKyselyTables: boolean
+
+  /**
+   * If this is true, the metadata contains the foreign tables.
+   */
+  withForeignTables?: boolean
```
igalklebanov: Why is this under a flag? Why not always grab them?

williamluke4: In-order to not change existing deployments. I figure there may be side effects so rolling it out like this would prevent that.

igalklebanov (commented):
Hey 👋 

Thanks! 🙏 

Looks good, just one question. 💪

src/dialect/database-introspector.ts
igalklebanov: > In-order to not change existing deployments. I figure they may be side effects so rolling it out like this would prevent that.

These changes would end up in the next minor version. You'd have to manually bump, knowing there might be breaking changes.

There's valuable information in "side effects" about how people are using introspection that will be lost.

Filtering, or allowing to filter, such tables, is a downstream problem that's trivial to solve.

Let's always include such tables (ditch the flag for now), and react to any issues that emerge.

src/dialect/database-introspector.ts
```diff
@@ -43,6 +48,7 @@ export interface DatabaseMetadata {
 export interface TableMetadata {
   readonly name: string
   readonly isView: boolean
+  readonly isForeignTable: boolean
```
igalklebanov:
```suggestion
  readonly isForeign: boolean
```

igalklebanov:
We've been fetching all types of tables in MySQL's introspector - including federated tables. Worth adding an `ENGINE = 'FEDERATED'` check to the output there for parity.

igalklebanov:
MS SQL Server has synonyms, and we're currently not introspecting them.

src/dialect/database-introspector.ts
williamluke4: 👍

williamluke4:
@igalklebanov Anything else i should add?

igalklebanov (approved):
LGTM! 🚀

Issue #1492, opened 2025-06-23
Allow explicit undefined in Updateable type (for exactOptionalPropertyTypes support)


y-hsgw:
Currently, the `Updateable<R>` type does not explicitly include | undefined for property values. When using TypeScript's `exactOptionalPropertyTypes` option, this causes a type error if you want to explicitly pass undefined to an update operation.

Example:

```ts

interface PersonTable  {
  id: Generated<number>;
  first_name: string;
}

type PersonUpdate = Updateable<PersonTable>;

interface PersonUpdateParams {
  firstName?: string;
}

export async function updatePerson(id: number, updateWith: PersonUpdateParams) {
  const personUpdate: PersonUpdate = {
    first_name: updateWith.firstName, // Type error
  };
  await db
    .updateTable('person')
    .set(personUpdate)
    .where('id', '=', id)
    .execute();
}
```

By contrast, passing `{ last_name: undefined }` directly to `.set()` does not cause a type error, because the property is just omitted.

How about explicitly allowing undefined in the [Updateable](https://github.com/kysely-org/kysely/blob/5647fe36f5ba79fb95363d4b13cdb18499636f65/src/util/column-type.ts#L203-L205) type as shown below, to resolve the above case?
```ts
export type Updateable<R> = DrainOuterGeneric<{
    [K in UpdateKeys<R>]?: UpdateType<R[K]> | undefined;
}>;
```

Similar issue: https://github.com/kysely-org/kysely/issues/170

koskimas:
I don't see any errors https://kyse.link/hbBsD

<img width="346" alt="Image" src="https://github.com/user-attachments/assets/c7378407-3aa2-4dc8-888c-dc47dd9f739f" />

y-hsgw:
@koskimas 
I believe you would see a type error if you enable `exactOptionalPropertyTypes` in your tsconfig.

<img width="603" alt="Image" src="https://github.com/user-attachments/assets/e70117b2-9f9d-40bf-a1ce-dbd1b4232bf1" />

https://stackblitz.com/edit/react-ts-8gcbohqj?file=test.ts

koskimas:
Oh, that's true.

taybin:
This is going to become more of an issue now that `noUncheckedIndexedAccess: true` is the new default for tsconfig.json files in typescript-5.9 (https://devblogs.microsoft.com/typescript/announcing-typescript-5-9-beta/#minimal-and-updated-tsc---init)

[...]

PR #1496, opened 2025-06-26
feat(utils): Allow explicit undefined in Updateable type (for exactOptionalPropertyTypes support)

y-hsgw:
Hey👋
I tried allowing the Updateable type to accept explicit undefined values to better support TypeScript's `exactOptionalPropertyTypes` option. What do you think?
I would appreciate your review!

Fixes: https://github.com/kysely-org/kysely/issues/1492

### DIFF
--- a/src/util/column-type.ts
+++ b/src/util/column-type.ts
@@ -201,5 +201,5 @@ export type Insertable<R> = DrainOuterGeneric<
  * ```
  */
 export type Updateable<R> = DrainOuterGeneric<{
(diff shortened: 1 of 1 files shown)


koskimas (approved):

koskimas:
I removed the `breaking change` label. This is more a bug fix than a breaking change. This just makes Kysely work the same with or without `exactOptionalPropertyTypes`. Without it, nothing changes.

PR #1507, opened 2025-07-08
fix `AsyncDisposable` usage erroring for older TypeScript versions.

### LINKED ISSUE #1499: Kysely does not work with Typescript below version 5.2 because of AsyncDisposable
Hey guys, when trying to compile my project I'm getting:

```typescript
../../node_modules/.pnpm/kysely@0.28.2/node_modules/kysely/dist/cjs/kysely.d.ts:54:92 - error TS2304: Cannot find name 'AsyncDisposable'.

54 export declare class Kysely<DB> extends QueryCreator<DB> implements QueryExecutorProvider, AsyncDisposable {
```

`AsyncDisposable` was introduced in TypeScript 5.2, your docs suggest that Kyseli works with 4.6 and above which doesn't seem to be the case.

It looks like the change was introduced in #1167.

Maybe you could suggest using Typescript 5.2 or suggest that some polyfill is needed.

igalklebanov:
Hey 👋 

closes #1499.

### DIFF
--- a/test/node/src/async-dispose.test.ts
+++ b/test/node/src/async-dispose.test.ts
@@ -6,6 +6,7 @@ import {
   PostgresAdapter,
   PostgresIntrospector,
   PostgresQueryCompiler,
+  QueryId,
   QueryResult,
   RootOperationNode,
   sql,
@@ -17,7 +18,6 @@ describe('async dispose', function () {
     const steps: string[] = []
 
     {
-      // @ts-ignore - `using` was only introduced in TS 5.2
       await using db = new Kysely({
         dialect: {
           createAdapter: () => new PostgresAdapter(),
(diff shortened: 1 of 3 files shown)


test/node/src/async-dispose.test.ts
```diff
@@ -17,7 +18,6 @@ describe('async dispose', function () {
     const steps: string[] = []
 
     {
-      // @ts-ignore - `using` was only introduced in TS 5.2
```
igalklebanov, line 20: we're excluding the entire file from the older version typechecks via a CI script.

test/node/tsconfig.json
```diff
@@ -1 +1,11 @@
-{"compilerOptions":{"target":"ES2022","lib":["ESNext"],"module":"CommonJS","outDir":"dist","skipLibCheck":true},"default":{"compilerOptions":{"target":"ES2022","lib":["ESNext"],"module":"CommonJS","outDir":"dist","skipLibCheck":true},"default":{"compilerOptions":{"target":"ES2022","lib":["ESNext"],"module":"CommonJS","outDir":"dist","skipLibCheck":true},"default":{"extends":"../../tsconfig-base.json","include":["src/**/*"],"compilerOptions":{"target":"ES2022","lib":["ESNext"],"module":"CommonJS","outDir":"dist","skipLibCheck":true}},"extends":"../../tsconfig-base.json","include":["src/**/*"],"exclude":["src/async-dispose.test.ts"]},"exclude":["src/async-dispose.test.ts"],"extends":"../../tsconfig-base.json","include":["src/**/*"]},"exclude":["src/async-dispose.test.ts","src/async-dispose.test.ts"],"extends":"../../tsconfig-base.json","include":["src/**/*"]}
```
igalklebanov, line 1: 😨

PR #1510, opened 2025-07-13
feat: add PGlite dialect.

### LINKED ISSUE #1165: Feature request: pglite dialect
According to the pglite docs:
> [PGlite](https://pglite.dev/) is a WASM Postgres build packaged into a TypeScript/JavaScript client library, that enables you to run Postgres in the browser, Node.js and Bun, with no need to install any other dependencies. It's under 3mb Gzipped.

There is discussion on a kysely-pglite dialect [here](https://github.com/dnlsandiego/kysely-pglite) and [here](https://github.com/electric-sql/pglite/discussions/128).

igalklebanov:
Hey 👋 

closes #1165.

This PR introduces a PGlite core dialect. It crossed the 1M downloads per week and there's no stopping it. People use it the browser, or in the backend. It's also a nice alternative to Docker for local development and testing.

The caveat is, it supports a single connection, it'll lag behind PostgreSQL releases a bit, and due to some technical restrictions, might produce results in different order (spotted with `MERGE INTO ... RETURNING` then regular PostgreSQL distributions.

Seeing that single-connection databases are a growing thing (there are a lot of SQLite flavors out there), the connection mutex, that was used in the SQLite driver, was moved to the runtime driver. Dialects can now, through the adapter, declare `supportsMultipleConnections: false` to use it.

Tests required a refactor in dialect types and lists - to support sql specs (e.g. `postgres`) and variants (e.g. `postgres` and `pglite`). This was a cheap way of enabling PGlite for everything PostgreSQL, while not requiring to add it everywhere we `testSql`.


igalklebanov:
All green. Requires some extra passes for documentation, and cleaning up.

koskimas (dismissed):

src/dialect/dialect-adapter.ts
```diff
@@ -18,6 +18,13 @@ export interface DialectAdapter {
    */
   readonly supportsCreateIfNotExists: boolean
 
+  /**
+   * Whether or not this dialect supports multiple connections at the same time.
+   *
+   * If this is false, Kysely will use a single connection for all database operations.
+   */
+  readonly supportsMultipleConnections: boolean
```
koskimas: Doesn't this break existing 3rd party dialects? Make it optional? If all 3rd party dialects inherit the `DialectAdapterBase` there's no need, but do they?

src/dialect/dialect-adapter.ts
```diff
@@ -35,7 +42,7 @@ export interface DialectAdapter {
    * Whether or not this dialect supports the `output` clause in inserts
    * updates and deletes.
    */
-  readonly supportsOutput?: boolean
+  readonly supportsOutput: boolean
```
koskimas: Same here. Doesn't this break backwards compatibility?

koskimas (approved):
LGTM. One question about the adapter.

site/docs/getting-started/Summary.tsx
```diff
@@ -46,13 +44,17 @@ const dialectSpecificCodeSnippets: Record<Dialect, string> = {
         cb.notNull().defaultTo(sql\`current_timestamp\`)
       )
       .execute()`,
+  pglite: postgresqlCodeSnippet,
 }
 
+const postgresqlTruncateSnippet = `await sql\`truncate table \${sql.table('person')}\`.execute(db)`
+
```
jadejr, line 51: It's not postgres specific, so it should have a different variable name right?

igalklebanov: Nice catch!

src/dialect/dialect-adapter.ts
igalklebanov: Absolutely right!

src/dialect/dialect-adapter.ts
igalklebanov: It does. Reverted.

thenbe:
Hi. Thanks for working on this. I noticed that this was merged into `next` branch 7 months ago, yet it is still not released (and not on master branch). Is this on purpose or was this commit perhaps forgotten?

jadejr:
@thenbe It is part of 0.29 https://github.com/kysely-org/kysely/pull/1583

PR #1516, opened 2025-07-18
feat: support dropping multiple types with schema.dropType(), cascade.

### LINKED ISSUE #1508: Support an array as input for `schema.dropType()`
Currently the `dropType()` method only supports a single string as input. Postgres, at least, supports a list of items, e.g. `DROP TYPE IF EXISTS webhook_events, mac_algorithm;`.
It would be nice to be able to generate that query with `db.schema.dropType(['webhook_events', 'mac_algorithm']).ifExists()`.

aantia:
Closes #1508

This PR adds support for arrays as input to the `schema.dropType()` method, e.g.
```ts
db.schema.dropType(['species', 'colors'])
```

While I was in the area I also added `.cascade()` ~~and `.restrict()`~~ methods, to completely support Postgres' `DROP TYPE`. I based the implementation on the one for `DropConstraintNode`.

I only added functionality tests as the existing `dropType()`  method didn't have type tests and I wasn't sure how to go about writing them.

### DIFF
--- a/src/operation-node/drop-type-node.ts
+++ b/src/operation-node/drop-type-node.ts
@@ -6,9 +6,8 @@ export type DropTypeNodeParams = Omit<Partial<DropTypeNode>, 'kind' | 'names'>
 
 export interface DropTypeNode extends OperationNode {
   readonly kind: 'DropTypeNode'
-  /** @deprecated Use `names` instead. */
-  readonly name?: SchemableIdentifierNode
-  readonly names: SchemableIdentifierNode[]
+  readonly name: SchemableIdentifierNode
+  readonly additionalNames?: SchemableIdentifierNode[]
   readonly ifExists?: boolean
   readonly modifier?: 'cascade' | 'restrict'
 }
@@ -22,10 +21,11 @@ export const DropTypeNode = freeze({
   },
 
   create(names: SchemableIdentifierNode | SchemableIdentifierNode[]): DropTypeNode {
+    names = Array.isArray(names) ? names : [names];
     return freeze({
       kind: 'DropTypeNode',
-      name: Array.isArray(names) ? undefined : names,
-      names: Array.isArray(names) ? names : [names],
+      name: names[0],
+      additionalNames: names.slice(1),
     })
   },
 
--- a/src/operation-node/drop-type-node.ts
+++ b/src/operation-node/drop-type-node.ts
@@ -2,7 +2,7 @@ import { freeze } from '../util/object-utils.js'
 import { OperationNode } from './operation-node.js'
 import { SchemableIdentifierNode } from './schemable-identifier-node.js'
 
-export type DropTypeNodeParams = Omit<Partial<DropTypeNode>, 'kind' | 'name'>
+export type DropTypeNodeParams = Omit<Partial<DropTypeNode>, 'kind' | 'names'>
 
 export interface DropTypeNode extends OperationNode {
   readonly kind: 'DropTypeNode'
--- a/src/operation-node/drop-type-node.ts
+++ b/src/operation-node/drop-type-node.ts
@@ -2,7 +2,10 @@ import { freeze } from '../util/object-utils.js'
 import { OperationNode } from './operation-node.js'
 import { SchemableIdentifierNode } from './schemable-identifier-node.js'
 
-export type DropTypeNodeParams = Omit<Partial<DropTypeNode>, 'kind' | 'name' | 'names'>
+export type DropTypeNodeParams = Omit<
+  Partial<DropTypeNode>,
+  'kind' | 'name' | 'additionalNames'
+>
 
 export interface DropTypeNode extends OperationNode {
   readonly kind: 'DropTypeNode'
@@ -20,8 +23,13 @@ export const DropTypeNode = freeze({
     return node.kind === 'DropTypeNode'
   },
 
-  create(names: SchemableIdentifierNode | SchemableIdentifierNode[]): DropTypeNode {
-    names = Array.isArray(names) ? names : [names];
+  create(
+    names: SchemableIdentifierNode | SchemableIdentifierNode[],
+  ): DropTypeNode {
+    if (!Array.isArray(names)) {
+      names = [names]
+    }
+
     return freeze({
       kind: 'DropTypeNode',
       name: names[0],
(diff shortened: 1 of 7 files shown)


src/operation-node/drop-type-node.ts
```diff
@@ -2,12 +2,13 @@ import { freeze } from '../util/object-utils.js'
 import { OperationNode } from './operation-node.js'
 import { SchemableIdentifierNode } from './schemable-identifier-node.js'
 
-export type DropTypeNodeParams = Omit<Partial<DropTypeNode>, 'kind' | 'name'>
+export type DropTypeNodeParams = Omit<Partial<DropTypeNode>, 'kind' | 'names'>
 
 export interface DropTypeNode extends OperationNode {
   readonly kind: 'DropTypeNode'
-  readonly name: SchemableIdentifierNode
+  readonly names: SchemableIdentifierNode[]
```
igalklebanov: This is correct. But how can we make it not break any existing compilers and plugins?

src/operation-node/drop-type-node.ts
```diff
@@ -18,10 +19,10 @@ export const DropTypeNode = freeze({
     return node.kind === 'DropTypeNode'
   },
 
-  create(name: SchemableIdentifierNode): DropTypeNode {
+  create(names: SchemableIdentifierNode[]): DropTypeNode {
```
igalklebanov: https://github.com/kysely-org/kysely/pull/1516/files#r2217039372

src/query-compiler/default-query-compiler.ts
```diff
@@ -1428,7 +1428,13 @@ export class DefaultQueryCompiler
...
     }
 
-    this.visitNode(node.name)
+    this.compileList(node.names)
+
+    if (node.modifier === 'cascade') {
+      this.append(' cascade')
+    } else if (node.modifier === 'restrict') {
+      this.append(' restrict')
+    }
```
igalklebanov: It's OK to trust internals that created the node.


```suggestion
    if (node.modifier) {
      this.append(` ${node.mofider}`)
    }
```

aantia:
@igalklebanov Will this work for compatibility? I've never really looked at any Kysely plugins or compilers, so I'm not quite sure what I'm doing.

igalklebanov:
> @igalklebanov Will this work for compatibility? I've never really looked at any Kysely plugins or compilers, so I'm not quite sure what I'm doing.

It's perfectly fine to ask and not get things right the first time. I'm more than happy to help!

A library operates within an ecosystem. Public APIs are important, and smoothening transitions is something that needs to be thought about.

When you rename a property in an operation node or change the signature of the function that creates such nodes - that's a breaking change that might affect (custom) dialects (their compilers) and plugins, potentially making them unuseable with the next version of this library.

Currently, the ecosystem knows of `name`, and that it always exists. Let's not break that promise. 
It also knows that in order to create a `DropTypeNode` you provide a single `name` as the first argument. Let's not break that promise.

What we can do:

1. Keep `node.name`, untouched. It will represent the leading type name in the query. This ensures downstream code that accesses `node.name` doesn't break.
1. Add `node.additionalNames?: SchemableIdentifierNode[]`. It will represent any other type names (excluding the leading type name). The cost is slighly more complex compilation code, and it's a price we're willing to pay.
1. Add a second `additionalNames?: SchemableIdentifierNode[]` argument in `DropTypeNode.create`, leaving the first function argument (`name`) untouched. The cost is a `slice` operation before invoking `DropTypeNode.create`, and it's a price we're willing to pay. OR, turn the first argument into a `SchemableIdentifierNode | SchemableIdentifierNode[]` union, and handle both cases within the creation logic.

aantia:
This should now be ready for re-review. Thanks for the help :)

src/operation-node/drop-type-node.ts
```diff
@@ -2,12 +2,14 @@ import { freeze } from '../util/object-utils.js'
 import { OperationNode } from './operation-node.js'
 import { SchemableIdentifierNode } from './schemable-identifier-node.js'
 
-export type DropTypeNodeParams = Omit<Partial<DropTypeNode>, 'kind' | 'name'>
+export type DropTypeNodeParams = Omit<Partial<DropTypeNode>, 'kind' | 'names'>
```
igalklebanov: This type is used in the `cloneWith` below. We don't want `name` and `names` to be overriden after `create`.

```suggestion
export type DropTypeNodeParams = Omit<Partial<DropTypeNode>, 'kind' | 'name' | 'names'>
```

src/query-compiler/default-query-compiler.ts
```diff
@@ -1428,7 +1428,11 @@ export class DefaultQueryCompiler
       this.append('if exists ')
     }
 
-    this.visitNode(node.name)
+    this.compileList([node.name, ...(node.additionalNames ?? [])])
```
igalklebanov: Let's simplify this a bit.

```suggestion
    this.visitNode(node.name)
    
    if (node.additionalNames?.length) {
      this.append(', ')
      this.compileList(node.additionalNames)
    }
```

src/parser/identifier-parser.ts
```diff
@@ -16,6 +16,14 @@ export function parseSchemableIdentifier(id: string): SchemableIdentifierNode {
   }
 }
 
+export function parseSchemableIdentifierArray(id: string | string[]): SchemableIdentifierNode[] {
+  if (Array.isArray(id)) {
+		return id.map(parseSchemableIdentifier)
+	} else {
+		return [parseSchemableIdentifier(id)]
+	}
```
igalklebanov: Let's simplify this a bit.

```suggestion
  if (!Array.isArray(id)) {
    id = [id]
  }
  
  return id.map(parseSchemableIdentifier)
```

src/schema/schema.ts
```diff
@@ -311,12 +311,20 @@ export class SchemaModule {
    *   .ifExists()
    *   .execute()
    * ```
+   * 
+   * ```ts
```
igalklebanov:
```suggestion
   * You can also provide multiple type names:
   * 
   * ```ts
```

igalklebanov (approved):
Removed `.restrict()` to align with drop view.
Applied my suggestions.
Ran formatter.

PR #1521, opened 2025-07-22
Fix fn.any to support ColumnType references

### LINKED ISSUE #1519: `fn.any` doesn't handle references of columns defined with `ColumnType`.
Hey 👋 

Currently `fn.any` does not handle references of columns of type `ColumnType`.
As a result, you always end up with a `KyselyTypeError` because `ColumnType` does not extend an array.

Example:

```ts
export interface PersonTable {
  id: Generated<number>;
  first_name: string;
  last_name: string | null;
  ids: ColumnType<number[] | null>; // if you replace the type with `number[] | null`, there is not type error later.
}

const rows = await db
  .selectFrom("person")
  .select(["first_name"])
  .where((eb) => eb("id", "=", eb.fn.any("ids")))
//                             ^^^^^^^^^^^^^^^^ Argument of type 'KyselyTypeError<"any(expr) call failed: expr must be an array">' is not assignable to parameter of type 'OperandValueExpressionOrList<Database, "person", "id">'.
  .execute();
```
https://kyse.link/a8N3R

Ruthenz:
Closes #1519 

## Summary
Fixes `fn.any` function to properly handle columns defined with `ColumnType<T[]>`.

Previously, using `fn.any` with `ColumnType` columns would result in:
KyselyTypeError<"any(expr) call failed: expr must be an array">

## Problem
The issue occurred because `fn.any` was using `ExtractTypeFromStringReference` which returns
the raw `ColumnType` structure, but the array type check expected the actual select type.

## Solution
Changed to use `ExtractTypeFromReferenceExpression` which properly extracts the `SelectType`
from `ColumnType`, similar to how other functions like `max`/`min` handle it.

## Example
```typescript
interface PersonTable {
  id: Generated<number>;
  ids: ColumnType<number[] | null>;
}

// This now works without type errors
const rows = await db
  .selectFrom("person")
  .select(["first_name"])
  .where((eb) => eb("id", "=", eb.fn.any("ids"))) // ✅ Fixed!
  .execute();

Changes

- Fixed: fn.any type definition in src/query-builder/function-module.ts
- Added: Comprehensive test cases for ColumnType with fn.any

Testing

- ✅ Added failing test case that reproduces the issue
- ✅ Verified tests fail before the fix
- ✅ Verified tests pass after the fix
- ✅ All existing type tests still pass
- ✅ Code formatting and linting passed

### DIFF
--- a/src/query-builder/function-module.ts
+++ b/src/query-builder/function-module.ts
@@ -655,7 +655,7 @@ export interface FunctionModule<DB, TB extends keyof DB> {
   any<RE extends StringReference<DB, TB>>(
     expr: RE,
   ): Exclude<
-    ExtractTypeFromStringReference<DB, TB, RE>,
+    ExtractTypeFromReferenceExpression<DB, TB, RE>,
     null
   > extends ReadonlyArray<infer I>
     ? ExpressionWrapper<DB, TB, I>
--- a/test/typings/test-d/expression.test-d.ts
+++ b/test/typings/test-d/expression.test-d.ts
@@ -1,18 +1,20 @@
 import {
   expectAssignable,
-  expectNotAssignable,
   expectError,
+  expectNotAssignable,
   expectType,
 } from 'tsd'
 import {
+  ColumnType,
   Expression,
   ExpressionBuilder,
+  Generated,
   Kysely,
   SqlBool,
   expressionBuilder,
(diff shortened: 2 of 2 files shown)


koskimas:
Perfect. Thank you!

PR #1526, opened 2025-07-28
feat: support narrowing by deep object keys in NarrowPartial

ethanresnick:
This allows kysely to support slightly more cases of denormalized data (in particular, json columns with discriminated unions in them). It makes `NarrowPartial` behave more like type intersection in this case.

See the new test, which would've previously failed with a `KyselyTypeError` unless the user wrote:

```ts
.$narrowType<{ discriminatedUnionProfile: { auth: { type: 'token', token: string }, tags: string[] } }>()
```

Now, the user just has to write:

```ts
.$narrowType<{ discriminatedUnionProfile: { auth: { type: 'token' } } }>()
```

No existing tests are affected.


src/util/type-utils.ts
```diff
@@ -153,19 +153,19 @@ export type Equals<T, U> =
     ? true
     : false
 
-export type NarrowPartial<O, T> = DrainOuterGeneric<
```
ethanresnick: The `DrainOuterGeneric` never should've been wrapped around the conditional type; it only makes sense on the object type in the true branch. But `NarrowPartial` didn't used to be a conditional type, and then I (in a long ago PR) made `NarrowPartial` conditional but didn't move the `DrainOuterGeneric`.

test/typings/test-d/select.test-d.ts
```diff
@@ -115,6 +115,11 @@ async function testSelectSingle(db: Kysely<Database>) {
     .$narrowType<NarrowTarget>()
     .execute()
 
+  expectType<
+    | { callback_url: string; queue_id: null }
+    | { callback_url: null; queue_id: string }
+  >(r15)
```
ethanresnick, line 121: `r15` was unused/never tested at all before. so this was just a missing test; its result is not changed by this pr

ethanresnick:
@igalklebanov I forced pushed this to fix the merge conflicts. What's the best way to move this forward?

igalklebanov:
> @igalklebanov I forced pushed this to fix the merge conflicts. What's the best way to move this forward?

Hey 👋 

Sorry about the delay.

Seems like the tests are failing. I love the capability this thing might enable.

Issue #1210, opened 2024-11-05
add `thenRef` in `eb.case`.


igalklebanov:
Hey 👋

`then` is currently confusing. A `thenRef` would be an alignment with other builders, and make a clear distinction between value-first and ref-first.

ericsodev:
Hey, I'd like to try working on this!

igalklebanov:
It's yours!

Swing by Discord if you have any questions.

igalklebanov:
Parked in `next`, ready for 0.29.0.

PR #1531, opened 2025-08-02
feat: add `thenRef` method in `eb.case`

ericsodev:
Closes https://github.com/kysely-org/kysely/issues/1210.

This PR adds a `thenRef` method when using the case builder:

```ts
eb
  .case('gender')
  .when(sql.lit('male'))
  .thenRef('first_name') // Equates to the existing .then(eb.ref('first_name'))
  .else(null)
  .end()
  .as('name')
```


igalklebanov (approved):
LGTM! 🚀 

Corrected a result type bug.
Changed test cases a bit.

PR #1545, opened 2025-08-10
feat: Sensible `pg` defaults

### LINKED ISSUE #1544: `pg` JSON serialization is inconsistent with raw values serialization
In Postgres if you run:

```
SELECT now()::timestamptz;
SELECT to_json(now()::timestamptz);
```

you get different formats for timestamptz:

```
2025-08-10 14:44:40.687342+00
2025-08-10T14:44:40.687342+00:00
```

This can be surprising for new users as the same table may return two different formats for the same column depending on whether they are fetched from a regular select or in a JSON serialization context (e.g. `jsonArrayFrom` or `jsonObjectFrom`).

The proposal is to add an helper to set sensible defaults in `pg.Client` to make it so records fetched from `jsonObjectFrom` and regular selects return the same data.

shoooe:
Closes #1544

Note to self:
Run tests with: 
```
pnpm build && pnpm test:node:build && pnpm mocha --timeout 15000 test/node/dist/sensible-pg-defaults.test.js
```


src/dialect/postgres/postgres-dialect-config.ts
```diff
@@ -39,6 +39,11 @@ export interface PostgresDialectConfig {
    * Called every time a connection is acquired from the pool.
    */
   onReserveConnection?: (connection: DatabaseConnection) => Promise<void>
+
+  /**
+   * @todo: docs
+   */
+  types?: Record<number, (value: string) => any>
```
shoooe, line 55: Naming is taken from the equivalent option in `pg`: https://node-postgres.com/apis/client#new-client

src/dialect/postgres/postgres-driver.ts
```diff
@@ -35,6 +35,14 @@ export class PostgresDriver implements Driver {
 
   async acquireConnection(): Promise<DatabaseConnection> {
     const client = await this.#pool!.connect()
+
+    if (!!this.#config.types) {
+      for (const [typeId, parserFn] of Object.entries(this.#config.types)) {
+        const id = Number(typeId)
+        client.setTypeParser(id, parserFn)
+      }
+    }
```
igalklebanov: this should happen only once, when there's no connection cached for the client.

src/dialect/postgres/postgres-driver.ts
```diff
@@ -35,6 +35,14 @@ export class PostgresDriver implements Driver {
 
   async acquireConnection(): Promise<DatabaseConnection> {
     const client = await this.#pool!.connect()
+
+    if (!!this.#config.types) {
```
igalklebanov:
```suggestion
    if (this.#config.types) {
```

src/helpers/postgres.ts
```diff
@@ -1,3 +1,4 @@
+import { types } from 'pg'
```
igalklebanov: We should not use any types or anything runtime from other libraries.

src/helpers/postgres.ts
igalklebanov, line 1: Regexes and validation and surprise runtime errors are overkill. The data is coming from the driver/database, trust that it's valid.

It should be a simple, replace ` ` with `T`, and add minutes to the offset.

src/helpers/postgres.ts
igalklebanov: Instead, hardcode the `oid`s, and comment next to each which data type it is. Linking to the `pg-types` source code is also good.

src/dialect/postgres/postgres-driver.ts
shoooe: You mean I should move this under the `if (!connection)`?

igalklebanov: Probably. Do you see any reason it should happen more than once?

shoooe: I don't think so 👍

.gitignore
```diff
@@ -4,3 +4,4 @@ dist
 node_modules
 .vscode
 .idea
+.DS_Store
```
shoooe, line 7: This is a pretty common macOS file that gets generated by the OS.

shoooe:
@igalklebanov The failing test says:

> browserType.launch: Executable doesn't exist at /home/runner/.cache/ms-playwright/chromium_headless_shell-1181/chrome-linux/headless_shell

which doesn't look like an actual test failure caused by this PR. Is this perhaps also in the main trunk?

PR #1550, opened 2025-08-16
fix: filter schemas the current user does not have access to in postres introspector

### LINKED ISSUE #1537: issue with migrations using postgres + schemas with limited access
Hi there, so I found out some issues with Kysely,

For context, I'm using kysely 0.28.4 with postgres driver,

The database is set up to have multiple schemas, each with its independent user.
Each schema has its migration tables and is an independent entity in nearly all aspects,

When I try to migrate (using kysely-ctl), the migration fails because kysely is trying to call "getTable" here ->
https://github.com/kysely-org/kysely/blob/master/src/migration/migrator.ts#L464

But this query fails because the user performing the migration does not have access to all schemas, only their own.

My workaround for now is to give each user access to all schemas, but I think it should not be necessary, if Kysely would only check the table for the current schema instead of getting all tables and then checking for the schema

chanon:
Fixes #1537.
The issue was that the current `PostgresIntrospector` tries to query information from all tables from all schemas except for `pg_*` schemas, `information_shema` and `crdb_internal` schema for CockroachDB.

This adds a check to filter out any schemas that the current user/connection does not have access to.

### DIFF
--- a/src/dialect/postgres/postgres-introspector.ts
+++ b/src/dialect/postgres/postgres-introspector.ts
@@ -75,6 +75,8 @@ export class PostgresIntrospector implements DatabaseIntrospector {
       .where('ns.nspname', '!=', 'information_schema')
       // Filter out internal cockroachdb schema
       .where('ns.nspname', '!=', 'crdb_internal')
+      // Only schemas where we are allowed access
+      .where(sql<boolean>`has_schema_privilege(ns.nspname, 'USAGE')`)
       // No system columns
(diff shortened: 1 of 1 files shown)


ematthewsBW:
thanks for fixing this @chanon. Would love to see this one merged, as the bug is causing us some pain 😭😭😭

igalklebanov (approved):
Hey 👋 

Thanks! 🙏 

LGTM! 🚀

PR #1554, opened 2025-08-20
fix(WithSchemaTransformer): `using` table of `merge into` queries unhandled.

alexander-azizi-martin:
The `withSchema` method is documented to set the default schema for all table references that don't explicitly specify a schema ([docs](https://kysely.dev/docs/recipes/schemas#2)). However, it currently fails to apply the schema to join tables in `mergeInto` statements.

When using `withSchema` with `mergeInto`, the schema is not applied to the table specified in the `.using()` clause:

```typescript
await db
 .withSchema(tenant)
 .mergeInto('person as target')
 .using('pet as source', 'source.owner_id', 'target.id')
 .whenMatchedAnd('target.has_pets', '!=', 'Y')
 .thenUpdateSet({ has_pets: 'Y' })
 .whenNotMatchedBySourceAnd('target.has_pets', '=', 'Y')
 .thenUpdateSet({ has_pets: 'N' })
 .executeTakeFirstOrThrow()
```

In this example, the `person` table correctly receives the schema prefix, but the `pet` table in the `.using()` clause does not.

This PR updates the `WithSchemaTransformer` class to properly handle join tables in `mergeInto` statements.

### DIFF
--- a/test/node/src/with-schema.test.ts
+++ b/test/node/src/with-schema.test.ts
@@ -1,31 +1,40 @@
+import type { Kysely } from '../../..'
 import {
   clearDatabase,
   destroyTest,
   initTest,
-  TestContext,
+  type TestContext,
   testSql,
   NOT_SUPPORTED,
   createTableWithId,
   DIALECTS,
   insert,
   limit,
+  type Database,
+  type Pet,
 } from './test-setup.js'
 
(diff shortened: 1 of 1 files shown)


igalklebanov (approved):
Hey 👋 

Thanks! 🙏 

The test case was not reproducing the bug - made sure it does.

PR #1572, opened 2025-09-05
perf: optimize `randomString` utility function

threeal:
This PR optimizes `randomString` by generating characters directly from their ASCII codes with `String.fromCharCode`. This reduces memory usage and eliminates per-character string concatenation, making the function faster and more efficient.

### DIFF
--- a/src/util/random-string.ts
+++ b/src/util/random-string.ts
@@ -1,78 +1,10 @@
-const CHARS = [
-  'A',
-  'B',
-  'C',
-  'D',
-  'E',
-  'F',
-  'G',
-  'H',
-  'I',
-  'J',
-  'K',
-  'L',
-  'M',
-  'N',
-  'O',
-  'P',
-  'Q',
-  'R',
-  'S',
-  'T',
-  'U',
-  'V',
-  'W',
-  'X',
-  'Y',
-  'Z',
-  'a',
-  'b',
-  'c',
-  'd',
-  'e',
-  'f',
-  'g',
-  'h',
-  'i',
-  'j',
-  'k',
-  'l',
-  'm',
-  'n',
-  'o',
-  'p',
-  'q',
-  'r',
-  's',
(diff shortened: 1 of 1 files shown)


src/util/random-string.ts
```diff
@@ -1,78 +1,10 @@
...
-}
-
-function randomChar() {
-  return CHARS[~~(Math.random() * CHARS.length)]
+  return String.fromCharCode(
+    ...Array.from({ length }, () => {
+      const n = Math.floor(Math.random() * 62)
+      if (n < 26) return 65 + n             // 'A'-'Z'
+      if (n < 52) return 97 + (n - 26)      // 'a'-'z'
+      return 48 + (n - 52)                  // '0'-'9'
```
igalklebanov, line 7: use of "magic numbers" is a no-go for maintainability/readability.

src/util/random-string.ts
```diff
@@ -1,78 +1,10 @@
...
-    chars += randomChar()
-  }
-
-  return chars
-}
-
-function randomChar() {
-  return CHARS[~~(Math.random() * CHARS.length)]
+  return String.fromCharCode(
+    ...Array.from({ length }, () => {
```
igalklebanov, line 3: `...` runs the risk of stack overflow given that functions have a limit on the amount of arguments they can accept.

igalklebanov:
> This reduces memory usage

by how much? is it worth the problems introduced with the implicit/complex solution?

igalklebanov:
> eliminates per-character string concatenation

This is not a problem in modern engines.

igalklebanov:
> making the function faster and more efficient

by how much? is it worth the problems introduced with the implicit/complex solution?

igalklebanov:
Hey 👋 

Judging by your profile I can tell this is automated. We don't like this behavior over here. Bye.

PR #1573, opened 2025-09-05
feat: utilize `Promise.withResolvers`

threeal:
Utilize [`Promise.withResolvers`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers), with fallback support for returning a deferred promise if `Promise.withResolvers` is not available, replacing the `Deferred` class.


igalklebanov:
Hey 👋 

Thanks! 🙏 

There's absolutely no true good reason to add this at this point. I'm afraid Imma have to turn it down.

PR #1598, opened 2025-09-28
feat: add `whenRef(lhs, op, rhs)` in `eb.case`.

### LINKED ISSUE #1592: add `whenRef` in `eb.case`.
Hey 👋

A `whenRef` would be an alignment with `whereRef`, and make a clear distinction between value-first and ref-first.

Should be similar to `whereRef` and #1210.

iam-abdul:
closes #1592.

- Added whenRef() method to CaseBuilder and CaseWhenBuilder classes
- Added conditional typing to restrict whenRef to searched CASE only
- Support mixing whenRef with existing when/then/else chains
- Added test cases and type tests


src/query-builder/case-builder.ts
```diff
@@ -238,6 +271,14 @@ interface Whenable<DB, TB extends keyof DB, W, O> {
       ? KyselyTypeError<'when(value) is only supported when using case(value)'>
       : W,
   ): CaseThenBuilder<DB, TB, W, O>
+
+  whenRef<RE extends ReferenceExpression<DB, TB>>(
```
igalklebanov, line 282: Let's add documentation for it.

```suggestion
  /**
   * Adds a `when` clause to the case statement, where both sides of the 
   * operator are references to columns.
   *
   * The normal `when` method treats the right hand side argument as a 
   * value by default. `whenRef` treats it as a column reference.
   */
  whenRef<RE extends ReferenceExpression<DB, TB>>(
```

pnpm-lock.yaml
igalklebanov, line 1: should be reverted.

test/typings/test-d/case.test-d.ts
```diff
@@ -121,6 +121,16 @@ async function testCase(eb: ExpressionBuilder<Database, 'person'>) {
   expectError(eb.case().when('gender', '??', 'male').then('Mr.').end())
   expectError(eb.case().when('gender', '=', 42).then('Mr.').end())
   expectError(eb.case().when('male').then('Mr.').end())
+
+  expectError(
+    eb.case().whenRef('no_such_column', '=', 'last_name').then('Mr.').end(),
+  )
+  expectError(
+    eb.case().whenRef('first_name', '??', 'last_name').then('Mr.').end(),
```
igalklebanov: let's make the error reason stupid obvious.

```suggestion
    eb.case().whenRef('first_name', 'no_such_operator', 'last_name').then('Mr.').end(),
```

pnpm-lock.yaml
iam-abdul: done

igalklebanov: you deleted the file.. 😄

iam-abdul: sorry my bad, restored it from the next branch 🙈

igalklebanov: you've restored it from an outdated local branch perhaps?

iam-abdul: ah yes, i might have pulled it from my fork, corrected it now

test/node/src/case.test.ts
```diff
@@ -351,5 +352,177 @@ for (const dialect of DIALECTS) {
...
+            'select case when "first_name" = "last_name" then ?',
+            'when "gender" = ? then ?',
+            'else ? end as "title" from "person"',
+          ],
+          parameters: ['name_match', 'female', 'female_match', 'no_match'],
+        },
+      })
+
+      await query.execute()
+    })
```
igalklebanov: these are kinda redundant. let's remove them.

the new test case we'll keep verifies that `whereRef` compilation works, and that rhs column references are allowed against real engines.

the test cases we're removing don't verify anything new.

test/typings/test-d/case.test-d.ts
```diff
@@ -166,4 +176,9 @@ function testCaseValue(eb: ExpressionBuilder<Database, 'person'>) {
   expectError(eb.case('no_such_column').when('male').then('Mr.').end())
   expectError(eb.case('gender').when('robot').then('Mr.').end())
   expectError(eb.case('gender').when('gender', '=', 'male').then('Mr.').end())
+
+  // whenRef should not be available for simple CASE
```
igalklebanov: what is a "simple case"? let's remove this confusing comment.

test/node/src/case.test.ts
iam-abdul: agreed, 

i have removed those

test/typings/test-d/case.test-d.ts
iam-abdul: removed

igalklebanov (approved):
LFGTM! 🚀

PR #1605, opened 2025-10-06
setRef inside UpdateTableBuilder

iam-abdul:
Added setRef to UpdateQueryBuilder to allow:
.setRef('target', 'source')

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1474,9 +1474,6 @@ packages:
   '@electric-sql/pglite@0.3.8':
     resolution: {integrity: sha512-VlAz/R7mktifp9IHzNvjxWJM8p3fPH2lHpustYuRSOXOpXiAMTlA5qqxcufPaDnfee6CZCE9qrT1MHDT7riSHg==}
 
-  '@electric-sql/pglite@0.3.8':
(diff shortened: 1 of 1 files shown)


iam-abdul:
@igalklebanov what you think about this implementation. 
I am yet to implement 
```
.setRef({
  target: 'source',
})
```

wanted to get your feedback before i start on this

src/parser/reference-parser.ts
```diff
@@ -109,6 +109,30 @@ export type ExtractColumnNameFromOrderedColumnName<C extends string> =
...
+  TB extends keyof DB,
+  Columns extends string,
+  TargetType,
+> = {
+  [C in Columns]: UpdateType<
+    ExtractRawTypeFromReferenceExpression<DB, TB, C>
+  > extends TargetType
+    ? C
+    : never
+}[Columns]
```
igalklebanov, line 126: this is premature. demands deeper observation and benching.
proceed without all of this.

iam-abdul: ok just RE on both key and value (similar to where?)

igalklebanov: yeah, that's good enough for now.

iam-abdul: ok 🫡

Issue #1606, opened 2025-10-06
Support cast "signed" / "unsigned" for MySQL


Patrick-Ullrich:
Currently `cast` does not accept `signed` and `unsigned`.

However in mysql:
```
SELECT cast("0" AS integer)
```

will result in:
`Query 1 ERROR at Line 1: : You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'integer)' at line 1`

The correct syntax be:
```
SELECT cast("0" AS signed)
```
or
```
SELECT cast("0" AS unsigned)
```

PR #1607, opened 2025-10-06
fix: valid cast options per dialect

Patrick-Ullrich:
WIP - still requires tests etc. But logically what I think should do the trick to support the correct cast types

### DIFF
--- a/src/expression/expression-builder.ts
+++ b/src/expression/expression-builder.ts
@@ -76,6 +76,10 @@ import {
   DataTypeExpression,
   parseDataTypeExpression,
 } from '../parser/data-type-parser.js'
+import {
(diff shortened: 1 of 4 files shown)


Patrick-Ullrich:
https://github.com/kysely-org/kysely/issues/1606

koskimas:
We don't have dialect specific code in the core.

Patrick-Ullrich:
@koskimas could we just add unsigned | signed in there then? I was hesitant on that approach because they are not column types, which is currently the source for the CAST. 

Another approach could be not having dialect specific ones here, but create a new type that has all the CAST valid types for all databases (not dialect specific code), which will have a complete list.

Any other ideas on how to solve the linked issue?

PR #1608, opened 2025-10-07
fix: Migrator attempts to create custom migration schema even if it exists and fails on dialects that don't support `if not exists`.

### LINKED ISSUE #1552: Migrator attempts to create schema even when it exists
Hi, it seems `migrator.js` attempts to create schema even when it already exists:

```ts
  async #ensureMigrationTableExists(): Promise<void> {
    if (!(await this.#doesTableExists(this.#migrationTable))) {
      try {
        if (this.#migrationTableSchema) {
          // ⚠️ note the absence of existence check here,
          //    the whole block is also likely redundant
          await this.#createIfNotExists(
            this.#props.db.schema.createSchema(this.#migrationTableSchema),
          )
        }
```

I think this should have been taken care of by:

```ts
  async #ensureMigrationTablesExists(): Promise<void> {
    await this.#ensureMigrationTableSchemaExists()  // ✔️ creates schema
    await this.#ensureMigrationTableExists()        // ✔️ uses it
    await this.#ensureMigrationLockTableExists()
    await this.#ensureLockRowExists()
  }
```

Or am I missing something?

In practice this breaks my use case with custom Oracle 19 adapter built with [kysely-oracledb](https://github.com/griffiths-waite/kysely-oracledb).

austin-hall-skylight:
Fixes #1552.
Removes call  to `createIfNotExists` for `migrationTableSchema` in `ensureMigrationTableExists` since it will always be checked in `ensureMigrationTablesExists`. This enables using the `migrationTableSchema` property in `MigratorProps` with MS SQL Server as the migrator will no longer attempt to recreate the schema every time migrations are run.

### DIFF
--- a/src/migration/migrator.ts
+++ b/src/migration/migrator.ts
@@ -76,9 +76,9 @@ export class Migrator {
    * The returned array is sorted by migration name.
    */
   async getMigrations(): Promise<ReadonlyArray<MigrationInfo>> {
-    const executedMigrations = (await this.#doesTableExists(
-      this.#migrationTable,
-    ))
+    const tableExists = await this.#doesTableExist(this.#migrationTable)
+
+    const executedMigrations = tableExists
       ? await this.#props.db
           .withPlugin(this.#schemaPlugin)
           .selectFrom(this.#migrationTable)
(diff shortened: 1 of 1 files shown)


src/migration/migrator.ts
```diff
@@ -799,7 +793,7 @@ export interface MigratorProps {
    * the project, to the end of time or prepare to manually migrate the migration
    * tables.
    *
-   * This only works on postgres.
+   * This only works on postgres and mssql.
```
austin-hall-skylight, line 815: I have updated the relevant test and tested it manually in my project that is using Kysely. Is there anything else that I need to check to confirm that `migrationTableSchema` works with mssql?

igalklebanov: No, this is fine.

PR #1615, opened 2025-10-10
feat: Add support for dropping temporary tables with temporary() modifier

### LINKED ISSUE #874: dropTable temporary modifier
currently, `createTable` has a `temporary` modifier, while `dropTable` does not.

while `dropTable` works fine with temporary tables, it would be useful to also add a temporary modifier for the sake of safety.

szalonna:
Closes #874

### DIFF
--- a/test/node/src/schema.test.ts
+++ b/test/node/src/schema.test.ts
@@ -1513,7 +1513,7 @@ for (const dialect of DIALECTS) {
           })
         })
 
-        if (dialect === 'mysql') {
+        if (sqlSpec === 'mysql') {
           it('should drop a temporary table if it exists', async () => {
             const builder = ctx.db.schema.dropTable('test').temporary().ifExists()
             


site/docs/migrations.mdx
```diff
@@ -88,6 +88,14 @@ export async function down(db: Kysely<any>): Promise<void> {
...
 ```
 
+You can also drop temporary tables using the `temporary()` modifier:
+
+```ts
+export async function down(db: Kysely<any>): Promise<void> {
+  await db.schema.dropTable('temp_calculations').temporary().execute()
+}
+```
+
```
igalklebanov: does not fit this page's intent.
too small of a feature to be highlighted.

```suggestion
```

src/schema/drop-table-builder.ts
```diff
@@ -13,6 +13,20 @@ export class DropTableBuilder implements OperationNodeSource, Compilable {
     this.#props = freeze(props)
   }
 
+  /**
+   * Adds the "temporary" modifier.
+   *
+   * Use this to drop a temporary table.
```
igalklebanov:
```suggestion
   *
   * This is only supported by some dialects like MySQL.
```

test/node/src/schema.test.ts
```diff
@@ -1479,6 +1479,44 @@ for (const dialect of DIALECTS) {
...
+              sql: 'drop temporary table `test`',
+              parameters: [],
+            },
+            mssql: NOT_SUPPORTED,
+            sqlite: {
+              sql: 'drop temporary table "test"',
+              parameters: [],
+            },
+          })
+        })
```
igalklebanov: the next test case already covers it.

```suggestion
```

test/node/src/schema.test.ts
```diff
@@ -1479,6 +1479,44 @@ for (const dialect of DIALECTS) {
...
+              sql: 'drop temporary table if exists `test`',
+              parameters: [],
+            },
+            mssql: NOT_SUPPORTED,
+            sqlite: {
+              sql: 'drop temporary table if exists "test"',
+              parameters: [],
+            },
+          })
+        })
```
igalklebanov: This is only supported in MySQL.

```suggestion
        if (dialect === 'mysql') {
          it('should drop a temporary table if it exists', async () => {
            const builder = ctx.db.schema.dropTable('test').temporary().ifExists()
            
            testSql(builder, dialect, {
              postgres: NOT_SUPPORTED,
              mysql: {
                sql: 'drop temporary table if exists `test`',
                parameters: [],
              },
              mssql: NOT_SUPPORTED,
              sqlite: NOT_SUPPORTED,
            })
          })
        }
```

test/node/src/schema.test.ts
```diff
@@ -1512,6 +1512,22 @@ for (const dialect of DIALECTS) {
             sqlite: NOT_SUPPORTED,
           })
         })
+
+        if (dialect === 'mysql') {
```
igalklebanov:
```suggestion
        if (sqlSpec === 'mysql') {
```

Issue #1385, opened 2025-03-13
sqlite: better-sqlite3 shouldn't be the default driver (as is), nor be used if more than one sqlite instance or other DBs are also used


DaemonSnake:
Hi,

As you know better-sqlite3 is the default sqlite drive in Kysely and is synchronous.

The issue with this is that it blocks the full event look during each query.
This make kysely async interface of it quite unorthodox as most Db clients async interfaces are expected to offload query execution outside the event loop.
Also better-sqlite3 is an opinionated client,
as can testify the `When is this library not appropriate?` ( https://www.npmjs.com/package/better-sqlite3 ).
It isn't meant to be used with many sqlite DBs and combined with async code.

This makes it a poor fit for a default driver of kysely currently.

The consequence here is that it can degrades overall performance of the application and unrelated DBs. 
In particular when the project:
- uses multiple sqlite
- contains some slow / big queries (for instance filling in temporary sqlite DB for local processing)
- have many small queries
etc.

The impact when this occurs can be catastrophic and very hard to find/debug.
For instance in our case it would:
- block unrelated pending postgresql transactions (commits getting delayed)
This would then cause `oldest_xoid_transaction_age` to increase,
degrading postgres performance and create very weird bugs in completely unrelated projects.
- same thing for queries on unrelated Sqlite DBs that would become much slower or look stuck
- prevent proper application sigterm handling / cleanup (causing incomplete / bad states)
- second long event loop delay almost constantly depending on the number of sqlite Dbs using better-sqlite3
just to mention the main ones.

This also goes against the trend of creating many small isolated sqlite DB (for instance 1 per user).
Reason being that calls to 3 unrelated DBs are done is series currently instead of in parallel.

We found the issue a while back and replace the default drive in one of place where we create temporary sqlite Dbs 
by using a custom dialect using the async lib sqlite3 ( https://www.npmjs.com/package/sqlite3 ) with a simple async mutex to prevent concurrent queries.

We thought this would be enough but forgot to migrate another cases where, we thought, it wouldn't be necessary.
After adding a query_timeout to postgresql's pool config of 1min, we discovered it was not so.

We made sure it was the source of the issue
by creating an sqlite extension that adds a function that sleep for 1 min.
Added an interval that logs and ran a query that uses the above extension.
We confirmed this way that better-sqlite queries block the event loop.

We replaced the default driver with our custom one and the before/after speaks for itself:
[...]

Issue #1292, opened 2024-12-13
alternate dialects in the multiverse of sqlite


uncomplexity:
I'm currently fine using the [better-sqlite3](https://github.com/WiseLibs/better-sqlite3) dialect, switching from postgresql to sqlite sped up my dx so fast.

Just crossed my mind though, since on postgresql it defaults to `pg` but there's an alternate one based on `postgres` which is also very cool.. on sqlite do you see any benefits on using the [nodejs internal sqlite](https://nodejs.org/api/sqlite.html) they're working on, and also on turso's [limbo](https://github.com/tursodatabase/limbo)?

#### Links mentioned

- https://github.com/WiseLibs/better-sqlite3
- https://nodejs.org/api/sqlite.html
- https://github.com/tursodatabase/limbo

jlarmstrongiv:
There are several dialects available at https://github.com/subframe7536/kysely-sqlite-tools

However, the Node.js dialect is not supported yet https://github.com/subframe7536/kysely-sqlite-tools/issues/13

arthurfiorette:
would be sick to have node's built in sqlite driver into kysely as a native dialect!

subframe7536:
I think `node:sqlite`'s api is similar to `better-sqlite3`, so maybe we can directly use `SqliteDialect` without any other efforts

igalklebanov:
It is not similar. Getting `Error: Unknown named parameter '0'` errors in tests.

mmgoodnow:
haven't played with any migrations yet, but I took a stab here. https://github.com/mmgoodnow/kysely-node-sqlite/blob/df086d566e9e9cefb0b6d8edc253a8cd428c8d2b/dialect/sqlite-dialect.ts

subframe7536:
I have made a generic dialect for sqlite: https://github.com/subframe7536/kysely-sqlite-tools/blob/master/packages/dialect-generic-sqlite

Here is the implemention for `node:sqlite`:

```ts
import { DatabaseSync } from 'node:sqlite'
import { Kysely } from 'kysely'
import { buildQueryFn, GenericSqliteDialect, parseBigInt } from 'kysely-generic-sqlite'
import type { IGenericSqlite } from 'kysely-generic-sqlite'

const db = new Kysely({
  dialect: new GenericSqliteDialect(
    () => createSqliteExecutor(new DatabaseSync(':memory:')),
  ),
})

[...]

PR #1625, opened 2025-10-31
Migrate away from `better-sqlite3` to `node:sqlite`

nikeee:
See discussion at #1292.

Opening a draft PR, so we can keep things updated while kysely or node:sqlite changes until Node.js 20 is phased out and all supported Node versions support `node:sqlite`. If it is decided to not do this migration, we can still use this dialect in a separate package.

Related https://github.com/nodejs/node/issues/57445

Maybe #1385 is related in case node:sqlite also gets an Async variant (currently, it is explicitly called DatabaseSync). See https://github.com/nodejs/node/issues/54307

No intent to merge.

### DIFF
--- a/src/dialect/sqlite/sqlite-driver.ts
+++ b/src/dialect/sqlite/sqlite-driver.ts
@@ -8,11 +8,10 @@ import { Driver } from '../../driver/driver.js'
 import { SelectQueryNode } from '../../operation-node/select-query-node.js'
 import { parseSavepointCommand } from '../../parser/savepoint-parser.js'
 import { CompiledQuery } from '../../query-compiler/compiled-query.js'
-import { QueryCompiler, RootOperationNode } from '../../query-compiler/query-compiler.js'
+import { QueryCompiler } from '../../query-compiler/query-compiler.js'
 import { freeze, isFunction } from '../../util/object-utils.js'
 import { createQueryId } from '../../util/query-id.js'
 import { SqliteDialectConfig } from './sqlite-dialect-config.js'
-import { QueryNode } from '../../operation-node/query-node.js'
 
 export class SqliteDriver implements Driver {
   readonly #config: SqliteDialectConfig


igalklebanov:
Hey 👋 

Thanks! 🙏 

This is way too early and ignores the process. 🔒 

It could live as an organizational dialect if you're up for it. WDYT?

nikeee:
No problem! Just keeping a track of the changes as the change seems to align with the goals of Kysely. No intend to get this merged any time soon. Or at all. Just here for the visibility because it seems to be what aligns with the design philosophy and what a lot of users requested.

If that's a thing that you don't want external contributors to do, that's fine too. I'm open to whatever it takes to have something like this on the long run. Feel free to amend to this branch! :)

igalklebanov:
> Just keeping a track of the changes

`node:sqlite` is not stable yet.
Node.js 20 EOL is in 30.4.2026, but will only exit the rotation on AWS Lambda (and Vercel usually) in 1.7.2026 - 8 months.

Too soon.

> If that's a thing that you don't want external contributors to do, that's fine too.

We do want external contribution. There is a process and it is important for many reasons.

nikeee:
> node:sqlite is not stable yet.

I opened this PR as a draft PR, so we can keep things updated as the Node API stabilizes while taking advantage of kysely's extensive test suite.

> It could live as an organizational dialect

That would be great and offers a nice migration path.

nikeee:
> `node:sqlite` is not stable yet. Node.js 20 EOL is in 30.4.2026, but will only exit the rotation on AWS Lambda (and Vercel usually) in 1.7.2026 - 8 months.
> 
> Too soon.

Update: node:sqlite is in RC, supported by bun and Node 20 is EOL and out of rotation in AWS (I think?).

I shipped the dialect in a separate package and I'm using it since. Let me know if you are open for it to be contributed.

PR #1637, opened 2025-11-21
feat: add minimal types to speed up type checking

juanrgm:
With these minimal types, there will no longer be performance issues during type checking, nor the classic TS2589 error (`Type instantiation is excessively deep and possibly infinite`) when attempting to create generic functions that absorb Kysely’s types.

Having TypeScript check all the complex types that existed in the `*QueryBuilder` types is very costly. By using minimal types, we avoid these checks and limit ourselves to the generic data only.

Later on, Kysely’s own code could use these minimal types to speed up type checking.


igalklebanov:
Hey 👋 

This breaks assignability, which would cause a lot of breaking for a lot of people.
It also increases instantiations for subquery cases in the order by benchmark.

PR #1658, opened 2025-12-12
[WIP] Add more options to CamelCasePlugin

### LINKED ISSUE #1627: CamelCasePlugin incorrectly cases column names starting with `_`
When using the `CamelCasePlugin` I've noticed that a column name `_id` is cased as `_Id` in my output. I would expect the leading underscore to be effectively ignored, and I'd see names like `_id`, `_fooBar`, etc.

wirekang:
This PR will add more properties to `CamelCasePluginOptions`:

| property | description |
| --- | --- |
| `ignoreSurroundingUnderscores?: boolean` | Ignore leading/trailing underscores, add as-is after conversion |
| `removeSurroudingUnderscores?: boolean` | Remove leading/trailing underscores before conversion |
| `customMapping?: Record<string,string>` | Provide custom conversion rule |

In case of #1627, I think `_id` -> `_Id` conversion is a bug. So What's the correct one? `_id` or `id`? When I tested a bunch of snake_case to camelCase implementations, the golden rule was to remove leading underscores(`_id` -> `id`). But it would be great if there are more options to control some opinioned behavior, since there is no standard.

Closes #1627

#### WIP

First of all, I added a test file to test CamelCasePlugin independently.
Since the maintainer might reject this approach, I haven't implemented anything yet and also haven't declared all the test cases.

### DIFF
--- a/test/node/src/camel-case-plugin.test.ts
+++ b/test/node/src/camel-case-plugin.test.ts
@@ -1,0 +1,119 @@
+import {
+  CamelCasePlugin,
+  CamelCasePluginOptions,
+  createQueryId,
+  CreateTableNode,
+} from '../../..'
+import { expect } from './test-setup'
+
+describe('CamelCasePlugin', () => {
+  type TestCaseWithOptions = {
+    name: string
+    options?: CamelCasePluginOptions
+    schema: [source: string, expected: string][]
+    result: [source: string, expected: string][]
+  }
+  const testCaseWithOptions: TestCaseWithOptions[] = [
+    {
+      name: 'undefined',
+      options: undefined,
+      schema: [
+        ['asis', 'asis'],
+        ['fooBar', 'foo_bar'],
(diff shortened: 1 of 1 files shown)


test/node/src/camel-case-plugin.test.ts
```diff
@@ -0,0 +1,122 @@
...
+        ['asis', 'asis'],
+        ['fooBar', 'foo_bar'],
+        ['aFOOBar', 'a_foobar'],
+        ['_fooBar', '_foo_bar'],
+      ],
+      result: [
+        ['asis', 'asis'],
+        ['foo_bar', 'fooBar'],
+        ['a_f_o_o_bar', 'aFOOBar'],
+        ['_id', '_Id'],
```
wirekang, line 43: Currently, this test case is passing. I think it should modified like `['_id', 'id']`

PR #1662, opened 2025-12-18
feat: add synchronous query execution support

### LINKED ISSUE #1556: Support synchronous execution.
Hey 👋 

There's an ergonomic advantage in synchronous APIs, especially when called in UI frameworks that don't like `async`/`await`. There's also a potential performance benefit (probably extremely small) when Promises are out of the picture.

We currently only support `async`/`await` APIs (`execute`/`executeTakeFirst`/`executeTakeFirstOrThrow`/`executeQuery`). Not ideal.

These users could apply ["splitting query building and execution"](https://kysely.dev/docs/recipes/splitting-query-building-and-execution), but it feels less "native" (not in-chain) and more complex (DIY).

We should **consider** adding methods such as:

```ts
const results = builder.executeSync()
```

...and matching executor internals.

Dialects will be able to optionally support it, and declare they do in their adapters.

OR..

Provide a standard "sync executor" helper they could import and use that applies "splitting query building and execution".

lobomfz:
closes #1556

this PR adds sync execution methods for dialects that support it (SQLite with better-sqlite3).

- `executeSync()`, `executeTakeFirstSync()`, `executeTakeFirstOrThrowSync()` on all query builders
- `executeQuerySync()` on Kysely class
- `executeSync()` on RawBuilder
- `supportsSyncExecution` flag on DialectAdapter
- `transformResultSync` on KyselyPlugin
- `SyncNotSupportedError` and `PluginSyncNotSupportedError` errors

this duplicates a lot of methods which isn't ideal, so i'm open to suggestions on how to reduce that

### DIFF
--- a/src/dialect/dialect-adapter-base.ts
+++ b/src/dialect/dialect-adapter-base.ts
@@ -24,6 +24,10 @@ export abstract class DialectAdapterBase implements DialectAdapter {
     return false
   }
 
+  get supportsSyncExecution(): boolean {
+    return false
+  }
+
   abstract acquireMigrationLock(
     db: Kysely<any>,
     options: MigrationLockOptions,
--- a/src/dialect/dialect-adapter.ts
+++ b/src/dialect/dialect-adapter.ts
@@ -38,6 +38,11 @@ export interface DialectAdapter {
   readonly supportsOutput?: boolean
 
   /**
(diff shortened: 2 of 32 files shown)


zemjr:
Hi, is it possible to implement this? We would like to use kysely with cloudflare durable object sqlite, wich is synchronous.

Issue #1154, opened 2024-09-24
bug: Multiple migrations are run in a single transaction


shoooe:
If you have a migration 0001 with this:

```
export async function up(db: Kysely<unknown>): Promise<void> {
  await db.schema.createType("some_enum").asEnum(["value_1", "value_2"]).execute();
}
```

and a migration 0002 with this:

```
export async function up(db: Kysely<unknown>): Promise<void> {
  await sql`ALTER TYPE some_enum ADD VALUE 'value_3';`.execute(
    db
  );
}
```

and then some migration 0003 that uses `some_enum` with value `value_3` then:
 - running migration 0001, 0002 and then at a later time 0003 will succeed
 - running all three migrations in one go will result in this error:

> error: unsafe use of new value "value_3" of enum type some_enum

From what I understand from the code this is caused by `migrateToLatest` running all pending migrations in a single transaction. 

Shouldn't each migration be independent and run in its own transaction? 
IME this is the behaviour that most other migration systems adopt (e.g. Django migrations).

Despite this ^ I'm loving this library more and more every day, so thank you again for this magical tool. ❤️

koskimas:
Having individual transaction for each migration would certainly make other things easier too. For example we could easily implement disabling transactions for individual migrations. There has been a feature request open for that forever.

However, there was a reason why it was implemented the way it is. I just can't remember that right now.

Changing the behaviour could also be a major breaking change and I don't know if we can do it without a massive deprecation period.

shoooe:
Oh true. For reference to others this is what I went with:

```typescript
  const migrations = await migrator.getMigrations();
  const pendingMigrations = migrations.filter((m) => isUndefined(m.executedAt));

  if (isEmpty(pendingMigrations)) {
    console.info(`✅ Migrations are up to date!`);
  }

  for (const migration of pendingMigrations) {
    const { error, results } = await migrator.migrateTo(migration.name);

    results?.map((it) => {
      if (it.status === "Success") {
        console.info(
          `✅ Migration "${it.migrationName}" was executed successfully`
        );
      } else if (it.status === "Error") {
        console.error(`❌ Failed to execute migration "${it.migrationName}"`);
      }
    });

    if (error) {
      console.error("❌ Failed to migrate");
      console.error(error);
      process.exit(1);
    }
  }
```

koskimas:
Let's leave this open. It might make sense to have individual transactions.

igalklebanov:
[...]

PR #1671, opened 2026-01-03
feat: Per-migration transaction configuration

### LINKED ISSUE #352: feature request: non-transactional DDL for concurrent index creation

lourd:
Closes #352

Creates the ability for individual migrations to opt out of being executed in a transaction, depending on the adapter's support for transactions, the Migrator configuration, and the options given to the migrate method.

Follows pretty closely to the Knex implementation as suggested by @tgriesser.

If transactions are enabled globally and an individual migration opts out, then all of the migrations before it are run in a transaction, then it is run outside of a transaction, then all the migrations after it are run in a transaction.

The behavior is the same when migrating up and down.

I decided that if a transaction is passed to the Migrator (#1480), it results in an error. And if any of the options disabling transactions are used, either globally or per migrate call (#1517), those take precedence.

Basically this feature only takes effect when transactions aren't disabled and the Migrator has responsibility over transactions.


lourd:
I'm pretty sure using this feature is incompatible with passing in a transaction to the Migrator #1480. Looking for direction on what the precedence logic should be for that, as well as the option introduced in #1517

lourd:
Hey @igalklebanov, eager to hear your feedback!

Philipinho:
Great work @lourd.

lourd:
Hey @igalklebanov, just rebased on `next`, should be good to go here. Let me know if there's anything I can do to help making reviewing or merging easier

src/migration/migrator.ts
```diff
@@ -23,6 +26,36 @@ export interface Migration {
...
+   * When transactions are enabled globally, consecutive migrations with the same
+   * transaction behavior are batched together in a single transaction.
+   *
+   * When transactions are disabled globally, each migration with `transaction: true`
+   * runs in its own separate transaction.
+   *
+   * Note: Setting this to `true` has no effect if the dialect doesn't support
+   * transactional DDL.
+   */
+  readonly transaction?: boolean
```
igalklebanov: If there wasn't such a strong case to keep `knex` naming and behavior - familiarity and lots of training data, I'd call this something more explicit and perhaps go beyond `boolean`. 

e.g. `transactionMode?: 'never' | 'shared' | 'file-scoped'`, where "never" means no matter what, never runs this file's changes in a transaction, "shared" means reuse or create "global" transaction (the default mode), "file-scoped" which means always commit previous transaction group if available and start a transaction just for this file.

Not for now. Something to consider after this effort.

lourd: I think that's a great idea, covers more use cases than a boolean can. Happy to make those changes, let me know!

igalklebanov: Not for this iteration. Let's keep it `knex`y and minimal as possible.

shoooe:
@lourd @igalklebanov Anything I can do to help?

lourd:
@shoooe, thanks for the offer. Last I talked with @igalklebanov this is slated for 0.30, he wanted to get 0.29 out first. I need to rebase on latest and resolve the merge conflicts.

shoooe:
Alright, thanks very much for working on this @lourd @igalklebanov. 🙏

igalklebanov (approved):
Hey 👋 

Thanks for this PR! 🙏 

It provided a solid, working and tested implementation. 💪 

The heart of it - `config.transaction` escape hatch, erroring when the migrator is handed a transaction, symmetric up/down handling - survived intact in what I eventually pushed. I did,  however, reshape the transaction-scoping model around it pretty substantially:

- The transaction batching from the original design (migrations grouped into shared transactions on either side of an opted-out one) is gone. In its place there's an explicit `transactionMode: 'per-run' | 'per-migration' | 'none'` on `MigratorProps`/`MigrateOptions` (call-level takes precedence over instance-level). `disableTransactions` stays as a deprecated alias for `'none'`.

- `config: { transaction: false }` is only honored when `transactionMode` is `'per-migration'`. In `'per-run'`/`'none'`, any per-migration config is an error that names the fix, rather than silently changing how the run is grouped.

- Explicitly requesting a transactional mode on dialects without transactional DDL (MySQL, SQLite) errors instead of silently no-opping.

- Default behavior is unchanged (`'per-run'` = today's whole-run transactions). Multi-migration runs without an explicit mode log a one-time notice that the default might change to `'per-migration'` in a future version.

I really wanted "islands" of atomicity to be a thing. But after digging into the pains and lessons learned throughout the industry - `knex`, `typeorm`, `mikro-orm`, `drizzle-kit`, `prisma`, Rails, Flyway, Liquibase, Goose, Alembic, etc. - I became convinced that any transaction shared across migrations is a footgun: 

- It increases the odds of failures that a commit in-between would have prevented - e.g. PostgreSQL enum value adds, foreign-key triggers blocking table alterations, etc.

- It makes those failures environment-dependent (e.g. local development vs. CI) where the pace/batching in which migrations are applied can make them succeed or fail unexpectedly. This is the most reported migration pain in all the tools I mentioned.

Kysely is built on explicitness and safety. Silently regrouping the run when a `transaction: false` shows up didn't align with that - and it also conflicted with @koskimas' take in #1154 - per-migration config only honored when per-migration setting is enabled. I also agree with the fact our default behavior shares the same problems and might need changing to `'per-migration'` after a long deprecation period which a new warning log hints at.

I have also added much needed documentation in the site, and a `defineMigration` helper to help adhere to the migration contract.

lourd:
Makes great sense @igalklebanov, thanks for the thorough explanation

PR #1672, opened 2026-01-04
Fix/1669 generic string references (attempt #2)

### LINKED ISSUE #1669: Generic helper with constrained table set rejects string column refs (ReferenceExpression invariance?)

briandastous:
@koskimas, @igalklebanov: Sorry for the churn with https://github.com/kysely-org/kysely/pull/1670. Hopefully this one is more reviewable. This PR touches a lot less code, and I have tried adding more targeted test coverage. If this still is not up to snuff, I get it, I do not want to waste your time.

Fixes #1669.

## Problem

Generic helpers over table unions are rejected, which forces duplication for every join shape. Minimal failure case:

```ts
interface DB {
  person: { id: number; name: string }
  pet: { id: number; owner_id: number }
  toy: { id: number; owner_id: number }
}

function wherePersonHasName<TB extends Exclude<keyof DB, 'person'>, O>(
  qb: SelectQueryBuilder<DB, 'person' | TB, O>,
  name: ExtractTypeFromReferenceExpression<DB, TB | 'person', 'person.name'>,
) {
  return qb.where('person.name', '=', name)
}

declare const withPet: SelectQueryBuilder<DB, 'person' | 'pet', { id: number }>
declare const withToy: SelectQueryBuilder<DB, 'person' | 'toy', { id: number }>

wherePersonHasName(withPet, 'alice')
wherePersonHasName(withToy, 'alice')
```

Even though `person` is explicitly included in the table union, `person.name` is not recognized as a `ReferenceExpression<DB, TB | 'person'>` inside the helper under the old types.

## Solution

Make `AnyColumn`, `ExtractColumnType`, and `AnyColumnWithTable` distributive across `TB` so unions like `TB | 'person'` preserve concrete `person.<col>` string-literal members (with `NonNullable` for optional table shapes). This is a type-only change.

## Compatibility note

This is a type-level tightening driven by `AnyColumn`, `ExtractColumnType`, and `AnyColumnWithTable` now distributing over `TB` (and applying `NonNullable` to table shapes). That makes table/column correlations precise, but can surface new compile errors where helpers were typed against a narrower DB than the caller (e.g. `DB` vs `DB & { extra_table }`). Runtime behavior is unchanged; fix by making helpers generic in `DBT extends DB`.

```ts
// before (can fail once DB is widened)
function insertPerson(qb: InsertQueryBuilder<DB, 'person', InsertResult>) { ... }

// after (preserves the caller's DB)
function insertPerson<DBT extends DB>(
  qb: InsertQueryBuilder<DBT, 'person', InsertResult>,
) { ... }
```

## Tests

Added or updated typing coverage:

- `test/typings/test-d/generic-string-references.test-d.ts`:
  - `AnyColumn` / `AnyColumnWithTable` assignability and error cases, including correlated table/column rejections.
  - `ExtractTypeFromReferenceExpression` for table-qualified and nullable table shapes.
  - Generic helper usage across multiple join shapes without duplicating filters (TB-generic).

- `test/typings/test-d/generic.test-d.ts`:
  - DBT-generic helper case for `person.name` references (excluded for TS < 5.4 via `scripts/exclude-test-files-for-backwards-compat.mts`).

Implementation note:
- Updated `test/node/src/test-setup.ts` and `test/node/src/with-schema.test.ts` to use a DB-generic insert helper because the stricter table/column correlations make narrower-DB helpers reject widened DBs (no casts added).

Tests run:

- `pnpm -C tmp/kysely-1669 test`

### DIFF
--- a/src/util/type-utils.ts
+++ b/src/util/type-utils.ts
@@ -35,17 +35,18 @@ import type { MergeResult } from '../query-builder/merge-result.js'
  * // Columns == 'id' | 'name' | 'species'
  * ```
  */
-export type AnyColumn<DB, TB extends keyof DB> = {
-  [T in TB]: keyof DB[T]
-}[TB] &
-  string
+export type AnyColumn<DB, TB extends keyof DB> = TB extends any
+  ? keyof NonNullable<DB[TB]> & string
+  : never
 
 /**
  * Extracts a column type.
  */
-export type ExtractColumnType<DB, TB extends keyof DB, C> = {
-  [T in TB]: C extends keyof DB[T] ? DB[T][C] : never
-}[TB]
+export type ExtractColumnType<DB, TB extends keyof DB, C> = TB extends any
+  ? C extends keyof NonNullable<DB[TB]>
+    ? NonNullable<DB[TB]>[C]
+    : never
+  : never
 
 /**
  * Given a database type and a union of table names in that db, returns
@@ -78,9 +79,9 @@ export type ExtractColumnType<DB, TB extends keyof DB, C> = {
  * // Columns == 'person.id' | 'pet.name' | 'pet.species'
  * ```
  */
-export type AnyColumnWithTable<DB, TB extends keyof DB> = {
-  [T in TB]: `${T & string}.${keyof DB[T] & string}`
-}[TB]
+export type AnyColumnWithTable<DB, TB extends keyof DB> = TB extends any
+  ? `${TB & string}.${AnyColumn<DB, TB>}`
+  : never
 
 /**
  * Just like {@link AnyColumn} but with a ` as <string>` suffix.
--- a/test/node/src/test-setup.ts
+++ b/test/node/src/test-setup.ts
@@ -234,17 +234,66 @@ export async function destroyTest(ctx: TestContext): Promise<void> {
   await ctx.db.destroy()
 }
 
+type TablesWithId<DB> = {
+  [K in keyof DB]: 'id' extends keyof NonNullable<DB[K]> ? K : never
+}[keyof DB]
+
+// Helper for TS 4.7/4.8 to narrow the `id in row` check.
+function hasId(row: object): row is { id: unknown } {
+  return Object.prototype.hasOwnProperty.call(row, 'id')
+}
+
+// Runtime guard to enforce numeric ids in test helpers.
+function assertHasNumericId<Row extends object>(
+  row: Row,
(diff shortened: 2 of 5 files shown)


koskimas:
Hmm.. This looks too good to be true. I'm 100% sure I tried to implement it like that and hit a bunch of issues. But the typing tests seem to pass. We'll need to really test this, but if this really works, it should be merged.

igalklebanov:
There's an instantiations drop in the order by benchmark which is a positive.
The changes in test setup are extremely suspicious. Smells like something breaking elsewhere.

EDIT:
Yep.

PR #1676, opened 2026-01-13
feat: Add `NULLIF()` implementation

### LINKED ISSUE #1429: Add `nullif(value1, value2)`.
Hey 👋 

Essentially, it's a shorthand for:

```ts
eb
  .case()
  .when('col1', '=', eb.ref('col2'))
  .then(null)
  .else(eb.ref('col1'))
  .end()
```

and exists in all major databases.

bakasmarius:
Closes #1429.
Sorry it took longer than I expected, it wasn't as easy to implement as I initially thought 😅 
Any comments with suggestions to edit/rename/refactor/test/etc are welcome 😉


src/operation-node/null-if-node.ts
igalklebanov, line 1: There's no need for a node for this. It's a function, nothing special.

`src/operation-node/function-node.ts` is all that's needed.

src/operation-node/operator-node.ts
```diff
@@ -73,6 +73,7 @@ export const OPERATORS = [
   ...UNARY_OPERATORS,
   'between',
   'between symmetric',
+  'nullif',
```
igalklebanov: it's not an operator. it's a function.

src/operation-node/select-query-node.ts
```diff
@@ -24,6 +24,7 @@ export interface SelectQueryNode extends OperationNode {
   readonly from?: FromNode
   readonly selections?: ReadonlyArray<SelectionNode>
   readonly distinctOn?: ReadonlyArray<OperationNode>
+  readonly nullIf?: OperationNode
```
igalklebanov: what?! it's not a clause. it's a function.

e.g. if it's used in `.select`, it should be in the `selections` array.

test/node/src/null-if.test.ts
```diff
@@ -0,0 +1,89 @@
...
+
+      expect(result).to.eql({
+        first_name: 'Jennifer',
+        null_dates_equal: null,
+        null_dates_different: '2000-01-01',
+        null_numbers_equal: null,
+        null_numbers_different: dialect === 'mysql' ? '1' : 1,
+        null_strings_equal: null,
+        null_strings_different: 'yes',
+        null_nulls_equal: 'yes',
```
igalklebanov: unnecessarily testing the most basic logical function for different data types.

test/node/src/null-if.test.ts
```diff
@@ -0,0 +1,89 @@
...
+    it('should return correct values for NULLIF without selection from a table', async () => {
+      const result = await db
+        .selectNoFrom((eb) => [
+          eb.nullif('2000-01-01', '2000-01-01').as('null_dates_equal'),
+          eb.nullif('2000-01-01', '2000-01-02').as('null_dates_different'),
+          eb.nullif(1, 1).as('null_numbers_equal'),
+          eb.nullif(1, 2).as('null_numbers_different'),
+          eb.nullif('yes', 'yes').as('null_strings_equal'),
+          eb.nullif('yes', 'no').as('null_strings_different'),
+          eb.nullif('yes', null).as('null_nulls_equal'),
```
igalklebanov: that name makes no sense. did you mean to do `eb.nullif(null, null)`?

test/node/src/null-if.test.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    before(async function () {
+      ctx = await initTest(this, dialect)
+
+      db = ctx.db as any
+    })
+
+    beforeEach(async () => {
+      await insertDefaultDataSet(ctx)
+    })
```
igalklebanov: not necessary, could just select without a `from` clause.

src/expression/expression-builder.ts
```diff
@@ -1115,6 +1117,33 @@ export interface ExpressionBuilder<DB, TB extends keyof DB> {
...
+   * ```sql
+   * select first_name, nullif(1, 1) as null_if_result from "person"
+   * ```
+   */
+  nullif<
+    T extends ReferenceExpression<DB, TB> | number | boolean | string | null,
+  >(
+    expr1: T,
+    expr2: T,
+  ): ExpressionWrapper<DB, TB, SqlBool>
```
igalklebanov: should be moved to `FunctionModule` for consistency.

src/expression/expression-builder.ts
```diff
@@ -1115,6 +1117,33 @@ export interface ExpressionBuilder<DB, TB extends keyof DB> {
...
+   * ```
+   *
+   * The generated SQL (PostgreSQL):
+   *
+   * ```sql
+   * select first_name, nullif(1, 1) as null_if_result from "person"
+   * ```
+   */
+  nullif<
+    T extends ReferenceExpression<DB, TB> | number | boolean | string | null,
```
igalklebanov: there's a bad mix of `ReferenceExpression` and `string` here.

there won't be autocompletion for references cause it's widened to `string`.

`ReferenceExpression` covers both string references to columns in query context AND `sql.val` and `sql.lit` usage. The original issue suggested the API to be ref-first for this. So, you can keep only `ReferenceExpression`.

src/expression/expression-builder.ts
```diff
@@ -1115,6 +1117,33 @@ export interface ExpressionBuilder<DB, TB extends keyof DB> {
...
+  >(
+    expr1: T,
+    expr2: T,
```
igalklebanov: APIs with `<T>(p0: T, p1: T)` are locked to the type of `p0`. This is not what you want here. You want to allow passing something assignable to `p0` in `p1`.

src/expression/expression-builder.ts
igalklebanov: `SqlBool` is wrong. It is first parameter or null.

src/query-builder/function-module.ts
```diff
@@ -860,5 +891,25 @@ export function createFunctionModule<DB, TB extends keyof DB>(): FunctionModule<
         ]),
       )
     },
+
+    nullif<T extends ReferenceExpression<DB, TB>>(
+      expr1: T,
+      expr2: number | boolean | string | null,
+    ): ExpressionWrapper<DB, TB, T | null> {
+      var v1 =
```
aymenhmaidiwastaken:
```suggestion
      const v1 =
```

src/query-builder/function-module.ts
```diff
@@ -860,5 +891,25 @@ export function createFunctionModule<DB, TB extends keyof DB>(): FunctionModule<
...
+        typeof expr1 === 'string'
+          ? parseValueExpression(expr1)
+          : isSafeImmediateValue(expr1)
+            ? parseSafeImmediateValue(expr1)
+            : parseReferenceExpression(expr1)
+      var v2 =
```
aymenhmaidiwastaken:
```suggestion
      const v2 =
```

aymenhmaidiwastaken (commented):
Small nit: the rest of the codebase uses `const`/`let` exclusively. These `var` declarations should be `const` since they're never reassigned.

src/query-builder/function-module.ts
bakasmarius: I wonder how I missed that 😄 
@igalklebanov maybe add some lint rules for that?

igalklebanov: it's on the list.

src/operation-node/operator-node.ts
igalklebanov: why was this resolved without change?

bakasmarius: I honestly don't remember, it's fixed now.

Issue #1115, opened 2024-08-12
Type inferration error on dynamic selection


dotMortis:
Typescript v5.4 - v5.5 (have not tested older versions)

tsconfig.json
```json
{
    "compilerOptions": {
        "allowSyntheticDefaultImports": true,
        "strictPropertyInitialization": true,
        "esModuleInterop": true,
        "skipLibCheck": true,
        "forceConsistentCasingInFileNames": true,
        "importHelpers": true,
        "sourceMap": true,
        "noImplicitAny": true,
        "declaration": false,
        "noFallthroughCasesInSwitch": true,
        "noImplicitReturns": true,
        "pretty": true,
        "strict": true,
        "strictNullChecks": true,
        "target": "ES2022",
        "module": "NodeNext",
        "lib": ["ES2022"],
        "resolveJsonModule": true,
        "outDir": "./dist",
        "declaration": true
    },
    "ts-node": {
        "esm": true
    }
}
```

The most important setting ist `"decleration": true"` otherwise the error wont show.

```typescript
import { Generated, Kysely, SelectExpression } from 'kysely';

type DB = {
    test: {
        id: Generated<number>;
        type: 'a' | 'b';
        updatedAt: number;
        createdAt: number;
    };
};

export class WorkflowTypeRepository {
    private readonly _db: Kysely<DB>;

    constructor(db: Kysely<DB>) {
        this._db = db;
    }

    all<SE extends SelectExpression<DB, 'test'>>(selections: ReadonlyArray<SE>) {
        return this._db.selectFrom('test').select(selections).execute();
    }
}
```

The goal is to create a function that allows me to select only fields I really need.
In the screenshots below its shown that the generic selection has been inferred correctly, but tsc can't build declerations for it.
It's necessary for me to be able to build the .d.ts files since it's a package in a monorepo.

Errors:
- The inferred type of 'all' cannot be named without a reference to '../../node_modules/kysely/dist/esm/dynamic/dynamic-reference-builder.js'. This is likely not portable. A type annotation is necessary.ts(2742)
- The inferred type of 'all' cannot be named without a reference to '../../node_modules/kysely/dist/esm/util/type-utils.js'. This is likely not portable. A type annotation is necessary.ts(2742)

![Screenshot 2024-08-12 203208](https://github.com/user-attachments/assets/f173d71e-72f9-40d0-bed4-ae9e9e02c558)
[...]

PR #1679, opened 2026-01-14
Add `ExtractColumnType` and `DrainOuterGeneric` type exports

mifi:
Fixes error seen in https://github.com/kysely-org/kysely/issues/1115#issuecomment-3442356231

```
The inferred type of '...' cannot be named without a reference to 'node_modules/kysely/dist/esm/util/type-utils.js'. This is likely not portable. A type annotation is necessary.
```

(when using `composite: true` in typescript)


igalklebanov:
Hey 👋 

Thanks! 🙏 

Can you add an example in `/test/composite-ts` that would fail otherwise?

mifi:
hi! thanks, i've added a file. i believe it will fail with such a simple case

test/composite-ts/jsonObjectFrom.mts
```diff
@@ -0,0 +1,24 @@
+import { Kysely } from 'kysely'
```
igalklebanov:
```suggestion
import { Kysely } from 'kysely'
import { jsonObjectFrom } from 'kysely/helpers/postgres'
```

PR #1691, opened 2026-01-25
fix: sql dialect returning order issue

### LINKED ISSUE #1631: `returning` needs to come before `order by` in sqlite for an `update..where` query.
Kysely appears to always put `returning` at the end, however that will cause an error.  See section `2.3` here https://sqlite.org/lang_update.html

This may be the relevant code:

```
        if (node.orderBy) {
            this.append(' ');
            this.visitNode(node.orderBy);
        }
        if (node.limit) {
            this.append(' ');
            this.visitNode(node.limit);
        }
        if (node.returning) {
            this.append(' ');
            this.visitNode(node.returning);
        }
```

hwisu:
Closes #1631

Howdy,
I made a small fix and added one test case.
I tried to get assigned, but since it looked like a small fix, I went ahead and opened a PR first.
Thanks for your understanding!

naive reproduction : https://github.com/hwisu/kysely-1631
discord request (from me) : https://discord.com/channels/890118421587578920/1073949928453328897/1462700413253062666

### DIFF
--- a/test/node/src/update.test.ts
+++ b/test/node/src/update.test.ts
@@ -436,21 +436,22 @@ for (const dialect of DIALECTS) {
     }
 
     if (dialect === 'sqlite') {
-      it('should place returning before order by and limit', async () => {
+      it('should order and limit the amount of updated rows and return', async () => {
         const query = ctx.db
           .updateTable('person')
           .set({ last_name: 'Barson' })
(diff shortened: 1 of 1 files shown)


hwisu:
Thanks for the feedback! Appreciate the cleanup and the simpler approach here. :taco:

PR #1696, opened 2026-02-01
refactor: bump minimum TypeScript version to 4.7.

igalklebanov:
Hey 👋 

This PR bumps the minimum supported TypeScript version to 4.7. 

This TypeScript version introduced `infer T extends S` syntax which can simplify/shorten types, and variance annotations (`in T, out S, in out R`) which can improve compile-time performance.

It adds some missing type benchmarks.

`infer T extends S` syntax was proven to result in wrong types OR slower types, at least with the types we currently have.

Variance annotations have not shown any significant improvement in instantiation counts or wall clock compilation times. Adding `out` to `Expression<T>` did slightly reduce instantiations (<1%) in some sub-query methods.

### DIFF
--- a/test/ts-benchmarks/with.bench.ts
+++ b/test/ts-benchmarks/with.bench.ts
@@ -89,15 +89,15 @@ bench('kysely.with(cte, selectQuery)', () => {
 
 bench('kysely.with(cte, insertQuery)', () => {
   return kysely.with('cte', kysely.insertInto('my_table').returningAll())
-}).types([5501, 'instantiations'])
+}).types([3562, 'instantiations'])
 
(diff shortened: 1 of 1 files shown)


test/ts-benchmarks/delete-from.bench.ts
igalklebanov, line 1: done to test `infer T extends keyof DB & string` in `DeleteFrom` type. found out the change slows it down significantly.

test/ts-benchmarks/merge-into.bench.ts
igalklebanov, line 1: done to test infer T extends keyof DB & string in `MergeInto` type. Found out the change slows it down.

PR #1697, opened 2026-02-03
feat: add `NonDehydrateable<T>` to allow opt-out from dehydration in JSON functions/helpers.

### LINKED ISSUE #1680: `ShallowDehydrateObject` bug
When I use `jsonArrayFrom` with a union string type as a database table property, it gets transformed to `number` which seems completely wrong to me. See example:
https://kyse.link/yOPTB

I tried directly with `ShallowDehydrateObject` and i can repro the same:

```ts
import { ShallowDehydrateObject } from "kysely";

type Input = {
  b: '1' | '2',
}

type Output = ShallowDehydrateObject<Input>
/*
type Output = {
    b: number;
}
*/
```

I believe the implementation of ShallowDehydrateObject is not adequate. Maybe we can replace it with something like `type-fest`'s `jsonify` instead? It is designed exactly for this and it resolves correctly:

```ts
import type { Jsonify } from 'type-fest'

type Input = {
  b: '1' | '2',
}

type Output = Jsonify<Input>
/*
type Output = {
    b: '1' | '2';
}
*/
```

or if you don't want to add another depedency, maybe just copy the code from there: https://github.com/sindresorhus/type-fest/blob/main/source/jsonify.d.ts

igalklebanov:
Hey 👋 

closes #1680.

This PR introduces `NonDehydrateable<T>` as a way to define a column that doesn't get dehydrated when referenced in JSON functions/helpers.

```ts
import type { NonDehydrateable } from 'kysely'

interface Database {
  my_table: {
    a_column: '1' | '2' | '3',
    another_column: NonDehydrateable<'1' | '2' | '3'>,
    column_too: NonDehydrateable<ColumnType<'1' | '2' | '3'>>
  }
}

const result = await db
  .selectNoFrom(
    jsonArrayFrom(db.selectFrom('my_table').selectAll()).as('dehydrated'),
  )
  .executeTakeFirstOrThrow()

result.dehydrated.a_column satisfies number // dehydrated to number!
result.dehydrated.another_column satisfies NumericString // not dehydrated to number!
result.dehydrated.column_too satisfies NumericString // not dehydrated to number!
```

### DIFF
--- a/src/util/type-utils.ts
+++ b/src/util/type-utils.ts
@@ -236,7 +236,7 @@ export type ShallowDehydrateObject<O> = {
  */
 export type ShallowDehydrateValue<T> = T extends null | undefined
   ? T
-  : NonNullable<T> extends { __kysely_dehydrate__?: false }
+  : '__kysely_dehydrate__' extends keyof NonNullable<T>
     ? T
     : T extends (infer U)[] | null | undefined
       ? Array<ShallowDehydrateValue<U>> | Extract<T, null | undefined>


mifi:
Cool thanks! Works nicely in my code. I did also try something very similar (using AI) which also worked for me: https://github.com/mifi/kysely/commit/89ed01b12933191097812fc21729c70aabe5b038?w=1

but I have no idea what's the difference between my implementation and yours. Shared it because I figured maybe you can get some inspiration from it.

igalklebanov:
> but I have no idea what's the difference between my implementation and yours. Shared it because I figured maybe you can get some inspiration from it.

Using a symbol for this can be problematic once hitting userland - problems with duplicate Kyselys that have surfaced over the years that cause comparisons by reference to fail.

igalklebanov:
Also, you don't want to decorate `ColumnType` and similar, you want to decorate the `SelectType` inside it, given `ColumnType` gets lost.

mifi:
sounds good. I guess nobody could ever have a column type which is an object with a key named `__dehydrate__` so it shouldn't be a problem anyways

igalklebanov:
Added a `__kysely_` prefix just in case.

mifi:
Although this works great, one thing I did notice is that when I look at the properties returned by `execute()` from fields defined with `NonDehydrateable` when queried without jsonArrayFrom/jsonObjectFrom, they still have the property `__kysely_dehydrate__` attached to them. For example:
```
courseCode: ("1" | "2" | "3" | "L" | "Y" | "S") & {
    __kysely_dehydrate__?: false | undefined;
}
```

It hasn't caused any issues for me (yet), but it just looks a bit confusing when inspecting the type in my IDE. Maybe we can somehow strip this property when `execute` is called?

igalklebanov:
Adding a stripping step in the result types would hurt everyone's type performance for the sake of a rare use case.

If this becomes a problem, we could look into adding a helper type you could use optionally OR even a type-level method you could invoke to use it in-chain.

PR #1698, opened 2026-02-04
feat: deny passing `BLOB`-like columns to SQLite JSON helpers

### LINKED ISSUE #1687: Deny passing `Buffer` columns to SQLite JSON helpers.
Hey 👋 

SQLite errors with `SqliteError: JSON cannot hold BLOB values` when a binary typed column or literal are passed to JSON functions.

The usual workaround is to omit such values from the data passed to these functions, or to cast the binary values to text or similar.

Currently, our SQLite JSON helpers allow passing such values. They shouldn't.

It started happening in SQLite version `3.45.0` (Jan 15th 2024) with the introduction of JSONB. 

For Kysely users, this started with `better-sqlite3@9.4.0` (Feb 1st 2024).

The minimum Node.js version we support is `20`. It is supported in `better-sqlite3@^8.4.0 || ^9.0.0`. Its EOL comes in April 30th, 2026.

So:

Breaking change for folks using `better-sqlite3@<9.4.0`.
Bug fix for folks using `better-sqlite3@^9.4.0`.

hwisu:
Hey 👋 

closes #1687.

This PR adds input checks that compile-time error in SQLite's JSON helpers when a `BLOB`-like column exists in the result object type.

It uses a localized `ExtractBlobs` helper type to check whether those objects have at least 1 `BLOB`-like column, and if so, `KyselyTypeError`s the input argument.

----

Original post:

Howdy 🖖,

This PR adds compile-time type checks to SQLite JSON helpers (`jsonArrayFrom`, `jsonObjectFrom`, `jsonBuildObject`) to reject `Buffer`/`Uint8Array` columns.

Passing binary values to JSON functions throws `SqliteError: JSON cannot hold BLOB values`. This change catches common cases at compile time, avoiding the runtime error when using these helpers.

  ### Changes

- Added `HasUint8Array<T>` and `ObjectHasUint8ArrayProperty<O>` type utilities
- SQLite JSON helpers now return `KyselyTypeError` when binary columns are detected
- Error message includes workarounds: `eb.cast(column, "text")` or `` sql`hex(column)` ``

### DIFF
--- a/src/helpers/sqlite.ts
+++ b/src/helpers/sqlite.ts
@@ -3,12 +3,23 @@ import type { SelectQueryNode } from '../operation-node/select-query-node.js'
 import type { SelectQueryBuilderExpression } from '../query-builder/select-query-builder-expression.js'
 import type { RawBuilder } from '../raw-builder/raw-builder.js'
 import { sql } from '../raw-builder/sql.js'
+import type { KyselyTypeError } from '../util/type-error.js'
 import { getJsonObjectArgs } from '../util/json-object-args.js'
 import type {
+  HasUint8Array,
+  ObjectHasUint8ArrayProperty,
   ShallowDehydrateObject,
   ShallowDehydrateValue,
   Simplify,
 } from '../util/type-utils.js'
+
+type ExpressionRecordHasUint8Array<
+  O extends Record<string, Expression<unknown>>,
+> = true extends {
+  [K in keyof O]: O[K] extends Expression<infer V> ? HasUint8Array<V> : false
+}[keyof O]
+  ? true
+  : false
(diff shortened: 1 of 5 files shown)


igalklebanov (approved):
Hey 👋 

Thanks! 🙏 

Good stuff! Massaged it a bit to provide feedback closer to the issue, and use simpler localized types.

hwisu:
I think adding things to type-utils was a bit rash of me. I wasn't thorough enough with any or the tests either. The fact that type inference didn't work at the function call site was also a problem. I learned a lot from your revisions. I hope I can make a better contribution next time. Thank you!

PR #1699, opened 2026-02-05
fix: emit ON COMMIT clause for CREATE TABLE AS form

### LINKED ISSUE #1629: ON COMMIT clause not emitted for CREATE TABLE AS form
```js
db.schema.createTable('b').temporary().onCommit('drop').addColumn('foo', 'text').compile().sql
```
compiles to 
```sql
create temporary table "b" ("foo" text) on commit drop
```
while
```js
db.schema.createTable('a').temporary().onCommit('drop').as(db.selectNoFrom(sql`'foo'`.as('foo'))).compile().sql
```
compiles to 
```sql
create temporary table "a" as select 'foo' as "foo"
```

version: 0.28.8

kaigritun:
## Summary

Fixes #1629

When using `.as()` to create a table from a select query, the `onCommit` clause was not being emitted in the compiled SQL.

### Before (broken)
```typescript
db.schema.createTable('a')
  .temporary()
  .onCommit('drop')
  .as(db.selectNoFrom(sql\`'foo' as "foo"\`))
  .compile().sql
// => create temporary table "a" as select 'foo' as "foo"
//    ^ missing ON COMMIT DROP
```

### After (fixed)
```typescript
// => create temporary table "a" as select 'foo' as "foo" on commit drop
```

## Changes

The `onCommit` and `endModifiers` handling was incorrectly nested inside the `else` block (for column definitions). Moved it outside the if/else so it's emitted regardless of whether we're using columns or a select query.

## Testing

Added a test case for `CREATE TABLE AS` with `onCommit('drop')` in the schema tests.

### DIFF
--- a/src/query-compiler/default-query-compiler.ts
+++ b/src/query-compiler/default-query-compiler.ts
@@ -639,16 +639,16 @@ export class DefaultQueryCompiler
       this.append(' (')
       this.compileList([...node.columns, ...(node.constraints ?? [])])
       this.append(')')
+    }
 
-      if (node.onCommit) {
-        this.append(' on commit ')
-        this.append(node.onCommit)
-      }
+    if (node.onCommit) {
+      this.append(' on commit ')
+      this.append(node.onCommit)
+    }
 
-      if (node.endModifiers && node.endModifiers.length > 0) {
-        this.append(' ')
-        this.compileList(node.endModifiers, ' ')
-      }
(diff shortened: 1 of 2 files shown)


igalklebanov:
https://github.com/kysely-org/kysely/blob/4f43c7d2280eabb586d1db4b4ae1361aa99eca3f/CONTRIBUTING.md

PR #1700, opened 2026-02-05
feat(postgres): add ALTER TABLE IF EXISTS support

### LINKED ISSUE #1272: Support Postgres's `ALTER TABLE IF EXISTS`
Would it be possible to support the `IF EXISTS` component for `ALTER TABLE` statements in Postgres?

https://www.postgresql.org/docs/current/sql-altertable.html

wotan-allfather:
Add support for the PostgreSQL-specific `IF EXISTS` modifier in ALTER TABLE statements. This allows queries to silently succeed when the target table doesn't exist, rather than throwing an error.

Implementation:
- Add `ifExists?: boolean` property to AlterTableNode
- Add `ifExists()` method to AlterTableBuilder
- Override visitAlterTable in PostgresQueryCompiler to emit IF EXISTS

Usage:
```ts
db.schema.alterTable('test')
  .ifExists()
  .addColumn('col', 'text')
  .execute()
```

Generates: `ALTER TABLE IF EXISTS "test" ADD COLUMN "col" text`

Closes #1272

### DIFF
--- a/src/dialect/postgres/postgres-query-compiler.ts
+++ b/src/dialect/postgres/postgres-query-compiler.ts
@@ -1,3 +1,4 @@
+import type { AlterTableNode } from '../../operation-node/alter-table-node.js'
 import { DefaultQueryCompiler } from '../../query-compiler/default-query-compiler.js'
 
 const ID_WRAP_REGEX = /"/g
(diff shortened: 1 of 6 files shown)


igalklebanov:
https://github.com/kysely-org/kysely/blob/master/CONTRIBUTING.md.

PR #1715, opened 2026-02-23
fix: use file specifiers for absolute paths in file-migration-provider

crowlKats:
This is necessary for Deno + JSR, as because an absolute path inside a JSR package in Deno refers to the root of the JSR server, not of the system being run on. Specifying that its a file path ensures that this works as expected in Deno.


crowlKats:
hmm i am surprised tests in node are failing; it does support file specifiers in `import` 🤔

igalklebanov:
Hey 👋 

You can pass a custom `import` function now, giving more control, and making this PR redundant.

PR #1721, opened 2026-02-25
docs: document immediate value behavior in case().then() and case().else()

### LINKED ISSUE #1593: possibly remove `isSafeImmediate` and `parseSafeImmediateValue` from `eb.case()`.
Hey 👋 

Value-first methods, in most cases, append to parameters - unless you use `eb.lit` or `sql.lit`.
`eb.lit('string')` is not allowed.

For some undocumented reason, `eb.case()`'s `then` and `else` values are injected into the query string for numbers, booleans and nulls. This is surprise behavior!

Investigate why that was chosen as a strategy.
Remove `isSafeImmediateValue` and `parseSafeImmediateValue` if there's no good reason.

mixelburg:
Closes #1593

The `then` and `else` methods on `CaseBuilder` have a subtlety that isn't documented anywhere: unlike most value-first methods in Kysely (which append values as query parameters), they inject `number`, `boolean`, and `null` directly into the SQL string as immediate literals.

This is surprising because the rest of the API consistently parameterizes values — only `eb.lit()` / `sql` are supposed to produce inline literals. The behavior exists because `eb.lit('string')` is not allowed inside `then`/`else`, so safe primitives are embedded inline as a workaround.

**Changes:**
- Added a **"Immediate values"** section to the JSDoc for `CaseThenBuilder.then` explaining the inline-embedding behavior and noting that strings are still parameterized
- Added the same section to `CaseWhenBuilder.else`
- Added JSDoc to `isSafeImmediateValue` and `parseSafeImmediateValue` in `value-parser.ts` explaining their purpose and why strings are excluded

### DIFF
--- a/src/parser/value-parser.ts
+++ b/src/parser/value-parser.ts
@@ -56,12 +56,29 @@ export function parseValueExpression(
   return ValueNode.create(exp)
 }
 
+/**
+ * Returns true for primitive values that are safe to embed directly in the SQL
+ * string as immediate literals: numbers, booleans and null.
+ *
+ * Strings are intentionally excluded — they must always be parameterized to
+ * prevent SQL injection. Use `eb.lit()` or `sql` to embed strings inline.
+ *
+ * This is used by {@link CaseThenBuilder.then} and {@link CaseWhenBuilder.else}
+ * to decide whether a value should be an immediate literal or a query parameter.
+ */
 export function isSafeImmediateValue(
   value: unknown,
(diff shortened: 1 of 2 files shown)


koskimas:
The real reason is that dialects like postgres are not able to infer the value type from parameters in `case` expressions, leading to runtime errors. Everything is assumed to be `TEXT`. Postgres is obviously able to infer the type of literals (well enough to work).

PR #1733, opened 2026-03-08
Add datetime2 data type support for MSSQL

### LINKED ISSUE #1596: can't store datetime in full precision milliseconds in mssql
## issue
MSSQL datetime only stores time to an accuracy of ~3.33 milliseconds. This means when you insert a JS date into a datetime column, SQL Server rounds the fractional seconds to the nearest 0, 3, or 7 ms increment.

Example:
JS: 2025-09-26T01:00:00.001Z
MSSQL stored: 2025-09-26T01:00:00.003Z

## solution
In order to resolve this, we must create the datetime field as `datetime2` - however as of now using `db.schema.alterTable(...).addField("name", "datetime2", ...)` won't work.

For more context, here's a post on stackoverflow talking about this specific issue regarding mssql:
https://stackoverflow.com/questions/7823966/milliseconds-in-my-datetime-changes-when-stored-in-sql-server

Additional info:
I require the `.addField` specifically as this is being built into a wrapper function used in the [better-auth](https://github.com/better-auth/better-auth/blob/canary/packages/better-auth/src/db/get-migration.ts) repo.

Br1an67:
Fixes #1596

## Summary
Adds `datetime2` as a recognized column data type for MSSQL. MSSQL's `datetime` type only stores time to ~3.33ms accuracy, causing rounding issues when storing JavaScript dates. The `datetime2` type provides better precision (100ns increments).

## Changes
- Added `datetime2` to `SIMPLE_COLUMN_DATA_TYPES` array
- Added `datetime2(\d+)` regex pattern to `COLUMN_DATA_TYPE_REGEX` for parameterized precision
- Added `datetime2` and `datetime2(${number})` to `ColumnDataType` type union
- Updated test to use `'datetime2'` instead of `sql\`datetime2\``

## Testing
- TypeScript compiles without errors
- `isColumnDataType('datetime2')` returns `true`
- `isColumnDataType('datetime2(7)')` returns `true`

```bash
git diff --stat HEAD~1
 src/operation-node/data-type-node.ts | 3 +++
 test/node/src/schema.test.ts         | 2 +-
 2 files changed, 4 insertions(+), 1 deletion(-)
```

### DIFF
--- a/src/operation-node/data-type-node.ts
+++ b/src/operation-node/data-type-node.ts
@@ -22,6 +22,7 @@ export type SimpleColumnDataType =
   | 'bytea'
   | 'date'
   | 'datetime'
+  | 'datetime2'
   | 'time'
   | 'timetz'
   | 'timestamp'
@@ -59,6 +60,7 @@ const SIMPLE_COLUMN_DATA_TYPES: Readonly<Record<SimpleColumnDataType, true>> =
     datemultirange: true,
     daterange: true,
     datetime: true,
+    datetime2: true,
     decimal: true,
     'double precision': true,
     float4: true,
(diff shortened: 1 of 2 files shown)


igalklebanov (approved):
LGTM! 🚀

igalklebanov (commented):
The email which the PR would have been merged with is clearly a bot. Cannot accept this.

PR #1736, opened 2026-03-09
fix(plugins): make CamelCasePlugin logic extensible

### LINKED ISSUE #1719: Make CamelCasePlugin logic extensible
`CamelCasePlugin` mapping logic (`canMap`) lives outside of the class, https://github.com/kysely-org/kysely/blob/master/src/plugin/camel-case/camel-case-plugin.ts#L176 thus not extensible.

- while one can override `mapRow` or of course just write their own plugin, most extension use-cases will be to modify the mapping logic that lives in `canMap`
- also canMap does not have access to key param for potential extension

I'd propose to move mapping logic within the class, so the plugin class can be extended and built upon and give it the access to key param, which can be used for logic evaluation. If this is something acceptable for improvement, happy to write small PR.

lukashroch:
- make CamelCasePlugin logic extensible by moving the canMap logic into the class
- give canMap function access to key property to be used for canMap extension

This allows plugin class to be extended by standard oop means, e.g. adding new construct options and overwriting with own canMap implemenation

 closes #1719

### DIFF
--- a/src/plugin/camel-case/camel-case-plugin.ts
+++ b/src/plugin/camel-case/camel-case-plugin.ts
@@ -154,8 +154,8 @@ export class CamelCasePlugin implements KyselyPlugin {
       let value = row[key]
 
       if (Array.isArray(value)) {
-        value = value.map((it) => (canMap(it, this.opt) ? this.mapRow(it) : it))
-      } else if (canMap(value, this.opt)) {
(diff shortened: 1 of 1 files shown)


lukashroch:
Is there a reason prevent the mapping logic not be extensible?

WMcKibbin:
@igalklebanov Is there any chance this can be revisited or implemented another way to allow for this?

PR #1737, opened 2026-03-09
docs: warn against running transactions in parallel

### LINKED ISSUE #1711: Mention in documentation for running transaction in parallel is not possible
It's not possible to run transaction in a `Promise.all` for example.

We should document this.

mixelburg:
## Summary

Running multiple Kysely transactions concurrently with `Promise.all` or similar constructs is not supported and leads to undefined behavior, but this wasn't documented anywhere. This PR adds clear documentation about this limitation.

### Site documentation
Added a `:::warning` admonition to the Simple Transaction docs page explaining why parallel transactions are not supported, with a clear ❌/✅ code example.

### JSDoc (inline API docs)
Added a "Parallel transactions" section to the `transaction()` method's JSDoc comment so the warning surfaces in IDE hover documentation for users who don't visit the docs site.

## Why parallel transactions don't work

A transaction occupies a single dedicated database connection and requires all its operations to run sequentially on that connection. When you start multiple transactions concurrently (e.g. via `Promise.all`), their queries interleave on the same connection, causing the BEGIN/COMMIT/ROLLBACK statements and data operations to mix — resulting in data corruption or errors.

## Related

Closes #1711

### DIFF
--- a/site/docs/examples/transactions/0010-simple-transaction.mdx
+++ b/site/docs/examples/transactions/0010-simple-transaction.mdx
@@ -11,6 +11,27 @@ the callback passed to the `execute` method,
 3. the exception is thrown again.
 Otherwise the transaction is committed.
 
+:::warning[Parallel transactions not supported]
+Running multiple transactions in parallel (e.g. with `Promise.all`) is not
+supported. A transaction uses a single dedicated database connection and all
+operations within it must run sequentially on that connection. Attempting to
+run two transactions concurrently — or to run queries inside a transaction
+and in the main connection at the same time — will cause them to interleave
+on the connection, leading to undefined behavior or errors.
+
+```ts
+// ❌ Don't do this — transactions cannot be parallelized
+await Promise.all([
+  db.transaction().execute(async (trx) => { /* ... */ }),
+  db.transaction().execute(async (trx) => { /* ... */ }),
+])
+
+// ✅ Run transactions sequentially instead
+await db.transaction().execute(async (trx) => { /* ... */ })
+await db.transaction().execute(async (trx) => { /* ... */ })
+```
+:::
+
 import { Playground } from '../../../src/components/Playground'
 
 import {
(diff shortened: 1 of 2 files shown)


dallenbaldwin:
I found [this](https://github.com/kysely-org/kysely/issues/1711#issuecomment-3925024781) comment before finding this PR while troubleshooting `Promise.all` usage in a transaction in a project. 

The comment mentions using `Promise.all` _inside_ a transaction is not supported, while a Promise.all _of_ transactions is. It looks like the changes to the docs in this PR doesn't reflect this or the initial comment wasn't correct.

In the spirit of clarity, it might be wise to update this PR to explicitly cover both places someone might try to use `Promise.all`.

```ts
const [a, b, c] = await Promise.all([
 db.transaction.execute(aTransaction),
 db.transaction.execute(bTransaction),
 db.transaction.execute(cTransaction),
])
// do something with a, b, c
```

```ts
await db.transaction.execute(async (t) => {
 const [a, b, c] = await Promise.all([
  t.insertInto(table).values(aValues).returningAll().executeTakeFirstOrThrow(),
  t.insertInto(table).values(bValues).returningAll().executeTakeFirstOrThrow(),
  t.insertInto(table).values(cValues).returningAll().executeTakeFirstOrThrow(),
 ])
 // do something in transaction with values of a, b, c
})
```

It might be worth noting, in the project referenced above, we did get `Promise.all` to work in the transaction at first (second example), but when an error came back in one of the promises, it crashed the node process instead of being caught by a try/catch wrapped around the transaction. We suspect, after seeing the above comment, we might have been getting lucky during initial sanity checks. 

I do recognize, after writing this up, how silly it might be to use `Promise.all` to get `a`, `b`, and `c` in the transaction when you could just write each inline, but in the moment that wasn't on our minds.

igalklebanov:
@dallenbaldwin it's a spam AI pull request. Ignore.

Issue #1707, opened 2026-02-14
Postgres introspector is too slow for databases with many schemas and tables


rubenferreira97:
When using Kysely with a Postgres database containing many schemas and tables, `migrateToLatest()` takes **~500 seconds** to start up, even when **no migrations need to be applied**.

My production database has **~40 schemas with ~200 tables each** (~8,000 tables total). Just calling `migrateToLatest()` blocks the application for minutes.

## Root cause

After profiling, I found two main bottlenecks:

### 1. `#parseTableMetadata` is O(n²)

`PostgresIntrospector.#parseTableMetadata` uses `Array.find()` to look up each table by name and schema for every column row:

```typescript
return columns.reduce<TableMetadata[]>((tables, it) => {
  let table = tables.find(
    (tbl) => tbl.name === it.table && tbl.schema === it.schema,
  )
  // ...
}, [])
```

For each of the *n* columns, it linearly scans the growing `tables` array, resulting in **O(n²) time complexity**. With ~50,000 columns, this in-memory step alone takes **~183 ms** per call.

### 2. `#doesTableExist` fetches all tables in the database

The migrator's `#doesTableExist` calls `getTables({ withInternalKyselyTables: true })` which introspects **every table in every schema**, then filters in JavaScript:

```typescript
const tables = await this.#props.db.introspection.getTables({
  withInternalKyselyTables: true,
})

return tables.some(
  (it) => it.name === tableName && (!schema || it.schema === schema),
)
```

With 8,000 tables, this fetches tens of thousands of column rows from `pg_catalog` **twice** (once for the migration table, once for the lock table), just to check if two specific tables exist. This is the dominant contributor to the ~500s startup.

## Benchmark: `#parseTableMetadata` (in-memory only)

| Scenario | Total Columns | Before (ms) | After (ms) | Speedup |
|---|---|---|---|---|
| Small (1 schema, 5 tables, 5 cols) | 25 | 0.002 | 0.002 | 1.1x |
| Medium (2 schemas, 20 tables, 10 cols) | 400 | 0.051 | 0.026 | **2.0x** |
| Large (3 schemas, 50 tables, 15 cols) | 2,250 | 0.731 | 0.133 | **5.5x** |
| XL (5 schemas, 100 tables, 20 cols) | 10,000 | 11.46 | 0.61 | **18.9x** |
| XXL (10 schemas, 200 tables, 25 cols) | 50,000 | 182.69 | 2.85 | **64x** |

[...]

PR #1776, opened 2026-04-02
refactor: optimize table metadata parsing in PostgresIntrospector

rubenferreira97:
**Motivation / Context**
This PR is a follow-up optimization to the O(n) improvements made in #1774 for the Postgres introspector bottleneck reported in #1707. 

While the previous implementation successfully reduced the time complexity to O(n) using a single `Map`, profiling showed that allocating composite string keys (`` `schema:${schema};table:${table}` ``) on every iteration created unnecessary overhead for databases with tens of thousands of columns. Furthermore, calling `Array.from(tableDictionary.values())` at the end added an extra iteration step.

**Changes**
* Replaced the single string-keyed `Map` with a nested structure: `Map<string, Map<string, TableMetadata>>` (Schema -> Table -> Metadata).
* Eliminated the composite string key allocations inside the hot loop.
* Introduced a flat `tables` array to push newly created references directly, removing the need for `Array.from()` at the end of the execution.

**Performance Impact**
I've manually benchmarked these changes locally to confirm the performance gains. To keep the PR clean and avoid polluting the repository with one-off testing files, I haven't included the benchmark scripts in the code changes.

This approach maintains the O(n) lookup complexity but significantly lowers the constant factor overhead by avoiding repetitive string instantiation. Based on previous benchmarks for large schemas (~50,000+ columns), this nested map approach benchmarks roughly ~5x faster (in node) than the string-key lookup method.

### DIFF
--- a/src/dialect/postgres/postgres-introspector.ts
+++ b/src/dialect/postgres/postgres-introspector.ts
@@ -105,28 +105,31 @@ export class PostgresIntrospector implements DatabaseIntrospector {
   }
 
   #parseTableMetadata(columns: RawColumnMetadata[]): TableMetadata[] {
-    const tableDictionary: Map<string, TableMetadata> = new Map()
+    const tables: TableMetadata[] = []
+    const schemas = new Map<string, Map<string, TableMetadata>>()
 
-    for (let i = 0, len = columns.length; i < len; i++) {
-      const column = columns[i]
+    for (const column of columns) {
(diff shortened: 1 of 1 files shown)


igalklebanov:
Merged in https://github.com/kysely-org/kysely/pull/1801/changes/36c6ceaa7b38b1556064f1ccac371ef7a9a11689.

PR #1793, opened 2026-04-11
docs(returning): remove outdated SQLite alias workaround

### LINKED ISSUE #1563: Docs for `returning` are inaccurate
The docs for `returning` mention a bug in SQLite that got patched over 4 years ago and was only present in a single release of SQLite.

>      * Note that on SQLite you need to give aliases for the expressions to avoid
>      * [this bug](https://sqlite.org/forum/forumpost/033daf0b32) in SQLite.
>      * For example `.returning('id as id')`.

the bug got fixed in this commit: https://sqlite.org/src/timeline?c=699c33990a9438f2&y=a and the fix was included in the 3.36.0 release

aymenhmaidiwastaken:
## Summary
- Removed the outdated SQLite bug warning from the `returning` JSDoc
- The referenced bug ([sqlite.org/forum/forumpost/033daf0b32](https://sqlite.org/forum/forumpost/033daf0b32)) was fixed in SQLite 3.36.0 (June 2021) via [this commit](https://sqlite.org/src/timeline?c=699c33990a9438f2&y=a)
- The `.returning('id as id')` workaround hasn't been necessary for over 4 years

Closes #1563

### DIFF
--- a/site/docusaurus.config.ts
+++ b/site/docusaurus.config.ts
@@ -47,11 +47,6 @@ export default {
           ],
           enableLlmsFullTxt: true,
           includeGeneratedIndex: false,
-          // Explicitly exclude auto-generated category index pages.
-          // These only list child pages and have no meaningful content for LLMs.
-          excludeRoutes: [
-            '/docs/category/**',
-          ],
           includePages: true,
           includeVersionedDocs: false,
           relativePaths: false,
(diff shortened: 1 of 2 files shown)


site/docusaurus.config.ts
```diff
@@ -47,6 +45,11 @@ export default {
           ],
           enableLlmsFullTxt: true,
           includeGeneratedIndex: false,
+          // Explicitly exclude auto-generated category index pages.
+          // These only list child pages and have no meaningful content for LLMs.
+          excludeRoutes: [
+            '/docs/category/**',
+          ],
```
igalklebanov:
```suggestion
```

igalklebanov (approved):
LGTM! 🚀

PR #1805, opened 2026-05-03
docs: add kysely-durable-objects to community dialects

jeffwilde:
Adds [`kysely-durable-objects`](https://github.com/jeffwilde/kysely-durable-objects) to the community dialects table — a Kysely dialect for Cloudflare Durable Object SQLite storage (`ctx.storage.sql`).

I left the existing `benallfree/kysely-do` row in place rather than replacing it, so both packages targeting this platform are listed. Happy to adjust label / wording if you'd prefer a different convention for two entries pointing at the same platform.

`kysely-durable-objects` is published to npm with provenance, MIT-licensed, and tested inside real `workerd` (via `@cloudflare/vitest-pool-workers`) across a `compatibility_date` matrix from DO-SQLite GA to today.


igalklebanov (approved):
Hey 👋 

Thanks! 🙏

Issue #596, opened 2023-07-18
when i insert empty list i get an error 


JIM-GLITCH:
```ts
import { pg } from "./connectDB";
pg.insertInto("data.advice")
        .values([])
        .execute()
```
when i insert empty list i get an error 
![image](https://github.com/kysely-org/kysely/assets/58733364/71693f6d-ac0c-4d8f-8f75-a3bdde629c66)

JIM-GLITCH:
it generates invalid sql sentence

wirekang:
What you are requesting to database is "insert nothing" via "INSERT" keyword. Check the length of input to prevent kysely from generating invalid SQL.

```ts
if (input.length === 0){
  // return or throw exception
}
pg.insertInto().values(input)

```

koskimas:
Yep, everything works as expected here.

Issue #1630, opened 2025-11-08
Invalid SQL when `insertInto.values` array is empty


GTonehour:
~~~ts
import { PostgresDialect } from "kysely";

const arr: Array<Row> = [];

db
.insertInto("table")
.values(arr);
~~~

Kysely@0.27.6 generates `"insert into \"table\" () values "` which PG fails to execute.

Instead of testing `arr.length > 0`, could Kysely generate no SQL in this case, or any SQL that PG wouldn't fail to execute?

PR #1824, opened 2026-05-11
feat: add HandleEmptyInsertValuesPlugin

Magnuti:
Adds a new opt-in plugin, `HandleEmptyInsertValuesPlugin`, for handling empty array inserts.

```ts
insertInto("t").values([]) // Invalid SQL causing a runtime error
```

This produces invalid SQL in most databases, causing a runtime error. This can happen if we forget to length check a list before insert.

When triggered, the plugin rewrites the query to `INSERT INTO t SELECT * FROM t WHERE 1 = 0`, which is valid SQL that inserts zero rows and returns `numInsertedOrUpdatedRows: 0n`.

```ts
plugins: [
    new HandleEmptyInsertValuesPlugin()
],

...

insertInto("t").values([]) // Valid SQL, no-op
```

Relevant to the discussion here https://github.com/kysely-org/kysely/pull/1473#issuecomment-2925158012 on PR #1473.

Also relevant to these closed issues #596, #1401, #1630.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -127,6 +127,7 @@ export * from './plugin/with-schema/with-schema-plugin.js'
 export * from './plugin/parse-json-results/parse-json-results-plugin.js'
 export * from './plugin/handle-empty-in-lists/handle-empty-in-lists-plugin.js'
 export * from './plugin/handle-empty-in-lists/handle-empty-in-lists.js'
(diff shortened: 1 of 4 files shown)


Magnuti:
Hi @igalklebanov, any chance you could share why this was closed?

Magnuti:
Could it also be possibility that we alter the `.execute()` statement so it does not execute instead of calling the database with `1 = 0` ?

PR #1838, opened 2026-05-14
fix: plugin chaining

bojanrajh:
This PR fixes a regression in result plugin chaning from https://github.com/kysely-org/kysely/pull/1798/changes#diff-c64dda439503d27dc1cbeaa4813abf0eb82cfd770654ec94fd9a5b679c1727f9R306

Each plugin now receives fresh args containing the result returned by the previous plugin, as it was done before.


bojanrajh:
@igalklebanov Any plans on fixing this? Seems like we can work it around by putting the `CamelCasePlugin` at the end of the `plugins` list, but this fix won't work for projects using multiple plugins with `transformResult()`.

bojanrajh:
Fixed in https://github.com/kysely-org/kysely/pull/1840

PR #1868, opened 2026-05-21
Fix CTE syntax in code sample

aenario:
The code sample is broken on https://kysely.dev/docs/examples/cte/simple-selects


igalklebanov:
Hey 👋 

The example is correct. The playground was missing 0.29.x versions. This is now resolved.

PR #1910, opened 2026-06-10
Some queries return multiple results

DavidTanner:
# Summary

This change updates the PostgreSQL query handling logic to correctly support the return types provided by the Node PostgreSQL (`pg`) library.

## Background

The existing implementation assumes that query execution always returns a single `QueryResult` object. While this is true for standard queries, the `pg` library can also return an array of `QueryResult` objects in scenarios where multiple statements are executed and multiple result sets are returned.

Because the code only handled a single result, it could incorrectly process or ignore valid responses returned by PostgreSQL.

## Changes

* Updated query result handling to support both:

  * `QueryResult`
  * `QueryResult[]`
* Added type checks and processing logic to correctly handle multiple result sets.
* Preserved existing behavior for queries that return a single result.
* Improved type safety and compatibility with the PostgreSQL client API.

## Testing

* Verified existing single-result queries continue to function as expected.
* Added validation for queries returning multiple result sets.
* Confirmed result processing behaves correctly for both return types.

## Impact

This change is backward compatible and improves support for PostgreSQL query execution scenarios that return multiple result sets, preventing potential runtime errors and ensuring all returned data can be processed correctly.

### DIFF
--- a/src/dialect/postgres/postgres-dialect-config.ts
+++ b/src/dialect/postgres/postgres-dialect-config.ts
@@ -93,7 +93,7 @@ export interface PostgresClient {
   query<R>(
     sql: string,
     parameters: ReadonlyArray<unknown>,
-  ): Promise<PostgresQueryResult<R>>
+  ): Promise<PostgresQueryResult<R> | PostgresQueryResult<R>[]>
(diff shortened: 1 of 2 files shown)


DavidTanner:
@igalklebanov why is the MR closed?  Can I get a comment?

PR #1919, opened 2026-06-27
fix: PostgreSQL and MSSQL migrations are not running exclusively when `disableTransactions: true`.

### LINKED ISSUE #1894: With `disableTransactions: true`, no locking mechanism prevents migrators from running concurrently
In my testing, if I run two migrators simultaneously with `disableTransactions: true`, both migrators will execute simultaneously. One migration will succeed and continue, while the other will execute the first migration file, but then fail when it attempts to insert a duplicate record into the `migrationTable`. This is likely to be problematic since we are running outside of a transaction.

---

In Postgres, the `pg_advisory_xact` will not hold a lock outside of a transaction.

---

I was expecting `migrationLockTable` to help here but it appears to be a vestige that doesn't actually do much of anything.

https://github.com/kysely-org/kysely/blob/e17673f3e8cfb5723de15c7f532917d9f7f15a7d/src/migration/migrator.ts#L451-L471

`#ensureLockRowExists()` - I suspected this function to throw an error if the lock record already exists to prevent a concurrent run, but it doesn't.

There also isn't anything that reads or writes to the `is_locked` column, which could also be used to prevent concurrent runs.

---

To workaround this, I have a Postgres specific solution:

<details><summary>workaround.ts</summary>

```ts
/**
 * Use an advisory lock to prevent migrations race
 *
 * Two concurrent calls to `kyselyMigrate` can lead to a race condition. Kysely's built-in advisory
 * lock doesn't help us because it doesn't run in a transaction.
 *
 * Our mechanism works by acquiring an advisory lock in a transaction and holding that transaction
 * open while the callback is called outside of the transaction. This isn't perfect. It's possible
 * the transaction could get killed due to a network error before the callback completes. However,
 * it does add significant protection.
 */
async function withMigrationLock<T>(db: Kysely<any>, lockKey: bigint | undefined, callback: () => Promise<T>): Promise<T> {
  // Random key for our transaction lock. (It's the same one Kysely uses plus 1)
  const DEFAULT_LOCK_KEY = BigInt("3853314791062309108");

  const lockHasStarted = Promise.withResolvers<void>();
  const lockHasEnded = Promise.withResolvers<void>();
  const migrationsCompleted = Promise.withResolvers<void>();
  void Promise.try(async () => {
    try {
      await db.transaction().execute(async (trx) => {
        await sql`select pg_advisory_xact_lock(${sql.lit(lockKey ?? DEFAULT_LOCK_KEY)})`.execute(trx);
        lockHasStarted.resolve();
        await migrationsCompleted.promise;
      });
    } catch (error) {
      lockHasStarted.reject(error);
      lockHasEnded.reject(error);
      throw error;
    }
    lockHasEnded.resolve();
  }).catch((error) => {
    lockHasStarted.reject(error);
    lockHasEnded.reject(error);
  });

  await lockHasStarted.promise;
  try {
    return await callback();
  } finally {
    migrationsCompleted.resolve();
    await lockHasEnded.promise;
  }
}
```

</details>

---

Anyway, this should be addressed by Kysely. What is the intended purpose of the lock table and what are the plans for it? Should the lock table be fixed or should it be removed?

morgan-coded:
Postgres migration locking was using a transaction-level advisory lock, which can be released too early when migrations are not protected by a surrounding transaction.

This switches the Postgres adapter to session-level advisory locking with `pg_try_advisory_lock`, the same one-hour timeout shape used by MySQL, and `pg_advisory_unlock` on release. For Kysely-owned transactional migrations, the migrator now keeps the migration body in a transaction but releases the lock from the same outer connection after the transaction settles, so the lock covers the whole migration run.

I added a real Postgres concurrent-migrator regression for #1894. It fails without the fix and passes with each migration applied once; I also ran build, node test build, the focused Postgres migration suite, ESM import tests, and export tests.

Fixes #1894

### DIFF
--- a/src/dialect/dialect-adapter.ts
+++ b/src/dialect/dialect-adapter.ts
@@ -103,10 +103,12 @@ export interface DialectAdapter {
    * }
    * ```
    *
-   * If `supportsTransactionalDdl` is `true` then the `db` passed to this method
-   * is a transaction inside which the migrations will be executed. Otherwise
-   * `db` is a single connection (session) that will be used to execute the
-   * migrations.
+   * If `supportsTransactionalDdl` is `true` and Kysely starts the migration
+   * transaction, the `db` passed to this method is the transaction inside which
+   * the migrations will be executed. The matching `releaseMigrationLock` call
+   * receives the same connection (session) after the transaction has settled.
+   * Otherwise `db` is a single connection (session) that will be used to
+   * execute the migrations.
    */
   acquireMigrationLock(
     db: Kysely<any>,
@@ -116,10 +118,11 @@ export interface DialectAdapter {
   /**
    * Releases the migration lock. See {@link acquireMigrationLock}.
    *
-   * If `supportsTransactionalDdl` is `true` then the `db` passed to this method
-   * is a transaction inside which the migrations were executed. Otherwise `db`
-   * is a single connection (session) that was used to execute the migrations
-   * and the `acquireMigrationLock` call.
+   * If `supportsTransactionalDdl` is `true` and Kysely starts the migration
+   * transaction, the `db` passed to this method is the same connection (session)
+   * after the transaction has settled. Otherwise `db` is the transaction or
+   * single connection that was used to execute the migrations and the
+   * `acquireMigrationLock` call.
    */
   releaseMigrationLock(
     db: Kysely<any>,
--- a/src/dialect/dialect-adapter.ts
+++ b/src/dialect/dialect-adapter.ts
@@ -103,12 +103,10 @@ export interface DialectAdapter {
    * }
    * ```
    *
-   * If `supportsTransactionalDdl` is `true` and Kysely starts the migration
-   * transaction, the `db` passed to this method is the transaction inside which
-   * the migrations will be executed. The matching `releaseMigrationLock` call
-   * receives the same connection (session) after the transaction has settled.
-   * Otherwise `db` is a single connection (session) that will be used to
-   * execute the migrations.
+   * If `supportsTransactionalDdl` is `true` then the `db` passed to this method
+   * is a transaction inside which the migrations will be executed. Otherwise
+   * `db` is a single connection (session) that will be used to execute the
+   * migrations.
    */
   acquireMigrationLock(
     db: Kysely<any>,
@@ -118,11 +116,10 @@ export interface DialectAdapter {
   /**
    * Releases the migration lock. See {@link acquireMigrationLock}.
    *
-   * If `supportsTransactionalDdl` is `true` and Kysely starts the migration
-   * transaction, the `db` passed to this method is the same connection (session)
-   * after the transaction has settled. Otherwise `db` is the transaction or
-   * single connection that was used to execute the migrations and the
-   * `acquireMigrationLock` call.
+   * If `supportsTransactionalDdl` is `true` then the `db` passed to this method
+   * is a transaction inside which the migrations were executed. Otherwise `db`
+   * is a single connection (session) that was used to execute the migrations
+   * and the `acquireMigrationLock` call.
    */
   releaseMigrationLock(
     db: Kysely<any>,
--- a/src/dialect/postgres/postgres-adapter.ts
+++ b/src/dialect/postgres/postgres-adapter.ts
@@ -3,8 +3,8 @@ import { sql } from '../../raw-builder/sql.js'
 import { DialectAdapterBase } from '../dialect-adapter-base.js'
 import type { MigrationLockOptions } from '../dialect-adapter.js'
 
-// Random id for our migration lock.
 const LOCK_ID = BigInt('3853314791062309107')
+const LOCK_TIMEOUT_MILLISECONDS = 60 * 60 * 1_000
 
 export class PostgresAdapter extends DialectAdapterBase {
   override get supportsTransactionalDdl(): boolean {
@@ -19,9 +19,16 @@ export class PostgresAdapter extends DialectAdapterBase {
     db: Kysely<any>,
     _opt: MigrationLockOptions,
   ): Promise<void> {
-    // Non-transactional migrations are autocommitted statement by statement,
-    // so use a session-level lock held until releaseMigrationLock.
-    await sql`select pg_advisory_lock(${sql.lit(LOCK_ID)})`.execute(db)
+    // in 1 RTT, acquire a session-level advisory lock with a timeout.
+    //
+    // if ever this runs inside a transaction, niche edge case we support - the
+    // user is responsible to set a different timeout value or disable (`0` value).
+    await sql`
+    with set_timeout as (
+      select set_config('lock_timeout', '${sql.lit(LOCK_TIMEOUT_MILLISECONDS)}', true) as config_val
+    )
+    select pg_advisory_lock(${sql.lit(LOCK_ID)})
+    from set_timeout`.execute(db)
   }
 
   override async releaseMigrationLock(
--- a/src/dialect/postgres/postgres-adapter.ts
+++ b/src/dialect/postgres/postgres-adapter.ts
@@ -3,8 +3,12 @@ import { sql } from '../../raw-builder/sql.js'
 import { DialectAdapterBase } from '../dialect-adapter-base.js'
 import type { MigrationLockOptions } from '../dialect-adapter.js'
 
-// Random id for our transaction lock.
+// Random id for our migration lock.
 const LOCK_ID = BigInt('3853314791062309107')
+// Mirrors the MySQL adapter's `get_lock` timeout so a migrator doesn't wait
+// forever for a competing migrator to release the lock.
+const LOCK_TIMEOUT_SECONDS = 60 * 60
+const LOCK_POLL_INTERVAL_MS = 100
 
 export class PostgresAdapter extends DialectAdapterBase {
   override get supportsTransactionalDdl(): boolean {
@@ -19,16 +23,38 @@ export class PostgresAdapter extends DialectAdapterBase {
     db: Kysely<any>,
     _opt: MigrationLockOptions,
   ): Promise<void> {
-    // Acquire a transaction level advisory lock.
-    await sql`select pg_advisory_xact_lock(${sql.lit(LOCK_ID)})`.execute(db)
+    // Kysely uses a single connection to run the migrations. Because of that,
+    // we can take a session level advisory lock and release it using
+    // `releaseMigrationLock`. We poll `pg_try_advisory_lock` instead of
+    // blocking on `pg_advisory_lock` so that a migrator doesn't wait forever for
+    // a competing migrator - it fails once the timeout is exceeded.
+    const startTime = Date.now()
+
+    while (true) {
+      const result = await sql<{
+        acquired: boolean
+      }>`select pg_try_advisory_lock(${sql.lit(LOCK_ID)}) as acquired`.execute(
+        db,
+      )
+
+      if (result.rows[0]?.acquired) {
+        return
+      }
+
+      if (Date.now() - startTime >= LOCK_TIMEOUT_SECONDS * 1000) {
+        throw new Error(
+          `Timed out after ${LOCK_TIMEOUT_SECONDS} seconds while waiting for the migration lock.`,
+        )
+      }
+
+      await new Promise((resolve) => setTimeout(resolve, LOCK_POLL_INTERVAL_MS))
+    }
   }
 
   override async releaseMigrationLock(
-    _db: Kysely<any>,
+    db: Kysely<any>,
     _opt: MigrationLockOptions,
   ): Promise<void> {
-    // Nothing to do here. `pg_advisory_xact_lock` is automatically released at the
-    // end of the transaction and since `supportsTransactionalDdl` true, we know
-    // the `db` instance passed to acquireMigrationLock is actually a transaction.
+    await sql`select pg_advisory_unlock(${sql.lit(LOCK_ID)})`.execute(db)
   }
 }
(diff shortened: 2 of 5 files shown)


src/dialect/dialect-adapter.ts
```diff
@@ -116,10 +118,11 @@ export interface DialectAdapter {
...
    *
-   * If `supportsTransactionalDdl` is `true` then the `db` passed to this method
-   * is a transaction inside which the migrations were executed. Otherwise `db`
-   * is a single connection (session) that was used to execute the migrations
-   * and the `acquireMigrationLock` call.
+   * If `supportsTransactionalDdl` is `true` and Kysely starts the migration
+   * transaction, the `db` passed to this method is the same connection (session)
+   * after the transaction has settled. Otherwise `db` is the transaction or
+   * single connection that was used to execute the migrations and the
+   * `acquireMigrationLock` call.
```
igalklebanov: This interface is only relevant for dialect builders. It should not talk about implementation detail of migrator/s. The original description is all that needs to be here for someone writing a dialect to succeed.

```suggestion
```

src/dialect/dialect-adapter.ts
```diff
@@ -103,10 +103,12 @@ export interface DialectAdapter {
...
-   * If `supportsTransactionalDdl` is `true` then the `db` passed to this method
-   * is a transaction inside which the migrations will be executed. Otherwise
-   * `db` is a single connection (session) that will be used to execute the
-   * migrations.
+   * If `supportsTransactionalDdl` is `true` and Kysely starts the migration
+   * transaction, the `db` passed to this method is the transaction inside which
+   * the migrations will be executed. The matching `releaseMigrationLock` call
+   * receives the same connection (session) after the transaction has settled.
+   * Otherwise `db` is a single connection (session) that will be used to
+   * execute the migrations.
```
igalklebanov: See https://github.com/kysely-org/kysely/pull/1919/changes#r3501878876


```suggestion
```

src/dialect/postgres/postgres-adapter.ts
```diff
@@ -19,16 +23,38 @@ export class PostgresAdapter extends DialectAdapterBase {
...
+      }
+
+      if (Date.now() - startTime >= LOCK_TIMEOUT_SECONDS * 1000) {
+        throw new Error(
+          `Timed out after ${LOCK_TIMEOUT_SECONDS} seconds while waiting for the migration lock.`,
+        )
+      }
+
+      await new Promise((resolve) => setTimeout(resolve, LOCK_POLL_INTERVAL_MS))
+    }
```
igalklebanov: The implementation here can result in starvation and is too far off the standard path to consider.

src/migration/migrator.ts
```diff
@@ -558,14 +564,23 @@ export class Migrator {
...
-      return run(this.#props.db)
+      return runWithLock(this.#props.db)
     }
 
     if (adapter.supportsTransactionalDdl && !disableTransactions) {
-      return this.#props.db.transaction().execute(run)
+      return this.#props.db.connection().execute(async (db) => {
+        try {
+          return await db.transaction().execute(async (trx) => {
+            await adapter.acquireMigrationLock(trx, lockOptions)
```
igalklebanov: this can and should happen before starting a transaction. you don't want to use excessive resources and then wait in queue for lock.

src/migration/migrator.ts
```diff
@@ -516,27 +516,33 @@ export class Migrator {
...
+        return await this.#migrateUp(db, state, step)
+      }
 
-        return { results: [] }
+      return { results: [] }
+    }
+
+    const runWithLock = async (
+      db: Kysely<any>,
+    ): Promise<MigrationResultSet> => {
```
igalklebanov, line 543: i think this could accept any callback really. and then be also reused in the DDL + !disableTrx block.

```ts
await runWithLock(cb: () => Promise<void>): Promise<void>
```

src/migration/migrator.ts
```diff
@@ -516,27 +516,33 @@ export class Migrator {
...
-        }
+      if (direction === 'Down') {
+        return await this.#migrateDown(db, state, step)
+      } else if (direction === 'Up') {
+        return await this.#migrateUp(db, state, step)
+      }
 
-        return { results: [] }
+      return { results: [] }
+    }
```
igalklebanov, line 538: you got the simpler refactor path here right - extract to make this usable with a session-level in and out of a transaction.. 💪

test/node/src/migration.test.ts
```diff
@@ -1174,13 +1221,17 @@ for (const dialect of DIALECTS) {
...
                 await setTimeout(20)
 
                 if (config.error) {
                   throw new Error(config.error)
                 }
 
+                if (config.transaction !== undefined) {
+                  expect(db.isTransaction).to.equal(config.transaction)
+                }
+
```
igalklebanov: this looks like a very unexpected place to put a test assertion in. it's a factory for test inputs.

test/node/src/migration.test.ts
```diff
@@ -905,6 +905,50 @@ for (const dialect of DIALECTS) {
...
 
+      if (variant === 'postgres') {
+        it('should not run concurrent migrators in parallel when disableTransactions is true (issue #1894)', async () => {
+          // Two migrators running `migrateToLatest` in parallel with
+          // `disableTransactions: true` must not execute migrations
+          // concurrently. Without a surrounding transaction, Postgres releases
+          // `pg_advisory_xact_lock` at the end of the statement that acquires
+          // it, so both migrators can run the same migration's `up` and the
+          // loser fails when it tries to insert a duplicate row into the
+          // migration table.
```
igalklebanov: the test case description already tells the story. this comment is redundant and already drifts from the implementation the test case is testing.

test/node/src/migration.test.ts
```diff
@@ -905,6 +905,50 @@ for (const dialect of DIALECTS) {
...
+            ['migration1', 'migration2'],
+            { disableTransactions: true },
+          )
+
+          const [resultA, resultB] = await Promise.all([
+            migratorA.migrateToLatest(),
+            migratorB.migrateToLatest(),
+          ])
+
+          // Neither migrator may error (no duplicate-key violation).
```
igalklebanov: this emphasizes the wrong thing - a technical error, where? why?

the why is more important. a specific migration that appears in both runs was not ran more than once.

test/node/src/migration.test.ts
```diff
@@ -905,6 +905,50 @@ for (const dialect of DIALECTS) {
...
+          expect(resultA.error).to.be.undefined
+          expect(resultB.error).to.be.undefined
+
+          // Each migration's `up` must run exactly once across both migrators.
+          const allExecutedUp = [...executedUpA, ...executedUpB]
+          expect(allExecutedUp.sort()).to.eql(['migration1', 'migration2'])
```
igalklebanov: a stronger assertion would be - the first run ran both migrations, the second run didn't run any migrations, ran a third migration that doesn't exist in the first run only.

test/node/src/migration.test.ts
```diff
@@ -905,6 +905,50 @@ for (const dialect of DIALECTS) {
...
+          // The migration table must contain exactly one row per migration.
+          const executed = await (ctx.db as Kysely<any>)
+            .selectFrom(DEFAULT_MIGRATION_TABLE)
+            .select('name')
+            .orderBy('name')
+            .execute()
+          expect(executed.map((it: { name: string }) => it.name)).to.eql([
+            'migration1',
+            'migration2',
+          ])
```
igalklebanov: having this side effect assertion (persisted state post-op) means the previous `up` assertions are redundant.

test/node/src/migration.test.ts
```diff
@@ -841,18 +847,18 @@ for (const dialect of DIALECTS) {
...
-        })
+        const shouldExecuteInTransaction =
+          ctx.db.getExecutor().adapter.supportsTransactionalDdl
+
+        const [migrator, executedUpMethods] = createMigrations(
+          [{ name: 'migration1', transaction: shouldExecuteInTransaction }],
+          {
+            disableTransactions: true,
+          },
+        )
```
igalklebanov: this contradicts the test case description - `disableTransactions` should be `false`.

morgan-coded:
Thanks, makes sense. Switched to blocking `pg_advisory_lock`, moved the lock to the outer connection before the transaction, and unified the three paths through one `runWithLock` callback. Also reverted the doc changes and reworked the #1894 test to the asymmetric case.

src/dialect/postgres/postgres-adapter.ts
```diff
@@ -19,16 +19,15 @@ export class PostgresAdapter extends DialectAdapterBase {
     db: Kysely<any>,
     _opt: MigrationLockOptions,
   ): Promise<void> {
-    // Acquire a transaction level advisory lock.
-    await sql`select pg_advisory_xact_lock(${sql.lit(LOCK_ID)})`.execute(db)
+    // Kysely uses a single connection/session to run migrations, so we take a
+    // session-level advisory lock and release it in releaseMigrationLock.
```
igalklebanov: this is not the reason why session-level is taken. it is taken because in some scenarios migrations run not in transactions and we need exclusivity covering autocommits too.

test/node/src/migration.test.ts
```diff
@@ -830,9 +830,16 @@ for (const dialect of DIALECTS) {
       })
 
       it('should execute in transaction if disableTransactions is false on the `Migrator` instance and transactionDdl supported', async () => {
-        const [migrator, executedUpMethods] = createMigrations(['migration1'], {
-          disableTransactions: false,
-        })
+        const shouldExecuteInTransaction =
+          ctx.db.getExecutor().adapter.supportsTransactionalDdl
+
+        const [migrator, executedUpMethods, , executedInTransactions] =
```
igalklebanov: nitpicky

switch to a plain object, this is not as clean + nothing seems to use the 3rd slot which makes this even worse.

test/node/src/migration.test.ts
```diff
@@ -905,6 +909,63 @@ for (const dialect of DIALECTS) {
...
+            migratorA.migrateToLatest(),
+            migratorB.migrateToLatest(),
+          ])
+
+          // A migration that both migrators know about must run exactly once,
+          // not once per migrator.
+          expect(resultA.error).to.be.undefined
+          expect(resultB.error).to.be.undefined
+
+          const allExecutedUp = [...executedUpA, ...executedUpB]
```
igalklebanov: this assumes there's an order to these.

instead you should track the timestamp when something ran and sort by it in ascending order. now you could assert that the array equals deeply: `[{ name: 'migration1', source: 'A' }, { name: 'migration2', source: 'A' }, { name: 'migration3', source: 'B' }]` in a declarative way.

if we were in `vitest` (which we're sadly not using yet) I'd just go with an inline snapshot to make it even more readable AND maintainable.

test/node/src/migration.test.ts
```diff
@@ -905,6 +909,63 @@ for (const dialect of DIALECTS) {
...
+              executedUpA[1] === 'migration2' &&
+              executedUpB.length === 1 &&
+              executedUpB[0] === 'migration3') ||
+              (executedUpB.length === 3 &&
+                executedUpB[0] === 'migration1' &&
+                executedUpB[1] === 'migration2' &&
+                executedUpB[2] === 'migration3' &&
+                executedUpA.length === 0),
+          ).to.equal(true)
+
```
igalklebanov: this is too verbose. see https://github.com/kysely-org/kysely/pull/1919/changes#r3504328934.

test/node/src/migration.test.ts
```diff
@@ -905,6 +909,63 @@ for (const dialect of DIALECTS) {
...
+          const executed = await (ctx.db as Kysely<any>)
+            .selectFrom(DEFAULT_MIGRATION_TABLE)
+            .select('name')
+            .orderBy('name')
+            .execute()
+          expect(executed.map((it: { name: string }) => it.name)).to.eql([
+            'migration1',
+            'migration2',
+            'migration3',
+          ])
```
igalklebanov: you could also drop the assertions earlier due to redundancy and just get a sorted by timestamp list of the migrations from the database and assert it equals deeply to expected rows.

morgan-coded:
Thanks, that makes sense. Fixed the lock comment to call out the non-transactional/autocommit case, switched the test helper to named fields, and collapsed the #1894 assertion to a single timestamp-sorted deep-equal. I kept that on an in-memory run-log rather than a DB query so it can also assert which migrator ran each migration.

morgan-coded:
Thanks, that makes sense. Moved the transaction check out of `createMigrations`; the helper now records `{ name, isTransaction }` for each run, and the transaction tests assert those facts in the test bodies.

igalklebanov:
I have added a lock timeout on the PostgreSQL side - in 1 RTT and only applies to the acquisition - unless a user passed his own transaction to a migrator.

I have made the MS SQL Server adapter also use session-level locks.

Ensure tests still pass and cover both of these dialects.

morgan-coded:
Thanks. CI is green on your `606c454`/`aabc4c4` head for Postgres and MSSQL across Node 22/24/26; the only red status I see is the Vercel deploy auth gate.

The migration suite runs through the `DIALECTS` loop for both dialects, and the dedicated advisory-lock assertion is Postgres-specific around `pg_advisory_xact_lock`. Would you like me to add a matching MSSQL session-lock assertion, or is the matrix coverage enough for this PR?

src/dialect/postgres/postgres-adapter.ts
```diff
@@ -19,16 +19,22 @@ export class PostgresAdapter extends DialectAdapterBase {
...
+    // in 1 RTT, acquire a session-level advisory lock with a timeout.
+    //
+    // if ever this runs inside a transaction, niche edge case we support - the
+    // user is responsible to set a different timeout value or disable (`0` value).
+    await sql`
+    with set_timeout as (
+      select set_config('lock_timeout', '${sql.lit(LOCK_TIMEOUT_MILLISECONDS)}', true) as config_val
+    )
+    select pg_advisory_lock(${sql.lit(LOCK_ID)})
+    from set_timeout`.execute(db)
```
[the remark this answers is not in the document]
igalklebanov: wrong lol.

igalklebanov (approved):
Fixed the test being flaky. Good to go. 🚀

PR #1944, opened 2026-07-17
fix: fix postgres password not being passed to control client

### LINKED ISSUE #1943: controlClient when cancelling queries in postgres does not receive password from options
When using the abort signal with `inflightQueryAbortStrategy` set to `cancel query` the postgres cancel query always fails with the error:
```
inflightQueryAbortHandler` failed in the background after abortion: SASL: SCRAM-SERVER-FIRST-MESSAGE: client password must be a string
```

I have debugged through the driver code to find the following:
image

The `password` field on `poolOptions` is set to non-enumerable:
image

So spreading the object causes the password field to be lost and so it will never be sent to the database, which causes the cancel query to not execute and log the given error.

I will already open a pr for this, since it's a small change, but feel free to do your own implementation

CakeWithDivinity:
Fixes #1943

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -37,7 +37,7 @@ services:
     image: 'postgres'
     environment:
       POSTGRES_DB: kysely_test
+      POSTGRES_PASSWORD: kysely
       POSTGRES_USER: kysely
-      POSTGRES_HOST_AUTH_METHOD: trust
     ports:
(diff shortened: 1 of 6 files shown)


igalklebanov (approved):
Thanks! 🙏 

Verified and added passwords to PostgreSQL test setup.

Issue #1947, opened 2026-07-18
add "dumb container type" for builder generics for cheap type-level comparisons.


igalklebanov:
Hey 👋 

Following https://github.com/microsoft/typescript-go/issues/4579, we:

> need to expose and promote usage of a dumb container type for SelectQueryBuilder's generics so that users match against it and not against SelectQueryBuilder itself.

I remember a PR a while back that looked relevant to this but broke assignability tests.

koskimas:
A dumb container type wouldn't help in the case where the user wants to call arbitrary functions on the argument. For that, it really needs to have those functions, which makes it slow again. Or are you suggesting that people would then cast that dumb type into a full `SelectQueryBuilder` inside the function?

Another issue is that people wouldn't realise they need to use the helper type. But there would at least be a simple solution people could point out to them.

We already have `SelectQueryBuilderExpression` which was created for this very reason for internal use. But it doesn't have the generics.
https://github.com/kysely-org/kysely/blob/master/src/query-builder/select-query-builder-expression.ts

igalklebanov:
> A dumb container type wouldn't help in the case where the user wants to call arbitrary functions on the argument. For that, it really needs to have those functions, which makes it slow again. Or are you suggesting that people would then cast that dumb type into a full SelectQueryBuilder inside the function?

Yeah, signature-level comparison by generics and something minimal that suggests `SelectQueryBuilder`-like to guard input/s, and body-level low cost type-level casting to `SelectQueryBuilder` for method calls.

> Another issue is that people wouldn't realise they need to use the helper type. But there would at least be a simple solution people could point out to them.

Another part of this is promotion of the do and the don't in examples and docs.

> We already have SelectQueryBuilderExpression which was created for this very reason for internal use. But it doesn't have the generics.

We could expand it's signature and expose it if fits and doesn't hurt type-level performance that matters more.

koskimas:
Me and Opus 5 found a solution that removes the infinite recursion and makes typescript 7 3-5x faster than typescript 6. The same solution fixes other performance issues I haven't been able to fix before.

PR coming up. The change itself is tiny.

koskimas:
Results for https://github.com/XiNiHa/kysely-tsgo-slower/blob/main/src/index.ts:

| compiler | instantiations before | instantiations after | change | check time before | check time after | change | total time before | total time after |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
[...]

PR #1960, opened 2026-07-30
fix infinite type check recursion

koskimas:
This issue has existed a long time and I haven't been able to figure out how to fix it. We had a bunch of type tests with a TODO comment about insane slowness. Those cases are now fixed. Mostly thanks to Opus 5.

This fix makes the benchmark linked in the typescript issue https://github.com/microsoft/typescript-go/issues/4579 ~100x faster on ts7. The pathological cases got about 50% faster on TS5/6 as well.

Results for https://github.com/XiNiHa/kysely-tsgo-slower/blob/main/src/index.ts:

| compiler | instantiations before | instantiations after | change | check time before | check time after | change | total time before | total time after |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| TS 5.9.2 | 599,095 | 83,325 | **−86%** | 0.51s | 0.22s | −57% | 0.95s | 0.67s |
| TS 6.0.3 | 601,143 | 93,113 | **−85%** | 0.52s | 0.25s | −52% | 0.68s | 0.48s |
| TS 7.0.2 | 9,795,115 | 94,536 | **−99.0%** | 2.794s | 0.056s | **−98%** | 2.81s | 0.073s |

See #1947 (maybe this closes it?)

### DIFF
--- a/src/query-builder/delete-query-builder.ts
+++ b/src/query-builder/delete-query-builder.ts
@@ -794,7 +794,9 @@ export class DeleteQueryBuilder<DB, TB extends keyof DB, O>
    * delete from `pet` order by `created_at` limit ?
    * ```
    */
-  limit(limit: ValueExpression<DB, TB, number>): DeleteQueryBuilder<DB, TB, O> {
+  limit<VE extends ValueExpression<DB, TB, number>>(
+    limit: VE,
+  ): DeleteQueryBuilder<DB, TB, O> {
     return new DeleteQueryBuilder({
       ...this.#props,
       queryNode: DeleteQueryNode.cloneWithLimit(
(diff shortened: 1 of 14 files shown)


igalklebanov (approved):
Hey 👋 

Amazing stuff! 💪 

The part about `unknown` short circuit is new to me in this scenario. Really nice find.

I believe I saw args-to-generics improvements before that when writing some of these benchmarks and playing around with the methods under test. Still so much to cover there.

Seems like everything gets a small hit in instantiations, but with TypeScript 7 headroom, we're golden and it's well worth it. And at the same time, nice wins in joins that accept a callback - interesting why.

Wonder if we have more of these recursions exploding now elsewhere. 😬

igalklebanov:
Doing a quick check around downstream libraries/codebases to find anything we might possibly break here.

koskimas:
Yeah it seems most fast types got a few percent slower but many slow types (thousands of instantiations) got significantly faster.

> Wonder if we have more of these recursions exploding now elsewhere. 

We do. At least assigning Transaction/ControlledTransaction to Kysely. Weirdly, inheritance doesn't make the check trivial 🤷‍♂️ I have a fix for that as well in the works.

PR #1962, opened 2026-07-31
speed up assigning Transaction to a Kysely

koskimas:
It's a very common thing to have a function that takes a Kysely as an argument and then to pass in a Transaction instance.

Before this commit it was extremely slow. Like hundreds of thousands of type instantations slow, even for a trivial database.

| Case | TS | Instantiations (before → after) | Check time (before → after) |
|---|---|---|---|
| Kysely assignable to Kysely | 5.9 | 11,713 → 11,731 | 60ms → 60ms |
| | 6.0 | 11,786 → 11,804 | 60ms → 60ms |
| | 7.0 | 56,344 → 56,362 | 25ms → 25ms |
| Transaction assignable to Kysely | 5.9 | 149,117 → **11,779** | 270ms → **60ms** |
| | 6.0 | 162,354 → **11,852** | 280ms → **60ms** |
| | 7.0 | 290,374 → **56,410** | 118ms → **24ms** |
| ControlledTransaction assignable to Kysely | 5.9 | 149,342 → **11,793** | 270ms → **60ms** |
| | 6.0 | 162,579 → **11,866** | 290ms → **60ms** |
| | 7.0 | 291,279 → **56,424** | 118ms → **25ms** |
| ControlledTransaction assignable to Transaction | 5.9 | 149,354 → *316,077* | 260ms → *480ms* |
| | 6.0 | 162,591 → *330,118* | 290ms → *500ms* |
| | 7.0 | 291,315 → *338,539* | 123ms → *173ms* |

This change is debatable. It doesn't actually remove the infinite recursion (TS7 still produces more instantations) and it makes assigning `ControlledTransaction` to `Transaction` slower.

I'm fine with not merging this. Let's first see if it's possible to get rid of the recursion alltogether.

### DIFF
--- a/src/kysely.ts
+++ b/src/kysely.ts
@@ -660,10 +660,6 @@ export type Transaction<DB> = TransactionMethods<DB> & Kysely<DB>
 
 /**
  * The methods a {@link Transaction} has on top of the {@link Kysely} ones.
- *
- * These are the members that make a transaction a transaction: the ones that
- * return another transaction rather than a plain `Kysely` instance, and the
- * ones a transaction doesn't support.
  */
 export interface TransactionMethods<DB> {
   /**
--- a/src/kysely.ts
+++ b/src/kysely.ts
@@ -642,7 +642,113 @@ export class Kysely<DB>
   }
 }
 
-export class Transaction<DB> extends Kysely<DB> {
+/**
+ * A {@link Kysely} instance that runs all its queries inside a transaction.
+ *
+ * You get one from {@link Kysely.transaction} or {@link Kysely.startTransaction}.
+ *
+ * A `Transaction<DB>` can be used anywhere a `Kysely<DB>` is expected. The
+ * opposite is not true - a plain `Kysely` instance is not a transaction.
+ *
+ * This is an intersection of {@link TransactionMethods} and {@link Kysely}
+ * rather than a subclass of `Kysely` for type check performance reasons.
+ *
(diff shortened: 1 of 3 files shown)


src/kysely.ts
```diff
@@ -642,128 +642,198 @@ export class Kysely<DB>
...
+ *
+ * You get one from {@link Kysely.transaction} or {@link Kysely.startTransaction}.
+ *
+ * A `Transaction<DB>` can be used anywhere a `Kysely<DB>` is expected. The
+ * opposite is not true - a plain `Kysely` instance is not a transaction.
+ *
+ * This is an intersection of {@link TransactionMethods} and {@link Kysely}
+ * rather than a subclass of `Kysely` for type check performance reasons.
+ *
+ * `Transaction` is still a class at runtime, so `db instanceof Transaction`
```
igalklebanov, line 656: we should not encourage `instanceof`.

src/kysely.ts
```diff
@@ -642,128 +642,198 @@ export class Kysely<DB>
...
+ * works.
+ */
+export type Transaction<DB> = TransactionMethods<DB> & Kysely<DB>
+
+/**
+ * The methods a {@link Transaction} has on top of the {@link Kysely} ones.
+ *
+ * These are the members that make a transaction a transaction: the ones that
+ * return another transaction rather than a plain `Kysely` instance, and the
+ * ones a transaction doesn't support.
```
igalklebanov:
```suggestion
```

PR #1963, opened 2026-07-31
remove recursive type checks from Kysely and subtypes

koskimas:
Remove `Kysely` assignment type check recursion. This one comes with a rather big caveat, that we might not want:

```ts
let k1: Kysely<{ a: A, b: B }>
let k2: Kysely<{ a: A }>
k2 = k1 // This no longer works!
```

Opus says it only worked so far because typescript gave up with the infinite recursion. If it was able to follow it through, the result would've been an error:

> On the assignability change you approved: Kysely<{a,b}> ↔ Kysely<{a}> no longer compiles in either direction. Worth knowing why it used to: the compiler didn't trust the variance it measured and fell back to a structural comparison, where methods match bivariantly. It was never a variance the library could have declared — I confirmed TypeScript rejects out DB with TS2636 because selectFrom takes keyof DB. Users who relied on it need $pickTables/$omitTables.

So do we want to keep that feature, which only works because typescript gives up over infinite recursion (assuming Opus is right)?

This is on top of #1962

### DIFF
--- a/src/kysely.ts
+++ b/src/kysely.ts
@@ -94,7 +94,33 @@ Symbol.asyncDispose ??= Symbol('Symbol.asyncDispose')
  *    in the database and values must be interfaces that describe the rows in those
  *    tables. See the examples above.
  */
-export class Kysely<DB>
+// Why `in out DB` here and on the other generic types in this file:
+//
+// To relate two *different* instantiations of the same generic type - say
+// `Kysely<DB1>` to `Kysely<DB2>` - TypeScript first needs the variance of each
+// type parameter. Nothing declared it, so the compiler measured it: it
+// instantiates the type twice with synthetic marker types and relates the two
+// results structurally, member by member.
+//
+// That measurement was the expensive part, and it didn't terminate on its own.
+// Members like `$extendTables` return the same type over a *recomputed* `DB`, so
+// comparing them starts the whole comparison again one level down, against a `DB`
+// the compiler hasn't seen before and so can't match against a pair already in
+// progress. It only stopped at the compiler's internal recursion limit - and that
+// limit isn't the same across versions, which is why this cost ~5x more on
+// TypeScript 7 than on 6. Relating two `Kysely` types was ~12k type
+// instantiations on TS6 and ~56k on TS7; it is now 2 on both.
+//
+// An explicit annotation skips the measurement entirely: the compiler just uses
+// the declared variance. `in out` (invariant) is the strictest option and the
+// only sound one here - TypeScript rejects `out DB` with TS2636, because
+// `selectFrom` and friends take `keyof DB`.
+//
+// This is not a technique the query builders can borrow. They deliberately let a
+// wider builder stand in for a narrower one, which is bivariance, and bivariance
+// can't be spelled as an annotation. See `test/typings/test-d/generic.test-d.ts`
+// for what keeps those in check instead.
+export class Kysely<in out DB>
   extends QueryCreator<DB>
   implements QueryExecutorProvider, AsyncDisposable
 {
@@ -665,7 +691,7 @@ export type Transaction<DB> = TransactionMethods<DB> & Kysely<DB>
  * return another transaction rather than a plain `Kysely` instance, and the
  * ones a transaction doesn't support.
  */
-export interface TransactionMethods<DB> {
+export interface TransactionMethods<in out DB> {
   /**
    * Always `true` for a transaction.
    *
@@ -901,7 +927,7 @@ export interface KyselyConfig {
   readonly log?: LogConfig
 }
 
-export class ConnectionBuilder<DB> {
+export class ConnectionBuilder<in out DB> {
   readonly #props: ConnectionBuilderProps
 
   constructor(props: ConnectionBuilderProps) {
@@ -929,7 +955,7 @@ export class ConnectionBuilder<DB> {
 
 interface ConnectionBuilderProps extends KyselyProps {}
 
-export class TransactionBuilder<DB> {
+export class TransactionBuilder<in out DB> {
   readonly #props: TransactionBuilderProps
 
   constructor(props: TransactionBuilderProps) {
@@ -996,7 +1022,7 @@ interface TransactionBuilderProps extends KyselyProps {
   readonly isolationLevel?: IsolationLevel
 }
 
-export class ControlledTransactionBuilder<DB> {
+export class ControlledTransactionBuilder<in out DB> {
   readonly #props: TransactionBuilderProps
 
   constructor(props: TransactionBuilderProps) {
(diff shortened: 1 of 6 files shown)


test/ts-benchmarks/generic.bench.ts
```diff
@@ -162,16 +178,39 @@ bench('MergeQueryBuilder assignable to narrower MergeQueryBuilder', () => {
...
-}).types([330097, 'instantiations'])
\ No newline at end of file
+}).types([136150, 'instantiations'])
+
+// The read-only mirror of the four cases above. It has the same shape for the
+// same reason, and had the same problem: before `ReadonlyTransaction` became an
+// intersection, the second of these cost 1.3M instantiations on TypeScript 7.
+bench('ReadonlyKysely assignable to ReadonlyKysely', () => {
+  return acceptsReadonlyKysely(readonlyKysely)
+}).types([26, 'instantiations'])
```
koskimas, line 197: This used to cost 1.3M instantations 😬

src/kysely.ts
```diff
@@ -94,7 +94,33 @@ Symbol.asyncDispose ??= Symbol('Symbol.asyncDispose')
...
+// An explicit annotation skips the measurement entirely: the compiler just uses
+// the declared variance. `in out` (invariant) is the strictest option and the
+// only sound one here - TypeScript rejects `out DB` with TS2636, because
+// `selectFrom` and friends take `keyof DB`.
+//
+// This is not a technique the query builders can borrow. They deliberately let a
+// wider builder stand in for a narrower one, which is bivariance, and bivariance
+// can't be spelled as an annotation. See `test/typings/test-d/generic.test-d.ts`
+// for what keeps those in check instead.
+export class Kysely<in out DB>
```
koskimas, line 97: Rather excessive AI babbling... Do we want this here at all?

igalklebanov: Probably not.

igalklebanov: Its better if this stuff is kept in PR descriptions - agents love getting context from gh api - when necessary.

src/readonly/readonly-query-creator.ts
```diff
@@ -10,7 +10,9 @@ import type {
 /**
  * Similar to {@link QueryCreator} but read-only.
  */
-export interface ReadonlyQueryCreator<DB> extends Pick<
+// `DB` is `in out` for the same reason it is on `Kysely`. See the comment above
+// the `Kysely` class in `src/kysely.ts`.
```
igalklebanov:
```suggestion
```

src/readonly/readonly-kysely.ts
```diff
@@ -58,7 +60,10 @@ import type { ReadonlyQueryCreator } from './readonly-query-creator.js'
  * db.deleteFrom('person') // typescript compiler error!
  * ```
  */
-export interface ReadonlyKysely<DB>
+// `DB` is `in out` on this and the other read-only types below for the same
+// reason it is on `Kysely`. See the comment above the `Kysely` class in
+// `src/kysely.ts`.
```
igalklebanov:
```suggestion
```

test/ts-benchmarks/generic.bench.ts
```diff
@@ -162,16 +178,39 @@ bench('MergeQueryBuilder assignable to narrower MergeQueryBuilder', () => {
 
 bench('Kysely assignable to Kysely', () => {
   return acceptsKysely(plainKysely)
-}).types([11784, 'instantiations'])
+}).types([2, 'instantiations'])
```
igalklebanov, line 181: amazing!

test/ts-benchmarks/generic.bench.ts
```diff
@@ -162,16 +178,39 @@ bench('MergeQueryBuilder assignable to narrower MergeQueryBuilder', () => {
...
 
 bench('ControlledTransaction assignable to Kysely', () => {
   return acceptsKysely(controlledTransaction)
-}).types([11846, 'instantiations'])
+}).types([64, 'instantiations'])
 
+// Still the expensive one: a `ControlledTransaction` and a `Transaction` share
+// members whose return types are these same two intersections, so the compiler
+// has to walk both in full. It doesn't recurse without bound - the count barely
+// moves between TypeScript versions - it's just a lot of members.
```
igalklebanov:
```suggestion
```

src/readonly/readonly-kysely.ts
```diff
@@ -193,36 +195,58 @@ export interface ReadonlyTransactionBuilder<DB> {
...
+ * {@link ReadonlyTransactionMethods} and {@link ReadonlyKysely} rather than an
+ * interface that repeats their members, for type check performance reasons.
+ *
+ * Spelling it this way is what makes a `ReadonlyTransaction` usable where a
+ * {@link ReadonlyKysely} is expected: the compiler finds the very same
+ * `ReadonlyKysely<DB>` among the intersection's members and stops there. As an
+ * interface of its own it had to relate the two structurally, and the members
+ * that return themselves over a recomputed `DB` - `$extendTables` and friends -
+ * made that recurse until the compiler gave up. That cost 1.3M type
+ * instantiations on TypeScript 7, five times what it cost on 6.
```
igalklebanov:
```suggestion
```

src/kysely.ts
```diff
@@ -94,7 +94,33 @@ Symbol.asyncDispose ??= Symbol('Symbol.asyncDispose')
...
+// results structurally, member by member.
+//
+// That measurement was the expensive part, and it didn't terminate on its own.
+// Members like `$extendTables` return the same type over a *recomputed* `DB`, so
+// comparing them starts the whole comparison again one level down, against a `DB`
+// the compiler hasn't seen before and so can't match against a pair already in
+// progress. It only stopped at the compiler's internal recursion limit - and that
+// limit isn't the same across versions, which is why this cost ~5x more on
+// TypeScript 7 than on 6. Relating two `Kysely` types was ~12k type
+// instantiations on TS6 and ~56k on TS7; it is now 2 on both.
```
igalklebanov:
```suggestion
```

src/kysely.ts
```diff
@@ -94,7 +94,17 @@ Symbol.asyncDispose ??= Symbol('Symbol.asyncDispose')
...
+//
+// An explicit annotation skips the measurement entirely: the compiler just uses
+// the declared variance. `in out` (invariant) is the strictest option and the
+// only sound one here - TypeScript rejects `out DB` with TS2636, because
+// `selectFrom` and friends take `keyof DB`.
+//
+// This is not a technique the query builders can borrow. They deliberately let a
+// wider builder stand in for a narrower one, which is bivariance, and bivariance
+// can't be spelled as an annotation. See `test/typings/test-d/generic.test-d.ts`
+// for what keeps those in check instead.
```
igalklebanov:
```suggestion
```

test/ts-benchmarks/generic.bench.ts
```diff
@@ -162,16 +178,35 @@ bench('MergeQueryBuilder assignable to narrower MergeQueryBuilder', () => {
...
 
 bench('ControlledTransaction assignable to Transaction', () => {
   return acceptsTransaction(controlledTransaction)
-}).types([330097, 'instantiations'])
\ No newline at end of file
+}).types([136150, 'instantiations'])
+
+// The read-only mirror of the four cases above. It has the same shape for the
+// same reason, and had the same problem: before `ReadonlyTransaction` became an
+// intersection, the second of these cost 1.3M instantiations on TypeScript 7.
```
igalklebanov:
```suggestion
```

PR #1971, opened 2026-08-06
Add missing beforeThrow calls in getInflightQueryAbortHandler

lourd:
Fixes a bug where the connection would leak for a dialect that didn't support one of the abort strategies. Discovered while working on https://github.com/kysely-org/kysely-postgres-js/pull/459


igalklebanov (approved):
LGTM! 🚀

